PR #36, opened 2020-05-19
Streamline lcr-modules and document contribution guidelines

BrunoGrandePhD:


rdmorin:
The contribution guidelines are much more clear and explicit now. Thanks for all the work you put into this!
I have a lingering question that had come up when I was attempting to implement IGV screenshots and I still am not clear on the answer. It seems like you are encouraging users to make all available options/parameters available in the snakemake rule that runs a script. Aside: for GATK that will be super fun. 
Seriously, though, how should we accomplish this when some parameters or options are only ever used in the context of other ones (or can be mutually exclusive)? I don't see how that can be specified or enforced in the config. Do we just expect users that set up the config to know?

whelena:
I think I accidentally did something, I'm not sure how to reverse it

CONTRIBUTING.md
```diff
@@ -285,47 +295,47 @@ Every other rule serve to complete the module. These other rules can vary consid
 An example rule that follows most of these principles is included below (taken from the `star` module).
 
 1. Each rule should only consist of one command, unless the rule uses standard tools like `gzip` for additional commands. Otherwise, split into multiple rules, optionally connected using `pipe()` or `temp()` to avoid intermediate files.
-   
-   > This guideline ensures that rules are module and can easily be rearranged by the user. It also ensures that tool-specific conda environments (_e.g._ `samtools`, `star`) can be used, which is not possible is more than one tool is used in a rule. 
-   
+
+   > This guideline ensures that rules are module and can easily be rearranged by the user. It also ensures that tool-specific conda environments (_e.g._ `samtools`, `star`) can be used, which is not possible is more than one tool is used in a rule.
```
whelena: `... rules are module ...`
is this part is a typo?

BrunoGrandePhD: Indeed, it is! Fixed!

BrunoGrandePhD:
> I think I accidentally did something, I'm not sure how to reverse it

No worries! I reverted the PR to a draft.

BrunoGrandePhD:
> It seems like you are encouraging users to make all available options/parameters available in the snakemake rule that runs a script. 

Where did you get this impression from the documentation? The module developer only includes the command-line arguments that are required to run the tool. Any arguments associated with a placeholder (like an input/output file) are included in the `shell` directive, and everything else is included in the `default.yaml` configuration file as a string. 

The module developer does **not** have to add every single command-line argument to the Snakefile or configuration file. They only need to add the ones they use. 

> Seriously, though, how should we accomplish this when some parameters or options are only ever used in the context of other ones (or can be mutually exclusive)? I don't see how that can be specified or enforced in the config. Do we just expect users that set up the config to know?

Default command-line arguments are determined by the module developer. If a module user wants to change these arguments, they should at least be familiar with the tool's command-line interface. If they try an incompatible combination of arguments, the underlying tool should throw an error. There is no reason for `lcr-modules` to prevent these situations if the tool will fail anyways.

rdmorin:
I probably misunderstood the intention of adding all command line arguments (with some commented out). If the person that implements the tool is not providing a means to use each of these in the rule, then what purpose do the commented command-line options serve? If a user un-comments one, must they then change the rule to allow that argument to be passed to the tool? I could also be fundamentally misunderstanding how options are passed to the tool and will look more carefully at this in the new documentation once it is in Master.

BrunoGrandePhD:
I added a clarification on why the options for STAR are commented out (see below). Note that this won't always be the case. 

> In the example below, the command-line options for STAR are commented out using `#!` because they require user intervention. Specifically, the value provided to the `--sjdbOverhang` argument should match the value provided to the `star_overhang` key under `reference_params` earlier in the configuration file (see above). A comment explains the user intervention that is required.

CONTRIBUTING.md
```diff
@@ -285,47 +295,47 @@ Every other rule serve to complete the module. These other rules can vary consid
...
 9. Re-use (or provide) tool-specific conda environments for each rule needing one, which are configured in the `default.yaml` file. This can be skipped if the rule only uses standard UNIX tools (_e.g._ `gzip`, `awk`) or if it uses the `run` directive (instead of the `shell` directive).
 
    > Conda environments simplify software installation for a module and ensure reproducibility by specifying tool versions. Even if a rule only uses standard UNIX tools, it might still be worth using the `coreutils` conda environment to avoid OS variations (_e.g._ GNU vs BSD for `sed`).
 
 10. Add the `threads` and `resources` (`mem_mb`) directives for all non-local rules, which are configured in the `default.yaml` file.
-    
+
     > These directives are essential for running the module on a compute cluster. The values should be as low as possible while ensuring that most jobs are run within a reasonable amount of time (to minimize time spent in the queue).
 
 11. Use the `shell` directive for rules with the `conda` directive. Use the `run` directive instead otherwise if more complicated logic is required.
```
whelena: '... instead otherwise ...' 
typo?

whelena:
For the utils module, is that limited to `bam_utils`? I'm wondering if there would be any other file type that might require a utilities snakefile. For example, before using vcf2maf, I had a `vcfutils` snakefile that annotates the vcf files.

When you give an example using `CFG["runs"]` I think it would be a good idea of showing an example of one so that users/developers can distinguish between `CFG["runs"]` and `CFG["samples"]`.

BrunoGrandePhD:
> For the utils module, is that limited to `bam_utils`? I'm wondering if there would be any other file type that might require a utilities snakefile. For example, before using vcf2maf, I had a `vcfutils` snakefile that annotates the vcf files.

For now, I was thinking of having just one master `utils` module for shared rules. This keeps it simpler for module developers. If it grows too large, we can think of splitting it into smaller modules. 

> When you give an example using `CFG["runs"]` I think it would be a good idea of showing an example of one so that users/developers can distinguish between `CFG["runs"]` and `CFG["samples"]`.

Could you clarify what you mean? I give an example for both `CFG["runs"]` and `CFG["samples"]` in the "Target rules" section. For example, I state: 

> This rule is usually named `_<module_name>_all` (_e.g._ `_star_all`), and expands all of the output files (the files symlinked into `99-outputs`) using either the samples table (`CFG["samples"]`) or the runs table (`CFG["runs"]`) depending on whether the module is run once per sample or once per tumour.

Let me know if I should add a clarification here.

whelena:
> Could you clarify what you mean? I give an example for both CFG["runs"] and CFG["samples"] in the "Target rules" section.

I think it would be useful to show the contents of the runs and samples dataframe to show the differences between them. I think it'll help build intuition on when to use each of them.

BrunoGrandePhD:
That's a good idea! I'll add example tables.

CONTRIBUTING.md
```diff
@@ -378,28 +535,91 @@ More details will be added later.
...
+#### Switch on wildcard value
+
+You can use the `op.switch_on_wildcard()` function to dynamically set the value of an input file or parameter for a snakemake rule based on the value of a wildcard. The first argument (`wildcard`) is the name of the wildcard, and the second argument (`options`) is a dictionary mapping possible values for the wildcard to the corresponding values that should be returned.
+
+This dictionary can make use of special keys. The most important one to note is the `"_default"` special key, whose associated value is selected if the wildcard value isn't among the other keys. You should check out the function docstring with `help(op.switch_on_wildcard)` to find out about the other special keys. See [below](#what-does-the-underscore-prefix-mean) for an explanation for the underscore prefix.
+
+By default, the `op.switch_on_wildcard()` will replace any placeholders (using the same format as the `shell` directive; _e.g._ `{wildcards.seq_type}`) with the actual values. This beheviour can be tweaked with the `format` (default = `True`) and `strict` (default = `False`) optional arguments. See the function docstring for more informaiton on these optional arguments.
```
rdmorin: docstring for more ~informaiton~ information

BrunoGrandePhD:
Because of a frustrating bug, this took longer than expect, but I can finally point to a nice web version of the oncopipe function documentation (generated automatically from the docstrings). 

https://lcr-modules.readthedocs.io/en/dev-bgrande/oncopipe.html

BrunoGrandePhD:
I've worked hard on migrating the documentation to ReadTheDocs. This enables automatic documentation of the `oncopipe` interface (based on the contents of the docstrings) as well as easy linking between different sections of the documentation. 

I have officially opened this PR for feedback on the documentation for users, which I have heavily reworked. You can read it using the link below (rather than reading the raw reStructuredText) and comment here just by referring to sections by name or by copying-pasting relevant passages that need additional work. 

https://lcr-modules.readthedocs.io/en/dev-bgrande/for_users.html

I'm going to work on the developer documentation tomorrow, which shouldn't need as much work since it's more or less in line with the latest best practices for `lcr-modules`.

BrunoGrandePhD:
Also, feel free to test the `manta` 2.0 and `star` 1.0 modules. Currently, the `star` module only takes in FASTQ files until the BAM-to-FASTQ module is implemented. If you want, you can run through the [Getting Started](https://lcr-modules.readthedocs.io/en/dev-bgrande/for_users.html#getting-started) instructions and test the demo snakefile.

BrunoGrandePhD:
The developer documentation has been updated. I'm sure there are things to improve, but I'm not going to let perfect get in the way of good. It's time to ship this! 

I made a small update to the manta 2.0 module. I'm going to run a quick test to make sure I didn't break anything.

PR #38, opened 2020-06-04
Module/strelka2/1.0

whelena:
# Pull Request Checklists

### Required

- [x] The snakemake rules follow the [design guidelines](../CONTRIBUTING.md#module-rules).

- [x] I tested the module on demo data (capture and mrna).
   
  ```bash
  ```

### If applicable

To be completed.


demo/Snakefile
```diff
@@ -46,12 +47,13 @@ config["lcr-modules"]["_shared"]["samples"] = SAMPLES
 # Load module-specific snakefiles
 include: "../modules/star/1.0/star.smk"
 include: "../modules/manta/2.0/manta.smk"
-
+include: "../modules/strelka/1.0/strelka.smk"
```
BrunoGrandePhD, line 50: Just in case we want to add a strelka1 module later, could you make sure the module name is `strelka2` everywhere? Thanks!

demo/run.sh
```diff
@@ -1,3 +1,4 @@
 #!/bin/bash
 snakemake --keep-going --latency-wait 120 --jobs 100 --cluster-sync "srun -J {rule} --cpus-per-task {threads} --mem {resources.mem_mb}" \
---use-conda --conda-prefix /projects/clc/usr/anaconda/workflow_prototype/lcr-modules-envs/ all
+--use-conda --conda-prefix /projects/clc/usr/anaconda/workflow_prototype/lcr-modules-envs/ all \
```
BrunoGrandePhD: This script should be general. Right now, it's tailored for your usage. Feel free to maintain a copy of a script for your purposes (e.g., `demo/run-clc.sh`) that you omit from the repo. I already updated the script in `master` to achieve this. You should be able to import that version by running this:

```
❯ cat demo/run.sh
#!/bin/bash
snakemake --keep-going --latency-wait 120 --jobs 100 --cluster-sync "srun -J {rule} --cpus-per-task {threads} --mem {resources.mem_mb}" \
--use-conda --conda-prefix /projects/clc/usr/anaconda/workflow_prototype/lcr-modules-envs/ all \
--rerun-incomplete

❯ git checkout master -- demo/run.sh

❯ cat demo/run.sh
#!/bin/bash

# Number of cores set by first command-line argument (default is 4)
NUM_CORES=${1:-4}

snakemake --cores "${NUM_CORES}" --printshellcmds --use-conda all
```

modules/strelka/1.0/config/default.yaml
```diff
@@ -0,0 +1,39 @@
...
+                run_unpaired_tumours_with: "unmatched_normal"
+                run_paired_tumours_as_unpaired: False
+            capture:
+                run_paired_tumours: True
+                run_unpaired_tumours_with: "unmatched_normal"
+                run_paired_tumours_as_unpaired: False
+            mrna:
+                run_paired_tumours: False
+                run_unpaired_tumours_with: "no_normal"
+                run_paired_tumours_as_unpaired: True
```
BrunoGrandePhD: I don't think Strelka is typically run on RNA-seq data, so I think you can delete this section.

BrunoGrandePhD: I noticed that you have rules to run Strelka in tumour-only mode or rather germline mode. Is this something that the CLC does?

modules/strelka/1.0/config/default.yaml
```diff
@@ -0,0 +1,39 @@
+lcr-modules:
+    
+    strelka:
+        inputs:
+            # Available wildcards: {seq_type} {genome_build} {tumour_id} {normal_id} {pair_status}
+            sample_bam: null 
+            sample_bai: null
+            candidate_indel: null
```
BrunoGrandePhD: These lines should be annotated by `UPDATE` comments.

whelena:
> You should leave the checkboxes in `.github/PULL_REQUEST_TEMPLATE.md` unchecked. That is the template used to pre-fill the pull request when you open it. You check directly in the PR.

 I will fix that

whelena:
> You should leave the empty placeholder files as is in the demo project. Otherwise, the demo snakefile fails on a dry run when the symlinks are broken (i.e., whenever the repo is cloned anywhere other than at the GSC).
> 

I made an error of running the test run from the demo directory but I will revert it back to the original placeholder files when the run is complete

modules/strelka/1.0/config/default.yaml
whelena: The snakefile I got from Lauren runs strelka on tumour-normal pairs only. I found the germline workflow but wasn't sure if it is tumour-only. Have you used it before?

BrunoGrandePhD:
Once this is ready for review, let me know!

BrunoGrandePhD:
@whelena: Can you confirm that you were able to run this module on some test data?

BrunoGrandePhD:
I launched the test. The only thing I needed to fix before it could run is adding the tabix conda environment for the `_strelka_input_vcf` rule. I pushed my fix in e429c8e. I also resolved the merge conflicts with master in f4501a3. 

I'm still waiting for the module to finish running. Once that's done, I'll see if any improvements can be made to the module.

whelena:
thank you!

BrunoGrandePhD:
The test ran to completion on six genome pairs! I'll make some quick comments now that you can hopefully address before I review more carefully this weekend.

.gitignore
```diff
@@ -28,14 +28,16 @@ share/python-wheels/
...
+# Files created when testing the demo project
 .snakemake/
 /demo/results/
 /demo/reference
 /demo/scratch
 /demo/data/*
 !/demo/data/samples.tsv
 
+demo/run_clc.sh
+
```
BrunoGrandePhD: Maybe we should come up with a more general pattern for excluding files from Git. Maybe we just ignore any files containing `tmp.`. And then, you can rename `demo/run_clc.sh` to either `demo/run_clc.tmp.sh` or `demo/tmp.run_clc.sh` (whatever you prefer). 

If you agree with this approach, then you can replace `demo/run_clc.sh` with `*tmp.*` in the `.gitignore` file.

demo/config.yaml
```diff
@@ -21,3 +21,9 @@ lcr-modules:
         inputs:
             sample_bam: "data/{sample_id}.bam"
             sample_bai: "data/{sample_id}.bam.bai"
+
+    strelka:
+        inputs:
+            sample_bam: "data/{sample_id}.bam"
+            sample_bai: "data/{sample_id}.bam.bai"
+            candidate_indel: "results/manta-2.0/99-outputs/vcf/{seq_type}--{genome_build}/candidateSmallIndels/{tumour_id}--{normal_id}--{pair_status}.candidateSmallIndels.vcf"
```
BrunoGrandePhD: We should try to think of a way to simplify these inter-module connections for users. Any ideas?

envs/strelka/strelka-2.9.10.yaml
```diff
@@ -0,0 +1,28 @@
...
+dependencies:
+  - _libgcc_mutex=0.1
+  - _openmp_mutex=4.5
+  - ca-certificates=2019.11.28
+  - certifi=2019.11.28
+  - libffi=3.2.1
+  - libgcc-ng=9.2.0
+  - libgomp=9.2.0
+  - libstdcxx-ng=9.2.0
+  - manta=1.6.0
```
BrunoGrandePhD: Is Manta a dependency of Strelka? Or was this added by you? If so, could you create a Strelka conda environment without this extraneous dependency since we already have a manta env? Thanks!

modules/strelka/1.0/config/default.yaml
```diff
@@ -0,0 +1,36 @@
+lcr-modules:
+    
+    strelka:
+        inputs:
+            # Available wildcards: {seq_type} {genome_build} {tumour_id} {normal_id} {pair_status}
```
BrunoGrandePhD, line 9: I would move this line to right before `candidate_indel` and add a separate comment here with:
```
# Available wildcards: {seq_type} {genome_build} {sample_id}
```

modules/strelka/1.0/config/default.yaml
```diff
@@ -0,0 +1,36 @@
...
+            sample_bam: null  # UPDATE
+            sample_bai: null  # UPDATE
+            candidate_indel: null  # UPDATE
```
BrunoGrandePhD: Could you replace the `null  # UPDATE` with `"__UPDATE__"`? I will be submitting a PR this weekend, which has moved to this new method of identifying values that need to be updated.

BrunoGrandePhD: Could you rename `candidate_indel` to `candidate_small_indels_vcf` for clarity?

modules/strelka/1.0/config/default.yaml
```diff
@@ -0,0 +1,36 @@
...
+                mrna: "--rna"
+                capture: "--exome"
+            strelka: "-m local"
+
+        conda_envs:
+            strelka: "{MODSDIR}/envs/strelka-2.9.10.yaml"
+            tabix: "{MODSDIR}/envs/tabix-0.2.6.yaml"
+            
+        threads:
+            strelka: 4
```
BrunoGrandePhD: I would go with a default of 12 like with Manta.

modules/strelka/1.0/config/default.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+        mem_mb:
+            strelka: 2000
```
BrunoGrandePhD: Have you done any testing with Strelka to check its peak memory usage?

modules/strelka/1.0/config/default.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+        pairing_config:
+            genome:
+                run_paired_tumours: True
+                run_unpaired_tumours_with: "unmatched_normal"
+                run_paired_tumours_as_unpaired: False
+            capture:
+                run_paired_tumours: True
+                run_unpaired_tumours_with: "unmatched_normal"
+                run_paired_tumours_as_unpaired: False
```
BrunoGrandePhD: A few of your files don't end with a newline (indicated by a red `–` in GitHub). You can avoid this in VS Code with `"files.insertFinalNewline": true`.

modules/strelka/1.0/strelka.smk
```diff
@@ -0,0 +1,213 @@
...
+        bam = CFG["dirs"]["inputs"] + "{seq_type}--{genome_build}/{sample_id}.bam",
+        bai = CFG["dirs"]["inputs"] + "{seq_type}--{genome_build}/{sample_id}.bam.bai"
+    run:
+        op.relative_symlink(input.bam, output.bam)
+        op.relative_symlink(input.bai, output.bai)
+
+
+rule _strelka_input_vcf:
+    input:
+        vcf = CFG["inputs"].get("candidate_indel")
```
BrunoGrandePhD: I've leaved against using `.get()` in modules because it will return None rather than triggering an error. 

Actually, does this module currently only run if the candidate indel file is provided? Is there a way to configure it such that it's optional?

modules/strelka/1.0/strelka.smk
```diff
@@ -0,0 +1,213 @@
...
+    output:
+        vcf = CFG["dirs"]["inputs"] + "{seq_type}--{genome_build}/vcf/{tumour_id}--{normal_id}--{pair_status}.candidateSmallIndels.vcf.gz"
+    conda:
+        CFG["conda_envs"]["tabix"]
+    shell:
+        "bgzip -c {input.vcf} > {output.vcf} && tabix -p vcf {output.vcf}"
```
BrunoGrandePhD: The output of Manta is already bgzip-compressed and tabix-indexed. I wonder if we can just point to this file before it gets decompressed by subsequent steps in the manta module (see example path below). I just don't know if Snakemake will properly configure the `_manta_run` rule as a dependency of this rule here. The [output of the `_manta_run` rule is a directory](https://github.com/LCR-BCCRC/lcr-modules/blob/master/modules/manta/2.0/manta.smk#L129), and this file would be within that directory, so I'm hoping that the dependency link would be created. Could you test that?

```
results/manta-1.0/02-manta/{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/results/variants/candidateSmallIndels.vcf.gz
```

BrunoGrandePhD: If this works, then we can just rely on symlinking here.

modules/strelka/1.0/strelka.smk
```diff
@@ -0,0 +1,213 @@
+#!/usr/bin/env snakemake
```
BrunoGrandePhD, line 1: The Strelka user guide has the same recommendation as Manta for excluding the small contigs. Could you try implementing that just like what I did with Manta. 

https://github.com/Illumina/strelka/blob/v2.9.x/docs/userGuide/README.md#improving-runtime-for-references-with-many-short-contigs-such-as-grch38

modules/strelka/1.0/strelka.smk
```diff
@@ -0,0 +1,213 @@
...
+        --normalBam={input.normal_bam}
+        --tumorBam={input.tumour_bam}
+        --referenceFasta={input.fasta}
+        --runDir=$(dirname {output.runwf})
+        --indelCandidates={input.indels}
+        {params.opts} 
+        > {log.stdout} 2> {log.stderr}
+        """)
+
+rule _strelka_configure_unpaired: # germline
```
BrunoGrandePhD, line 139: I'm afraid that users will accidentally run tumours in `no_normal` mode and expect a somatic variant calling model to be used (rather than the germline model used here). Maybe add a message directive to this rule warning the user about this. 

https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#messages

modules/strelka/1.0/strelka.smk
```diff
@@ -0,0 +1,213 @@
...
+
+
+# Example variant filtering rule (single-threaded; can be run on cluster head node)
+rule _strelka_run:
+    input:
+        runwf = CFG["dirs"]["strelka"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/runWorkflow.py"
+    output:
+        vcf_dir = directory(CFG["dirs"]["strelka"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/results/variants/")
+        #snp = CFG["dirs"]["strelka"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/results/variants/somatic.snvs.vcf.gz",
+        #indels = CFG["dirs"]["strelka"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/results/variants/somatic.indels.vcf.gz"
```
BrunoGrandePhD: Remove comments

modules/strelka/1.0/strelka.smk
```diff
@@ -0,0 +1,213 @@
...
+    threads:
+        CFG["threads"]["strelka"]
+    resources: 
+        mem_mb = CFG["mem_mb"]["strelka"]
+    shell:
+        op.as_one_line("""
+        {input.runwf} -j {threads} 
+        {params.opts} 
+        >{log.stdout} 2> {log.stderr}
+        """)
```
BrunoGrandePhD: You can probably collapse short commands like these to one line.

modules/strelka/1.0/strelka.smk
```diff
@@ -0,0 +1,213 @@
...
+
+
+rule _strelka_unzip:
+    input:
+        vcf_dir = rules._strelka_run.output.vcf_dir
+        #vcf = CFG["dirs"]["strelka"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/results/variants/{var_type}.vcf.gz"
```
BrunoGrandePhD: Remove comment

modules/strelka/1.0/strelka.smk
```diff
@@ -0,0 +1,213 @@
...
+    output:
+        vcf = CFG["dirs"]["filtered"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/{var_type}.vcf"
+    shell:
+        "zcat {input.vcf_dir}/{wildcards.var_type}.vcf.gz > {output.vcf}"
```
BrunoGrandePhD: Let's rely on `gzip -dc` instead of `zcat` for consistency. I don't know how standard `zcat` is.

modules/strelka/1.0/strelka.smk
```diff
@@ -0,0 +1,213 @@
...
+        mem_mb = CFG["mem_mb"]["strelka"]
+    shell:
+        op.as_one_line("""
+        {input.runwf} -j {threads} 
+        {params.opts} 
+        >{log.stdout} 2> {log.stderr}
+        """)
+
+
+rule _strelka_unzip:
```
BrunoGrandePhD: Since zip is its own file format, I could rename this more generally `_strelka_decompress_vcf`

modules/strelka/1.0/strelka.smk
```diff
@@ -0,0 +1,213 @@
...
+
+rule _strelka_filter:
+    input:
+        vcf = rules._strelka_unzip.output.vcf
+    output:
+        vcf = CFG["dirs"]["filtered"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/{var_type}.filt.vcf"
+    shell:
+        op.as_one_line("""
+        grep '^#' {input.vcf} > {output.vcf} && 
+        awk '{{ if ($7 == \"PASS\") {{ print $0 }} }}' {input.vcf} >> {output.vcf}
```
BrunoGrandePhD: You can simplify this command with:
```
awk 'BEGIN {FS=OFS="\t"} $0 ~ /^#/ || $7 == "PASS"'
```

If all you have is a conditional, you don't need to specify `print $0`.

modules/strelka/1.0/strelka.smk
```diff
@@ -0,0 +1,213 @@
...
+        """)
+
+
+# Symlinks the final output files into the module results directory (under '99-outputs/')
+rule _strelka_output_vcf:
+    input:
+        vcf = rules._strelka_filter.output.vcf
+    output:
+        vcf = CFG["dirs"]["outputs"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/{var_type}.vcf"
```
BrunoGrandePhD: I like organizing the output files by file type. Also, I would specify that these are only the passed variants. Here, it would be:
```
CFG["dirs"]["outputs"] + "vcf/{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}.{var_type}.passed.vcf"
```

Like in Manta: https://github.com/LCR-BCCRC/lcr-modules/blob/master/modules/manta/2.0/manta.smk#L200

modules/strelka/1.0/strelka.smk
```diff
@@ -0,0 +1,213 @@
...
+        vcf_files = ["somatic.snvs", "somatic.indels"]
+    vcf = expand(rules._strelka_output_vcf.output.vcf, var_type = vcf_files, **wildcards)
+    return vcf
+
+
+rule _strelka_dispatch:
+    input: 
+        vcf = _get_strelka_output
+    output:
+        dispatched = touch(CFG["dirs"]["outputs"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}.dispatched")
```
BrunoGrandePhD: I would store all of these dispatched files in a subdirectory like I did in Manta (similar to my comment about the `vcf/` subdirectory I made above). 

https://github.com/LCR-BCCRC/lcr-modules/blob/master/modules/manta/2.0/manta.smk#L278

modules/strelka/1.0/strelka.smk
```diff
@@ -0,0 +1,213 @@
...
+
+
+# Generates the target sentinels for each run, which generate the symlinks
+rule _strelka_all:
```
BrunoGrandePhD, line 247: You might want to expose both the filtered and unfiltered VCF files in the `99-outputs` directory. But when we implement variant annotation, we can just annotate the filtered VCF because it would take too much time otherwise.

modules/strelka/CHANGELOG.md
```diff
@@ -0,0 +1,18 @@
...
+
+This release was authored by Helena Winata.
+
+
+<!-- TODO: Explain each important module design decision below. -->
+
+- `_strelka_run` outputs to a directory because strelka creates vcf files in `${STRELKA_ANALYSIS_PATH}/results/variants`
+- `_strelka_dispatch` is used to call outputs based on the `{pair_status}` wildcard since outputs files are automatically generate.
+    - somatic workflow outputs `somatic.snvs.vcf.gz` and `somatic.indels.vcf.gz`
+    -germline workflow outputs `variants.vcf.gz`
```
BrunoGrandePhD: Maybe to be explicit, we should use `germline.snvs.vcf.gz` instead of `variants.vcf.gz`. This way, the user knows what they're getting.

.gitignore
whelena: ya I think that's a better approach

whelena:
> The test ran to completion on six genome pairs! I'll make some quick comments now that you can hopefully address before I review more carefully this weekend.

I'm currently running some testing on the bam_util modules but I will do the fixes some time tomorrow

modules/strelka/CHANGELOG.md
whelena: `variants.vcf.gz` is automatically generated. Is there a way I can specify the output or I could change the name when symlinking to the output directory?
Changing the name might be confusing if users want to trace back which files are which

envs/strelka/strelka-2.9.10.yaml
whelena: I exported it from a pre-existing environment but ya I can remove it.

demo/config.yaml
whelena: hmm I'm not sure, maybe we should use `config["lcr-modules"]["manta"]["dirs"]["outputs"] + "/vcf/{seq_type}--{genome_build}/candidateSmallIndels/{tumour_id}--{normal_id}--{pair_status}.candidateSmallIndels.vcf"` as a default in the snakefile ?
So it would we CFG["inputs"]["candidate_indel"] or default.

Not sure if this makes things simpler

modules/strelka/1.0/config/default.yaml
whelena: No I haven't. 2000 seems to be enough for mRNA and capture. What did you use for your genome samples?

modules/strelka/1.0/config/default.yaml
whelena: thank you!

modules/strelka/1.0/strelka.smk
whelena: currently yes, but I can make it optional

modules/strelka/1.0/strelka.smk
whelena: So it doesn't create the dependency if we try to symlink the vcf.gz. It would still work if we call _manta_all and _strelka_all in the Snakefile.

modules/strelka/1.0/strelka.smk
whelena: Thanks!

BrunoGrandePhD:
Let me know when your PR has been updated. I'll give it another test run on a larger dataset using numbers.

whelena:
I will push the commits sometime tomorrow. I was testing a modified version with conditional `candidate_small_indel`

whelena:
sound good, thank you!

BrunoGrandePhD:
How did you determine the 2000 MB of memory (`mem_mb`) for Strelka? All but one job got killed on numbers for exceeding the memory limit. I'm wondering if my jobs are using much more memory than your jobs.

whelena:
I ran a capture run on gphost using /usr/bin/time -v to get the max rss

BrunoGrandePhD:
I ran into some strange stalling when running the module on the numbers cluster. I'm trying again on a gphost to see if the issue is module- or numbers-related.

BrunoGrandePhD:
I just realized that the `_strelka_filter` rule doesn't specify the input and output file in the awk command, so it gets stuck waiting for input on stdin. Let me see if that fixes this.

whelena:
I can fix it and push the commit today

whelena:
Okay I'll test it out! Thank you!

whelena:
I get this error while running with null values in `candidate_small_indels_vcf` and `candidate_small_indels_tbi`
`SyntaxError:
Input and output files have to be specified as strings or lists of strings.
  File "/projects/clc/clc_scratch/projects/hwinata/dev-modules/test/Snakefile", line 58, in <module>
  File "/projects/clc/clc_scratch/projects/hwinata/dev-modules/lcr-modules/modules/strelka/1.0/strelka.smk", line 62, in <module>`

I wonder if there's a way to run a rule conditionally. So that if those are empty, the symlinking rules won't be executed

BrunoGrandePhD:
Can you push any changes you've made so I can take a look at line 62 (mentioned in the error)? I was expecting that `null`/`None` would evaluate to false in [this line](https://github.com/LCR-BCCRC/lcr-modules/blob/module/strelka2/1.0/modules/strelka/1.0/strelka.smk#L102) and that would end up requesting the dummy file rather than trying to symlink a real file. 

Maybe you can try adding `or ""` after each of these [two lines](https://github.com/LCR-BCCRC/lcr-modules/blob/module/strelka2/1.0/modules/strelka/1.0/strelka.smk#L57-L58)?

whelena:
I pushed one that fixes the null inputs. I put the the rules inside an if condition and it somehow works. I also edited the `_get_indel_cli_arg` function because the previous one still tries to push `--indelCandidates=results/strelka-1.0/00-inputs/capture--grch37/vcf/A88333--A88355--matched.dummy.tbi` as a command line argument

modules/strelka/1.0/strelka.smk
```diff
@@ -52,21 +52,25 @@ rule _strelka_input_bam:
...
-    run:
-        op.relative_symlink(input.vcf, output.vcf)
-        op.relative_symlink(input.tbi, output.tbi)
+        touch(CFG["dirs"]["inputs"] + "{seq_type}--{genome_build}/vcf/{tumour_id}--{normal_id}--{pair_status}.dummy.tbi")
 
 
-rule _strelka_dummy_vcf:
-    output:
-        touch(CFG["dirs"]["inputs"] + "{seq_type}--{genome_build}/vcf/{tumour_id}--{normal_id}--{pair_status}.dummy")
+if not CFG["inputs"]["candidate_small_indels_tbi"] and CFG["inputs"]["candidate_small_indels_vcf"]:
```
BrunoGrandePhD: Wait, does this work?? I never considered having a rule in an if statement.

whelena: It works for the null statements and I'm just about to test on the other condition.

whelena: so it doesn't work because I think they try to figure the DAG before the if statements?

whelena:
I'm getting 
`Missing input files for rule _strelka_input_vcf:
results/manta-2.1/02-manta/capture--grch37/A88333--A88355--matched/results/variants/candidateSmallIndels.vcf.gz.tbi
results/manta-2.1/99-outputs/vcf/capture--grch37/candidateSmallIndels/A88333--A88355--matched.candidateSmallIndels.vcf`

modules/strelka/1.0/strelka.smk
whelena: i think its working now

whelena:
I realised that the manta.smk in 2.1 is a symlink and `op.setup_module` still uses `version="2.1"`. I think that's why I have been getting `MissingInputException`

BrunoGrandePhD:
> I realised that the manta.smk in 2.1 is a symlink and `op.setup_module` still uses `version="2.1"`. I think that's why I have been getting `MissingInputException`

Omg, I'm so dumb... Could you copy the snakefiles for manta 2.1 and star 1.1 and update the version numbers? I'm so sorry about this!

whelena:
No worries, I just noticed it too but ya I can do that.

whelena:
```
configureStrelkaSomaticWorkflow.py --normalBam=results/strelka-1.0/00-inputs/bam/capture--grch37/TCRBOA7-N-WEX.bam --tumorBam=results/strelka-1.0/00-inputs/bam/capture--grch37/TCRBOA7-T-WEX.bam --referenceFasta=/projects/clc/clc_scratch/projects/hwinata/dev-modules/test/reference/genomes/grch37/genome_fasta/genome.fa --callRegions=results/strelka-1.0/01-chrom_bed/grch37.main_chroms.bed.gz --runDir=$(dirname results/strelka-1.0/02-strelka/capture--grch37/TCRBOA7-T-WEX--TCRBOA7-N-WEX--matched/runWorkflow.py) --indelCandidates=results/strelka-1.0/00-inputs/capture--grch37/vcf/TCRBOA7-T-WEX--TCRBOA7-N-WEX--matched.candidateSmallIndels.vcf.gz --exome > results/strelka-1.0/logs/launched-2020-06-29-at-13-55-13/02-strelka/capture--grch37/TCRBOA7-T-WEX--TCRBOA7-N-WEX-matched/strelka_configure.stdout.log 2> results/strelka-1.0/logs/launched-2020-06-29-at-13-55-13/02-strelka/capture--grch37/TCRBOA7-T-WEX--TCRBOA7-N-WEX--matched/strelka_configure.stderr.log
```

```
results/strelka-1.0/02-strelka/capture--grch37/TCRBOA7-T-WEX--TCRBOA7-N-WEX--matched/runWorkflow.py -m local --jobs 1 > results/strelka-1.0/logs/launched-2020-06-29-at-13-55-13/02-strelka/capture--grch37/TCRBOA7-T-WEX--TCRBOA7-N-WEX--matched/strelka_run.stdout.log 2> results/strelka-1.0/logs/launched-2020-06-29-at-13-55-13/02-strelka/capture--grch37/TCRBOA7-T-WEX--TCRBOA7-N-WEX--matched/strelka_run.stderr.log && rm -rf "$(dirname results/strelka-1.0/02-strelka/capture--grch37/TCRBOA7-T-WEX--TCRBOA7-N-WEX--matched/runWorkflow.py)/workspace/"
```

BrunoGrandePhD:
@whelena: For GitHub, your backticks (```) should be on their own lines, which gives this:

```
Lorem ipsum dolor sit amet, consectetur adipiscing elit. Proin convallis odio et maximus placerat. Phasellus facilisis ligula et leo sollicitudin, sed vehicula dui aliquet. Sed accumsan gravida nisi nec semper. 
```

Otherwise, you get this:

```Lorem ipsum dolor sit amet, consectetur adipiscing elit. Proin convallis odio et maximus placerat. Phasellus facilisis ligula et leo sollicitudin, sed vehicula dui aliquet. Sed accumsan gravida nisi nec semper. ```

**Comment by whelena** [comment]
oh okayy thank you!

**Comment by whelena** [comment]
I ran both with and without candidate indels outside of snakemake. I got the same error for with candidate indels and currently running the one without. It seems to be working since the error usually appear in the beginning of the run

**Comment by BrunoGrandePhD** [comment]
The following file doesn't exist. Could you clarify where the candidate small indel VCF file is?
```
/projects/clc/clc_scratch/projects/hwinata/dev-modules/test/results/strelka-1.0/00-inputs/capture--grch37/vcf/TCRBOA7-T-WEX--TCRBOA7-N-WEX--matched.candidateSmallIndels.vcf.gz
```

**Comment by whelena** [comment]
I deleted that when testing for no indels, it should be here: 
```
/projects/clc/clc_scratch/projects/hwinata/dev-modules/test/results/manta-2.1/99-outputs/vcf/capture--grch37/candidateSmallIndels/TCRBOA7-T-WEX--TCRBOA7-N-WEX--matched.candidateSmallIndels.vcf
```
or here for the .tbi file
```
/projects/clc/clc_scratch/projects/hwinata/dev-modules/test/results/manta-2.1/02-manta/capture--grch37/TCRBOA7-T-WEX--TCRBOA7-N-WEX--matched/results/variants/candidateSmallIndels.vcf.gz.tbi
```

**Comment by BrunoGrandePhD** [comment]
Why is the first file decompressed? It should be pointing to a bgzip-compressed file, like the one produced by manta. 
```
/projects/clc/clc_scratch/projects/hwinata/dev-modules/test/results/manta-2.1/02-manta/capture--grch37/TCRBOA7-T-WEX--TCRBOA7-N-WEX--matched/results/variants/candidateSmallIndels.vcf.gz
```
Make sure that the `.vcf.gz` and the `.vcf.gz.tbi` are both symlinked before passing the `.vcf.gz` file to Strelka.

**Comment by whelena** [comment]
Oh I think the unzipped file is from the augment_vcf rule. Oh wait no I think that's it, the vcf and the `vcf.gz.tbi` doesnt match because one goes through the `augment_vcf` rule

File: demo/config.yaml [hunk 28]
```diff
@@ -20,3 +20,9 @@ lcr-modules:
         inputs:
             sample_bam: "data/{sample_id}.bam"
             sample_bai: "data/{sample_id}.bam.bai"
+
+    strelka:
+        inputs:
+            sample_bam: "data/{sample_id}.bam"
+            sample_bai: "data/{sample_id}.bam.bai"
+            candidate_small_indels_vcf: "results/manta-2.0/99-outputs/vcf/{seq_type}--{genome_build}/candidateSmallIndels/{tumour_id}--{normal_id}--{pair_status}.candidateSmallIndels.vcf"
```
BrunoGrandePhD: I just realized that this file should be updated to:
```
results/manta-2.0/02-manta/{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/results/variants/candidateSmallIndels.vcf.gz
```

whelena: ya I think that's the issue, I'm rerunning it with the fixed one. Thank youu!

whelena: I replaced this and now the issue is that strelka couldn't figure out the dependency, and so I got the MissingInput error. I'm thinking of using `"results/manta-2.0/99-outputs/vcf/{seq_type}--{genome_build}/candidateSmallIndels/{tumour_id}--{normal_id}--{pair_status}.candidateSmallIndels.vcf"` as input but symlinking the original files still. Do you see a potential problem with this?

BrunoGrandePhD: I was under the impression that pointing to a file within an output directory would create that dependency link, but this and something I ran into yesterday has contradicted my impression. You could add the output directory as an input and the specific VCF file you want, but just don't use the directory in your command.

whelena: Using the manta output directory as an input results in a childIO error when you rerun using the same samples. I'm not sure why but that's why I used the final output instead.

whelena:
@BrunoGrandePhD Ready for merge, I fixed the issue with candidate small indels

demo/run_clc.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
BrunoGrandePhD: This file should be removed.

demo/config.yaml
```diff
@@ -20,3 +20,11 @@ lcr-modules:
         inputs:
             sample_bam: "data/{sample_id}.bam"
             sample_bai: "data/{sample_id}.bam.bai"
+
+    strelka:
+        inputs:
+            sample_bam: "data/{sample_id}.bam"
+            sample_bai: "data/{sample_id}.bam.bai"
+            candidate_small_indels_output: "results/manta-2.1/99-outputs/vcf/{seq_type}--{genome_build}/candidateSmallIndels/{tumour_id}--{normal_id}--{pair_status}.candidateSmallIndels.vcf"
```
BrunoGrandePhD: Can this line be removed now?

modules/strelka/1.0/strelka.smk
```diff
@@ -0,0 +1,256 @@
...
+    conda:
+        CFG["conda_envs"]["tabix"]
+    shell:
+        op.as_one_line("""
+        bgzip -c {input.bed} > {output.bedz}
+            &&
+        tabix {output.bedz}
+        """)
+
+def _get_indel_cli_arg(vcf_in = config["lcr-modules"]["strelka"]["inputs"]["candidate_small_indels_tbi"]):
```
BrunoGrandePhD: Since there are a few small edits needed before merging, would you prefix all module functions with the module name? Here, it would be `_strelka_get_indel_cli_arg`.

demo/config.yaml
```diff
@@ -20,3 +20,11 @@ lcr-modules:
...
+            candidate_small_indels_vcf: "results/manta-2.1/02-manta/{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/results/variants/candidateSmallIndels.vcf.gz"
+            candidate_small_indels_tbi: "results/manta-2.1/02-manta/{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/results/variants/candidateSmallIndels.vcf.gz.tbi"
```
BrunoGrandePhD: I thought you said that you can't directly point to the VCF files produced by Manta, no?

BrunoGrandePhD:
@whelena Please comment when the PR is ready for another round of review.

demo/config.yaml
whelena: No because I need to call the final output for candidatesmallindels, instead of the one generated by `_manta_run`. I think it's because the output is a directory so we'll get Missing InputException if `input.indels` is set to the vcf of tbi files that are in `02-manta`

modules/strelka/1.0/strelka.smk
whelena: okay!

demo/config.yaml
whelena: I can, I just can't use it as an input because it would trigger a `MissingInputException`

whelena:
I completely missed checking on this module but I have deleted `run_clc.sh` and I think the rest are ready for merging

rdmorin:
What's in the way of this being finalized, @whelena ? Do you need me to have a look at something specific?

whelena:
This is finalized, I was waiting for a final review before merging.

rdmorin:
I don't know how to allow this to be merged this because @BrunoGrandePhD requested changes and merging is blocked. Were those addressed?

whelena:
ya I addressed it and I think I can still merge it without his approval because I have admin permission.

PR #41, opened 2020-06-07
Module/sequenza/1.0

rdmorin:
# Pull Request Checklists
### Required

Not yet. I'm new to this concept and could use some more guidance. 
Genome test was a success

- [ ] I labelled every value that should (or might need to) be updated in the default configuration file with `UPDATE` comments.

  ```bash
  ```

### If applicable


BrunoGrandePhD:
Do you prefer that I comment with suggested improvements? Or are you okay with me making commits to this branch and then summarize my changes (with rationales) in a comment here?

rdmorin:
Feel free to make commits with rationale. Thanks

modules/sequenza/1.0/sequenza.smk
```diff
@@ -0,0 +1,226 @@
...
+rule _sequenza_unfiltered_analysis:
+    input:
+        merged_seqz = rules._sequenza_merge_seqz.output.merged_seqz,
+        normal_bam = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{normal_id}.bam"
+    output:
+        segments = CFG["dirs"]["plots"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/{tumour_id}_unfiltered_segments.txt"
+    log:
+        stdout = CFG["logs"]["plots"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/{tumour_id}_sequenza_unfiltered_analysis.stdout.log",
+        stderr = CFG["logs"]["plots"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/{tumour_id}_sequenza_unfiltered_analysis.stderr.log"
+    resources: mem_mb = 100000
```
BrunoGrandePhD: Does this step really need 100 GB of memory?

BrunoGrandePhD: I notice that `_sequenza_filtered_analysis` has the same thing.

rdmorin: Doubtful I'll test it with a genome and exome to see what is reasonable.

rdmorin: Peak memory usage was less than 20G for the single genome I tested. As long as we filter the SNVs, it shouldn't be worse for FFPE. The unfiltered analysis could be much worse for FFPE. I think we should reduce this to 20G and see how it goes. We can up it or do something like the following if necessary:

mem_mb=lambda wildcards, attempt: attempt * 20000

BrunoGrandePhD:
## Summary of my changes

This is still a work-in-progress. 

- Removed print statements. 
- Revised default `mem_mb` amounts and moved all of them to the default config file. 
- Added `threads` (even if equal to 1) for every non-local rule and moved all of them to the default config file. 
- Removed params that only contained a single wildcard. You can refer to these in `shell` using `{wildcards.tumour_id}`, for example. 
- Wrapped long commands using `oncopipe.as_one_line()`. 
- Removed any `mkdir` since snakemake automatically creates the output directories. 
- Created a `seqz` output subdirectory where the unfiltered and filtered SEQZ files will be stored. 
- Simplified output and log file names. 
- Moved `calc_sex_status` functionality into `run_sequenza` script. 
- Moved scripts from params to input files. 
- Created separate environments for the sequenza-utils Python package. 
- Stored these environment YAML files in the shared `envs/` and created symlinks. 
- Created rule in `reference_files` workflow to create GC WIG file. 
- (WIP) Collapsed filtered and unfiltered rules. 
- (WIP) Use chromosome lists and genome build versions from `reference_files` workflow. 
- (WIP) Created rule to download dbSNP VCF from NCBI and derive POS version. 
- (WIP) Created conda package using forked version of the `copynumber` R package with support for hg38. 
- (WIP) Created separate environments for the sequenza R package.

rdmorin:
Out of interest, if we always use Sequenza and sequenza-utils in the same workflow, what is the rationale for two separate conda environments?

BrunoGrandePhD:
In general, I've moved towards having package-focused environments rather than module-focused environments. The main reason is that packages can be versioned independently of one another, and the more specific environments can be re-used in different contexts. For example, I'm planning on adding rules in the `reference_files` workflow that will dynamically generate the GC content file from the genome FASTA file using `sequenza-utils`. 

Granted, the `sequenza` R package and `sequenza-utils` Python package are generally developed and versioned in sync by the developers, but we can imagine a situation where they update both to a new version, but one of the two packages has a bug. If the two packages are in separate environments, one can easily be updated without the other. 

Another reason to have package-focused environments is to simplify the process of debugging conda environments that don't work anymore (_e.g._ a random package no longer exists). If the environment is named after a specific package and version (_e.g._ `sequenza-utils-3.0.0`), it's self-documenting in that it's which package within the YAML file need to be set to a specific version. On the other hand, if you  mix different packages in an environment, you need another file (like a README) to list which packages need to specifically be versioned (_e.g._ `sequenza=3.0.0` and `sequenza-utils=3.0.0`).

rdmorin:
Does this need a lot more work still @BrunoGrandePhD ? I was hoping to deploy on GAMBL so I could test out the new header.smk and generate some results from some cases so I can compare the results to Battenberg. I also am waiting on submitting a PR for my Battenberg module because I don't want to repeat the same mistakes.

BrunoGrandePhD:
This does need more work. I need another few hours to wrap this up. I was busy this past weekend, so this will have to wait until this weekend. It'll be worth it!

rdmorin:
No problem. I can keep using my local branch.

BrunoGrandePhD:
I was able to implement the rest of the features I was hoping to get to (see below). I'm doing a test run with the demo project now, so I'll know if everything works tomorrow. A dry run was successful though. 

- Collapsed filtered and unfiltered rules. 
- Use chromosome lists and genome build versions from `reference_files` workflow. 
- Created rule to download dbSNP VCF from NCBI and derive POS version. 
- Created separate environments for the sequenza R package. 
- Force install forked copynumber from within R script

rdmorin:
This is great. Thanks Bruno!

rdmorin:
Are you able to resolve this one conflict, @BrunoGrandePhD ?

BrunoGrandePhD:
Yeah, I'll be able to resolve that conflict. I'm currently debugging the Sequenza R script. Everything up until this point, including the updated `reference_files` workflow, seems to be working.

BrunoGrandePhD:
I don't think my method for inferring sex from the depth ratio works for exomes. For the test data, the patient is male, but the ratio is 0.858, which is higher than 0.75 and thus interpreted as best corresponding to an XX genotype. My method also doesn't work on tumours with unmatched normal samples. 

Because I don't want to delay this any further, I think we can post the current version while only supporting tumour genomes with matched normal samples. In a future version, we can add support for a `sex` column in the sample table. What do you think?

rdmorin:
Sounds like a good strategy

BrunoGrandePhD:
Alright, let me do one more test with a few BL (hg38) samples to see if everything works on genomes from another genome build. 🤞

BrunoGrandePhD:
BL test running now. If all goes well, the new `reference_files` workflow will auto-generate the dbSNP VCF (auto-converting from chr-unprefixed to chr-prefixed) and the GC content WIG file. I also updated the list of "main chromosomes" to exclude chrY and chrM because these generally pose more problems than they're worth. Notice that the X chromosome and the genome version (`grch37` and `grch38`) are also generated for each genome build. They're silly one-line files, but they ensure the portability of the sequenza R script. If this works, it will have validated all of the time I reviewing on this module and will hopefully make running it on GAMBL much easier. 

```
Job counts:
	count	jobs
	1	calc_gc_content
	1	download_chromosome_x
	1	download_dbsnp_vcf
	1	download_main_chromosomes
	1	get_dbsnp_download
	1	get_main_chromosomes_download
	3	hardlink_download
	1	store_genome_build_info
	3	update_contig_names
	13
```

rdmorin:
Great. Thanks Bruno!

modules/sequenza/1.0/sequenza.smk
```diff
@@ -0,0 +1,240 @@
...
+
+
+rule _sequenza_bam2seqz:
+    input:
+        tumour_bam = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{tumour_id}.bam",
+        normal_bam = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{normal_id}.bam",
+        gc_wiggle = reference_files("genomes/{genome_build}/annotations/gc_wiggle.window_50.wig.gz"),
+        genome = reference_files("genomes/{genome_build}/genome_fasta/genome.fa")
+    output:
+        seqz = CFG["dirs"]["seqz"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/chromosomes/{chrom}.binned.seqz.gz"
```
rdmorin: wrap chromosome seqz in (temp) so they aren't retained unnecessarily?

modules/sequenza/1.0/src/python/cnv2igv.py
```diff
@@ -0,0 +1,232 @@
+import argparse
```
rdmorin: This file is now not necessary because it is in lcr-scripts, right?

BrunoGrandePhD:
I was able to mostly get this module to run on some of my BL data. However, I'm running into an issue where the conda environment for `run_sequenza` isn't being loaded properly. It keeps defaulting to the GSC R installation. Is this an issue that you experienced?

BrunoGrandePhD:
@whelena: Could you test this module on some of your data?

whelena:
@BrunoGrandePhD I'm getting this error
```
Missing input files for rule all:
/projects/clc/clc_scratch/projects/hwinata/dev-modules/test/reference/genomes/grch37/annotations/gc_wiggle.window_50.wig.gz
```
I couldn't find a rule `reference_files` workflow that generates this file

rdmorin:
I have not encountered that issue.

BrunoGrandePhD:
Are you using `reference_files` version 2.0? It's included in this PR. 

https://github.com/LCR-BCCRC/lcr-modules/pull/41/files#diff-dfbd79f4aead03d64d5a2e750a2b9527R135-R150

rdmorin:
That is unlikely. I think I would only be using code that has been merged to master.

BrunoGrandePhD:
Sorry, that was in response to @whelena's issue.

whelena:
@BrunoGrandePhD oh whoops sorry I didnt realize that, I'll retry it

UPDATE: 
I fixed the error by removing `genomes/grch37/annotations/gc_wiggle.window_50.wig.gz` from the `_sequenza_all` rule.

whelena:
@BrunoGrandePhD 
```
Warning message:
There are fewer than 1 million positions with depth data on the X chromosome. This usually means the input data was based on exomes, which will fail the sex inferrence step.
Inferring this patient as female.
Error in seq.default(0, max_coord_chr_i, by = 1e+07) :
  'to' must be a finite number
Calls: sequenza.results -> axis -> seq -> seq.default
```
I get this error message when running with `TCRBOA7-T-WEX` and `TCRBOA7-N-WEX`

BrunoGrandePhD:
This is probably caused by the inclusion of chromosome Y and MT in the analysis. I updated the `reference_files` workflow to exclude these two small chromosomes from the "main_chromosomes" file. Can you check that the `main_chromosomes.txt` file being used excludes these small chromosomes? If not, you can delete the `main_chromosomes` files to make sure they're regenerated with `reference_files` 2.0. 

```
# Add -delete at the end of the following command once you confirm that it finds the right files
find /path/to/reference_files/ -name "*main_chromosomes*"
```

whelena:
Thank you! It runs to completion now!

rdmorin:
Is this ready to merge?

Issue #32, opened 2020-05-01
Modutils: Subset sample table by exclusion criteria


lkhilton:
The `filter_samples()` function in `modutils` is currently able to filter the sample metadata based on things to _include_. Could you please implement the ability to specify samples to _exclude_?

BrunoGrandePhD:
Note to self: Add mention of `filter_samples()` to the documentation and demo Snakefile.

BrunoGrandePhD:
I've implemented this in the upcoming `bug-fixes` branch. Since the name `filter_samples()` is ambiguous, I added an `invert` argument to determines whether you are positively or negatively selecting samples. As shortcuts, you can also use the more appropriately named `keep_samples()` and `discard_samples()` wrapper functions. Here's how they work: 

```
>>> import oncopipe as op
Warning: The oncopipe package was imported outside of a snakefile. Most functions are designed to work within a snakefile. Some unexpected behaviour/errors might occur.

>>> samples = op.load_samples("../lcr-modules-dev/demo/data/samples.tsv")

>>> samples
       sample_id seq_type patient_id tissue_status genome_build
0  TCRBOA7-N-WEX  capture    TCRBOA7        normal       grch37
1  TCRBOA7-T-WEX  capture    TCRBOA7        tumour       grch37
2  TCRBOA7-T-RNA     mrna    TCRBOA7        tumour       grch37

>>> op.keep_samples(samples, seq_type="mrna")
       sample_id seq_type patient_id tissue_status genome_build
2  TCRBOA7-T-RNA     mrna    TCRBOA7        tumour       grch37

>>> op.discard_samples(samples, seq_type="mrna")
       sample_id seq_type patient_id tissue_status genome_build
0  TCRBOA7-N-WEX  capture    TCRBOA7        normal       grch37
1  TCRBOA7-T-WEX  capture    TCRBOA7        tumour       grch37
```

PR #48, opened 2020-06-15
Minor feature additions and bug fixes

BrunoGrandePhD:
In this PR, I've updated the code and docs to expect `__UPDATE__` strings in the default configuration files for anything that required user intervention. Before, the module developer was expected to annotate the values that needed to be updated using `UPDATE` comments. The advantage of using `__UPDATE__` is that I can search for them when setting up the module and raise an informative error regarding a placeholder that needs to be updated. 

I also added two convenience functions for filtering samples to address #32.


BrunoGrandePhD:
@whelena: Can you think of any way the new approach with `__UPDATE__` can be problematic?

whelena:
> @whelena: Can you think of any way the new approach with `__UPDATE__` can be problematic?

@BrunoGrandePhD i can't see it being a problem

BrunoGrandePhD:
Okay! In that case, feel free to approve the PR and merge it with master. Thanks!

BrunoGrandePhD:
Thanks!

PR #56, opened 2020-06-25
Module/varscan/1.0

whelena:
# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable

To be completed.


BrunoGrandePhD:
I'll tackle this on the weekend. Thanks, Helena!

BrunoGrandePhD:
I'm going to defer to others for this PR since I'm already juggling a few other PRs. Ping me with my username (@BrunoGrandePhD) if ever you need my help!

rdmorin:
I'll give it a try

rdmorin:
I'm a bit confused. I don't think I've seen the vcf2maf module. Should it be reviewed before this one or does it matter.

BrunoGrandePhD:
I've reviewed multiple modules in one PR before. Sometimes, it's easier to do it this way since they're interrelated.

rdmorin:
OK, so this one actually has both as a part of it? I'm kinda lost how to review these when each commit only shows things partially in the context of the full code. Do you have any nice links with an explanation of how to review a PR effectively?

modules/varscan/1.0/varscan.smk
```diff
@@ -0,0 +1,194 @@
...
+# Setup module and store module-specific configuration in `CFG`
+# `CFG` is a shortcut to `config["lcr-modules"]["varscan"]`
+CFG = op.setup_module(
+    name = "varscan",
+    version = "1.0",
+    subdirectories = ["inputs", "mpileup", "varscan", "maf","outputs"]
+)
+
+# Define rules to be run locally when using a compute cluster
+# TODO: Replace with actual rules once you change the rule names
```
rdmorin: Reminder to remove TODO lines

modules/varscan/1.0/varscan.smk
```diff
@@ -0,0 +1,194 @@
...
+        bam = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{sample_id}.bam"
+    run:
+        op.relative_symlink(input.bam, output.bam)
+        op.relative_symlink(input.bam + ".bai", output.bam + ".bai")
+
+
+rule _varscan_bam2mpu:
+    input:
+        bam = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{sample_id}.bam",
+        fasta = reference_files("genomes/{genome_build}/genome_fasta/genome.fa")
```
rdmorin: Does samtools mpileup require a .fai file? (I've forgotten) if so, it is probably better to explicitly specify it as an input

modules/varscan/1.0/varscan.smk
```diff
@@ -0,0 +1,194 @@
...
+        samtools mpileup {params.opts}
+        -f {input.fasta} {input.bam}
+        > {output.mpu}
+        2> {log.stderr}
+        """)
+
+
+rule _varscan_somatic:
+    input:
+        tumour_mpu = CFG["dirs"]["mpileup"] + "{seq_type}--{genome_build}/{tumour_id}.mpileup",
```
rdmorin: How big are these files from WGS data? These might be so large that this will require a pipe or combining two rules into one with piped commands.

modules/varscan/1.0/varscan.smk
```diff
@@ -0,0 +1,194 @@
...
+        {params.opts}
+        > {log.stdout} 2> {log.stderr}
+        """)
+
+
+rule _varscan_unpaired:
+    input:
+        tumour_mpu = CFG["dirs"]["mpileup"] + "{seq_type}--{genome_build}/{tumour_id}.mpileup",
+    output:
+        vcf = CFG["dirs"]["varscan"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/{vcf_name}.vcf"
```
rdmorin: Would it be a problem downstream if these were gzipped as they were created?

modules/varscan/1.0/varscan.smk
```diff
@@ -0,0 +1,194 @@
...
+        2> {log.stderr}
+        """)
+
+
+rule _varscan_somatic:
+    input:
+        tumour_mpu = CFG["dirs"]["mpileup"] + "{seq_type}--{genome_build}/{tumour_id}.mpileup",
+        normal_mpu = CFG["dirs"]["mpileup"] + "{seq_type}--{genome_build}/{normal_id}.mpileup"
+    output:
+        snp = CFG["dirs"]["varscan"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/snp.vcf",
```
rdmorin: Can these be gzipped as they are created?

BrunoGrandePhD:
You ask, and you shall receive! Here's a video by yours truly on reviewing PRs in GitHub: https://youtu.be/EQugWjCkoC4

I generally use the Files Changed tab, because I care about the difference relative to `master` and I don't care about how those differences are spread out in individual commits. 

There is an overwhelming amount of new files in this PR, so you might want to start the review by simply running testing the new modules. If you run into any bugs, you can post the command and working directory here for @whelena to debug. If you are so included, you can take a stab at debugging yourself, but by default, I think that onus is on the PR submitter, not the reviewer.

modules/varscan/1.0/varscan.smk
```diff
@@ -0,0 +1,194 @@
...
+        op.as_one_line("""
+        varscan mpileup2{wildcards.vcf_name} 
+        {input.tumour_mpu} 
+        {params.opts}
+        {params.cns}
+        > {output.vcf} 2> {log.stderr}
+        """)
+
+
+rule _varscan_symlink_maf:
```
rdmorin: Should this be called _varscan_symlink_vcf?

rdmorin: Actually, what does this do? Seems redundant with the rule below

modules/vcf2maf/1.0/config/default.yaml
```diff
@@ -0,0 +1,21 @@
...
+            vcf2maf: "--filter-vcf 0"
+            # --filter-vcf     A VCF for FILTER tag common_variant. Set to 0 to disable [~/.vep/ExAC_nonTCGA.r0.3.1.sites.vep.vcf.gz]
+            #--species        Ensembl-friendly name of species (e.g. mus_musculus for mouse) [homo_sapiens]
+            #--cache-version  Version of offline cache to use with VEP (e.g. 75, 84, 91) [Default: Installed version]
+
+        conda_envs:
+            vcf2maf: "envs/vcf2maf-1.6.18.yaml"
+            
+        threads:
+            vcf2maf: 2
```
rdmorin: Any reason not to set this higher by default? I usually use at least 12 threads.

rdmorin:
Thanks!

________________________________

demo/config.yaml
```diff
@@ -21,3 +21,7 @@ lcr-modules:
         inputs:
             sample_bam: "data/{sample_id}.bam"
             sample_bai: "data/{sample_id}.bam.bai"
+
+    varscan:
+        inputs:
+            sample_bam: "data/{sample_id}.bam"
```
rdmorin: Why is the .bai file not an input?

modules/varscan/1.0/config/default.yaml
```diff
@@ -0,0 +1,46 @@
+lcr-modules:  
+    varscan:
+        inputs:
+            # Available wildcards: {seq_type} {genome_build} {tumour_id} {normal_id}
+            sample_bam: "__UPDATE__"
```
rdmorin: Shouldn't sample_bai virtually always also be an input even if not explicitly specified in the shell command?

modules/varscan/1.0/config/default.yaml
```diff
@@ -0,0 +1,46 @@
...
+                genome: "--output-vcf"
+                mrna: "--output-vcf"
+            unpaired:
+                capture: "--min-coverage 20 --min-reads2 10 --min-avg-qual 20 --min-var-freq 0.01  --p-value 0.05 --strand-filter 0 --output-vcf 1 --min-strands2 0"
+                genome: "--output-vcf"
+                mrna: "--output-vcf"
+
+
+        conda_envs:
+            samtools: "{MODSDIR}/envs/samtools-1.9.yaml"
```
rdmorin: Seems redundant to specify the same yaml twice but not a big deal

modules/varscan/CHANGELOG.md
```diff
@@ -0,0 +1,17 @@
...
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [1.0] - 2020-06-22
+
+This release was authored by Helena Winata.
+
+
+- The `varscan` module supports somatic variant calling for paired tumour-normal samples and germline variant calling for unpaired (`no_normal`) samples.
+- Command line parameters for somatic and germline calls are specified for each `seq-type`.
+- Intermediate mpileup files are temporary due to file size.
```
rdmorin: how big are they?

modules/vcf2maf/1.0/vcf2maf.smk
```diff
@@ -0,0 +1,67 @@
...
+    "GRCh38": "GRCh38",
+}
+
+##### RULES #####
+
+rule _vcf2maf_run:
+    input:
+        vcf = "{out_dir}/{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/{vcf_name}.vcf",
+        fasta = reference_files("genomes/{genome_build}/genome_fasta/genome.fa"),
+        vep_cache = CONFIG["inputs"]["vep_cache"]
```
rdmorin: It looks like this assumes the user installed the cache themselves. Why did we do it this way instead of using my auto-install approach?

rdmorin:
I haven't tested this yet but wanted some general feedback based on a few of my comments first. My main concern right now is the temporary disc space that will be consumed by the output of mpileup from two separate genome bams.

modules/varscan/1.0/varscan.smk
whelena: I will symlink the .fai as inputs too

modules/varscan/1.0/varscan.smk
whelena: They're pretty large, would `pipe` or `temp` be better? In the past I always used `temp` but I think there's a memory issue because snakemake tries to generate all the mpileup files at the same time. When using `pipe` are the intermediate files (.mpileup) kept in the end?

modules/varscan/1.0/varscan.smk
whelena: I'm not sure I can try that as well

modules/varscan/1.0/varscan.smk
whelena: This part is a link to the `vcf2maf` module. I wrote the `vcf2maf` module similar to the `utils` module where it's just a collection of rules. It allows it to be used by multiple modules without making the code too complicated. 

So `_varscan_symlink_maf` symlinks the vcf file into a directory (`CFG["dirs"]["maf"]`), where vcf2maf is  executed.

modules/vcf2maf/1.0/config/default.yaml
whelena: No I'm working with small datasets so I thought setting it to 12 threads would increase the wait time for resources. I can change the default to 12 in the final commit

demo/config.yaml
whelena: I'm going to add that as an input, right now I'm assuming that the .bam and .bam.bai files are generated by the pipeline and I used 
```
op.relative_symlink(input.bam + ".bai", output.bam + ".bai")
```
in the `_varscan_input_bam`

modules/varscan/CHANGELOG.md
whelena: for `TCRBOA7-N-WEX ` and `TCRBOA7-T-WEX` from the demo file it's about 50Gb and it can take up to 500Gb for genome bams.

modules/vcf2maf/1.0/vcf2maf.smk
whelena: I haven't implemented the auto-install approach yet since I wasn;t sure if it should be part of the `reference_files` subworkflow instead

whelena:
@rdmorin Thank you for the extensive feedback, I will go through them in more detail. I will work on the edits before you start testing. I'm also using this PR for vcf2maf as I tested them together, but I'm thinking of separating vcf2maf as it's own PR.

modules/varscan/1.0/varscan.smk
whelena: oh thank you!

rdmorin:
Please do go ahead and split this PR into one for vcf2maf and one for varscan.

modules/varscan/1.0/varscan.smk
whelena: I decided to go for tempt because I have a `_varscan_somatic` and `_varscan_unpaired` rule which uses the output of the `_varscan_bam2mpu` rule.  For pipe we can only have one rule depending on the output. Let me know if there's you have suggestions on how to get around this.

modules/varscan/1.0/varscan.smk
whelena: I'm wondering if we should keep the output consistent (.vcf or .vcf.gz) between variant caller modules, since it would simplify the vcf2maf module. I think vcf2maf only takes non-gsipped vcf files, but correct me if I'm wrong.

rdmorin: I haven't tried it on gzipped directly but the module should be able to decompress them temporarily.

rdmorin: Come to think of it, they should probably be "bgzip2" compressed (which is what Strelka2 does by default)

rdmorin:
I have overhauled the varscan module and disabled some of the functionality. @whelena, can you have a look and see if you can roll the parallelization into the unpaired mode similar to what I did for paired (somatic) variant calling?

demo/config.yaml
rdmorin: It's better to leave .bai as an input because then snakemake can raise errors early if they're missing.

modules/varscan/1.0/header_hg19.txt
```diff
@@ -0,0 +1,43 @@
+##fileformat=VCFv4.1
```
rdmorin: We will need to run this on an hg38 genome and supply an equivalent file for hg38 bams. We will also need a way to choose the appropriate file. I'm sure @BrunoGrandePhD will have thoughts on how best to handle this.

whelena: we could use the `op.switch_on_wildcard` function to pick the right header depending on the genome build

rdmorin: Good idea

whelena: are the headers also different for somatic vs. unpaired runs?

rdmorin: Good point. It probably is at least slightly different. You'd have to run it and see what the header looks like

modules/vcf2maf/1.0/config/default.yaml
rdmorin: Reminder to update the thread number

whelena:
@rdmorin This module is essentially ready to merge. The only thing missing are the header files. Is the goal of the reheader rule to add the contig ids and lengths?

rdmorin:
It seems like I can't allow this to be merged either. It says "2 pending reviewers"

whelena:
I think I can technically merge any pull request. I can merge this too with your approval. I removed the pending reviewers, so I think now you can merge it.

rdmorin:
Go for it

rdmorin:
Correct


On Jul 23, 2020, at 1:10 PM, whelena <notifications@github.com> wrote:

﻿

@rdmorin<https://github.com/rdmorin> This module is essentially ready to merge. The only thing missing are the header files. Is the goal of the reheader rule to add the contig ids and lengths?

rdmorin:
> @rdmorin This module is essentially ready to merge. The only thing missing are the header files. Is the goal of the reheader rule to add the contig ids and lengths?

Yes

whelena:
> > @rdmorin This module is essentially ready to merge. The only thing missing are the header files. Is the goal of the reheader rule to add the contig ids and lengths?
> 
> Yes

is there a way of adding it without changing the entire header? because I think with the current approach, the unpaired  vcfs will be reheadered with a paired vcf header

rdmorin:
@whelena can we pick this up on Slack?

Issue #35, opened 2020-05-14
Implement support for sample-based modules for `op.switch_on_column()`


BrunoGrandePhD:

BrunoGrandePhD:
I could replace the `match_on` argument with a more general `wildcard_name` argument, which would allow the user to specify `"sample_id"`, `"tumour_id"`, or `"normal_id"`.

whelena:
@BrunoGrandePhD I implemented it into some of my modules and pushed the edit to master. Would the new edits be installed using `pip install oncopipe` or do users have to use `pip install -e lcr-modules/oncopipe`?

BrunoGrandePhD:
I created a new release on GitHub (v1.0.7), which automatically deploys to PyPI. 

https://github.com/LCR-BCCRC/lcr-modules/releases
https://pypi.org/project/oncopipe/

PR #57, opened 2020-06-25
Module/picard qc/1.0

whelena:
# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable

To be completed.


BrunoGrandePhD:
I'm excited to review this one! Another one to test run on the weekend.

demo/Snakefile
```diff
@@ -38,14 +40,16 @@ configfile: "config.yaml"
 
 # Use all samples as a default sample list for each module
 config["lcr-modules"]["_shared"]["samples"] = SAMPLES
-
+config["lcr-modules"]["picard_qc"]["samples"] = SAMPLES
```
BrunoGrandePhD: Why is this necessary? Shouldn't the module inherit from `_shared` if `samples` in `picard_qc` is left unset?

modules/picard_qc/1.0/config/default.yaml
```diff
@@ -0,0 +1,58 @@
...
+    
+    picard_qc:
+        inputs:
+            # Available wildcards: {seq_type} {genome_build} {sample_id}
+            sample_bam: "__UPDATE__"
+            intervals : "__UPDATE__"
+
+            # by default are generate by reference_files subworkflow
+            rrna_intervals: "genomes/{genome_build}/rrna_intervals/rRNA_int_gencode-33.txt"
+            refFlat: "genomes/{genome_build}/annotations/refFlat_gencode-33.txt"
```
BrunoGrandePhD: For the `star` module, I decided against including these in the config file. Because these files are tied to the `reference_files` workflow, they can't really be changed, so there's no point in including the paths here. In fact, I'm afraid the user will try to change them without realizing that they're wrapping in `reference_files()` in the Snakefile. Instead, you can use `reference_params` like I did in `star`.

modules/picard_qc/1.0/config/default.yaml
```diff
@@ -0,0 +1,58 @@
...
+        options:
+            alignment_summary: "VALIDATION_STRINGENCY=LENIENT"
+            insert_size: "VALIDATION_STRINGENCY=LENIENT"
+            hs_metrics: "VALIDATION_STRINGENCY=LENIENT LEVEL=ALL_READS CLIP_OVERLAPPING_READS=false"
+            rnaseq_metrics:
+                base: "VALIDATION_STRINGENCY=LENIENT"
+                strand:
+                    _default: "NONE"
+                    forward: "FIRST_READ_TRANSCRIPTION_STRAND"
+                    reverse: "SECOND_READ_TRANSCRIPTION_STRAND"
```
BrunoGrandePhD: You can put this under `switches` like I did in `manta` 2.1. Not a required change though. Just a suggestion for consistency.

modules/picard_qc/1.0/envs/picard-2.22.3.yaml
```diff
@@ -0,0 +1 @@
+/projects/clc/clc_scratch/projects/hwinata/dev-modules/lcr-modules/envs/picard/picard-2.22.3.yaml
```
BrunoGrandePhD: This should be a relative symlink

modules/picard_qc/1.0/picard_qc.smk
```diff
@@ -0,0 +1,323 @@
...
+    _picard_qc_all
+
+
+##### RULES #####
+
+
+# Symlinks the input files into the module results directory (under '00-inputs/')
+rule _picard_qc_input_bam:
+    input:
+        bam = CFG["inputs"]["sample_bam"]
```
BrunoGrandePhD: Are BAM indices required? If so, I would follow the same approach as `manta` 2.1. Otherwise, if the user is generating the BAM files themselves using custom rules and the BAM index in a separate rule, this won't trigger the latter.

modules/picard_qc/1.0/picard_qc.smk
```diff
@@ -0,0 +1,323 @@
...
+        I={input.bam} O={output.metrics} H={output.histogram}
+        > {log.stdout} 2> {log.stderr}
+        """)
+
+# capture metrics
+rule _picard_qc_hs_metrics:
+    input:
+        bam = rules._picard_qc_input_bam.output.bam,
+        fasta = reference_files("genomes/{genome_build}/genome_fasta/genome.fa"),
+        intervals = CFG["inputs"]["intervals"]
```
BrunoGrandePhD: Because it's so easy to generate an interval_list file from a BED file and BAM file, could we not include a rule for this here since we already are using picard? See example command below:

```
picard BedToIntervalList I=capture_regions.bed O=capture_regions.interval_list SD=BLGSP-71-06-00080-01A-11D.bam
```

modules/picard_qc/1.0/picard_qc.smk
```diff
@@ -0,0 +1,323 @@
...
+        fasta = reference_files("genomes/{genome_build}/genome_fasta/genome.fa"),
+        refFlat = reference_files(CFG["inputs"]["refFlat"])
+    output:
+        metrics = CFG["dirs"]["metrics"] + "{seq_type}--{genome_build}/{sample_id}/rnaseq_metrics"
+    log:
+        stdout = CFG["logs"]["metrics"] + "{seq_type}--{genome_build}/{sample_id}/rnaseq_metrics.stdout.log",
+        stderr = CFG["logs"]["metrics"] + "{seq_type}--{genome_build}/{sample_id}/rnaseq_metrics.stderr.log"
+    params:
+        opts = CFG["options"]["rnaseq_metrics"]["base"],
+        strand = op.switch_on_column("strand", CFG["samples"], CFG["options"]["rnaseq_metrics"]["strand"], match_on = "sample")
```
BrunoGrandePhD: First, good discovery of the `match_on` parameter!

Second, you're going to want to create a new schema file in the root `schemas/` directory and symlink it into the module's `schemas/` directory. You should probably make it an enum with only three (?) options: forward, reverse, or unstranded. Whatever terms you choose, update your switch accordingly.

modules/picard_qc/1.0/picard_qc.smk
```diff
@@ -0,0 +1,323 @@
...
+                    if i == 0:
+                        header = "SAMPLEID\t" + "\t".join(data[0].split("\t")[0:])
+                        out.write(header)
+                    if wildcards.metrics == "alignment_summary_metrics":
+                        line = s_id + "\t" + "\t".join(data[3].split("\t")[0:])
+                    else:
+                        line = s_id + "\t" + "\t".join(data[1].split("\t")[0:])
+                    out.write(line)
+                    f.close()
+        out.close()
```
BrunoGrandePhD: A lot is going on here. Could you briefly explain what's happening here?

BrunoGrandePhD: I wonder if the `run` portion can be captured as a bash script to avoid loading stuff into Python.

modules/utils/2.0/envs/samtools-1.9.yaml
```diff
@@ -0,0 +1,28 @@
+name: test-samtools
```
BrunoGrandePhD: Shouldn't this be a symlink?

modules/utils/2.0/envs/sambamba-0.7.1.yaml
```diff
@@ -0,0 +1,29 @@
+name: test-sambamba
```
BrunoGrandePhD: Shouldn't this be a symlink?

modules/utils/2.0/config/default.yaml
```diff
@@ -0,0 +1,31 @@
+## Lines commented out with `#!` are required for the module to run
+## Lines commented out with `#?` can optionally be user-configured
+## Lines commented out with `##` act as regular comments
```
BrunoGrandePhD: You can remove this now that we've moved to using `__UPDATE__`

modules/utils/2.0/config/default.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+lcr-modules:
+    utils:
+        inputs:
+            bed: "__UPDATE__"
```
BrunoGrandePhD: I'm not sure what this BED file is supposed to be. We might have to use a more specific name.

modules/utils/2.0/utils.smk
```diff
@@ -0,0 +1,171 @@
...
+        {input.bam} 
+        > {log.stdout}
+        2> {log.stderr}
+        """)
+
+
+rule: # create_interval_list from bed; default exomes
+    input:
+        bed = CONFIG["inputs"]["bed"],
+        seq_dict = reference_files("genomes/{genome_build}/genome_fasta/genome.dict")
```
BrunoGrandePhD, line 122: Should you be using `reference_files()` from within the `reference_files` workflow?

modules/utils/2.0/utils.smk
```diff
@@ -0,0 +1,171 @@
...
+    output: 
+        intervals = "reference/exomes/{genome_build}/interval/{id}_intervals.txt"
```
BrunoGrandePhD, line 124: What is the `id` wildcard supposed to be?

modules/utils/2.0/utils.smk
```diff
@@ -0,0 +1,171 @@
...
+        I={input.bed}
+        O={output.intervals}
+        SD={input.seq_dict}
+        2> {log}
+        &&
+        chmod a-w {output.intervals}
+        """)
+
+
+del CONFIG
```
BrunoGrandePhD: If you want, you can name the config `_UTILS` and avoid having to delete it.

oncopipe/oncopipe/__init__.py
```diff
@@ -83,7 +83,6 @@ def set_config(self, config):
```
BrunoGrandePhD: I noticed that this file has had a bunch of empty lines removed. Could this be reverted?

oncopipe/oncopipe/__init__.py
```diff
@@ -420,6 +403,8 @@ def _switch_on_column(
             sample_id = wildcards.tumour_id
         elif match_on == "normal":
             sample_id = wildcards.normal_id
+        elif match_on == "sample":
+            sample_id = wildcards.sample_id
```
BrunoGrandePhD: Ha! It looks like you addressed #35. Thanks!

workflows/reference_files/1.0/config/default.yaml
```diff
@@ -1,3 +1,5 @@
+reference_directory: "reference/"
```
BrunoGrandePhD: Where is this required? Were you running into an error if this was omitted?

BrunoGrandePhD:
I ran into the error below while running the following command. Maybe this has to do with the change I made to the `_get_sample_metrics` function, although I don't see how. I thought I left the logic intact. 
```
# Working directory: /home/bgrande/repos/adult_blgsp
nice snakemake --profile gphosts --cores 140 _picard_qc_all
```

Error:
```
MissingInputException in line 218 of /home/bgrande/repos/adult_blgsp/src/lcr-modules/modules/picard_qc/1.0/picard_qc.smk:
Missing input files for rule _picard_qc_merge_metrics:
results/picard_qc-1.0/01-metrics/mrna--hg38/BLGSP-71-19-00116-01A-01R/alignment_summary_metrics
results/picard_qc-1.0/01-metrics/mrna--hg38/BLGSP-71-06-00288-01A-01R/alignment_summary_metrics
results/picard_qc-1.0/01-metrics/mrna--hg38/BLGSP-71-29-00529-50A-01R/alignment_summary_metrics
results/picard_qc-1.0/01-metrics/mrna--hg38/BLGSP-71-08-00194-01A-11R/alignment_summary_metrics
results/picard_qc-1.0/01-metrics/mrna--hg38/BLGSP-71-27-00420-01A-01S/alignment_summary_metrics
results/picard_qc-1.0/01-metrics/mrna--hg38/BLGSP-71-30-00667-01A-01S/alignment_summary_metrics
results/picard_qc-1.0/01-metrics/mrna--hg38/BLGSP-71-27-00428-01A-01S/alignment_summary_metrics
results/picard_qc-1.0/01-metrics/mrna--hg38/BLGSP-71-19-00128-09A.1-01R/alignment_summary_metrics
results/picard_qc-1.0/01-metrics/mrna--hg38/BLGSP-71-06-00159-01A-01R/alignment_summary_metrics
results/picard_qc-1.0/01-metrics/mrna--hg38/BLGSP-71-30-00658-01A-01S/alignment_summary_metrics
results/picard_qc-1.0/01-metrics/mrna--hg38/BLGSP-71-06-00097-01A-01R/alignment_summary_metrics
results/picard_qc-1.0/01-metrics/mrna--hg38/BLGSP-71-29-00537-50A-01R/alignment_summary_metrics
results/picard_qc-1.0/01-metrics/mrna--hg38/BLGSP-71-06-00093-01B-01S/alignment_summary_metrics
results/picard_qc-1.0/01-metrics/mrna--hg38/BLGSP-71-06-00255-01C-01R/alignment_summary_metrics
```

BrunoGrandePhD:
Stand by! I think I know why this happened. I selected samples randomly to run through this module, but I think that different samples are selected when Snakemake relaunched internally. I'm going to set a seed and re-run.

modules/picard_qc/1.0/picard_qc.smk
whelena: Ya we can, so users need to provide a bed file instead of an intervals file?

demo/Snakefile
whelena: I think It's because I need `CFG["samples"]` in the `_get_sample_metrics` functions and the function is evaluated before the `op.setup_module`. So when I try to run it, I receive a KeyError that "samples" does not exist. I'm open to better ideas to go around this because I prefer not having to specify it in the project snakefile.

modules/picard_qc/1.0/config/default.yaml
whelena: okay sounds good, I wasn't sure if users are allowed to modify those

modules/picard_qc/1.0/config/default.yaml
whelena: so as a general rule should I put anything the require the `switch_on` functions under switches?

modules/picard_qc/1.0/picard_qc.smk
whelena: Hmm I think it uses the default (same directory are the original bam files) but I will ads .bai symlinking

modules/picard_qc/1.0/picard_qc.smk
whelena: This is to create a merged metrics file with all the samples from a seq_type. The reason we wrote it in python was because it was originally a shell script but at a certain number of samples it fails because it reached a max character limit. The original shell script looks something like 
```grep -v '#' {input[0]} | sed '/^$/d' | head -n 1 | cut -f 2- > {output}.header &&
        for file in {input}; do
            echo 'Processing file:' ${{file}};
            grep -v '#' ${{file}} | sed '/^$/d' | sed -n 2p | cut -f 2- >> {output}.tmp;
        done &&
        echo {params.samples} | sed 's/ /\\n/g' | paste - {output}.tmp > {output} &&
        echo -e 'sampleID' | paste - {output}.header | cat - {output} > {output}.tmp &&
        mv {output}.tmp {output} && rm {output}.header```

File: modules/utils/2.0/config/default.yaml (hunk 12 above)
Comment (whelena) [reply modules/utils/2.0/config/default.yaml]: this is to create the interval list used in picard_qc because we talked about providing the bed files and write the rule in the utils module?

File: modules/picard_qc/1.0/picard_qc.smk (hunk 6 above)
Comment (whelena) [reply modules/picard_qc/1.0/picard_qc.smk]: I forgot I wrote that rule in the `utils` module

File: modules/utils/2.0/utils.smk (hunk 14 above)
Comment (whelena) [reply modules/utils/2.0/utils.smk:124]: it's the id associated to the exome kit (e.g. S07604624)

File: modules/utils/2.0/utils.smk (hunk 13 above)
Comment (whelena) [reply modules/utils/2.0/utils.smk:122]: what do you mean by that?

File: modules/utils/2.0/utils.smk (hunk 15 above)
Comment (whelena) [reply modules/utils/2.0/utils.smk]: hmm okay!

File: modules/picard_qc/1.0/picard_qc.smk (hunk 7 above)
Comment (whelena) [reply modules/picard_qc/1.0/picard_qc.smk]: I realise if I do that then I will have to provide a non-empty `strand` column even if I'm not preocessing mrna samples. Would it be reasonable to convert all the `Nan` to `"-"`, which would not be evaluated for non-`mrna` samples and evaluated as unstranded` for `mrna` samples`

File: workflows/reference_files/1.0/config/default.yaml (hunk 18 above)
Comment (whelena) [reply workflows/reference_files/1.0/config/default.yaml]: No that was from when I was trying to create a rule and didn't realize the working directory is already in `reference/`

File: modules/utils/2.0/config/default.yaml [hunk 19]
```diff
@@ -0,0 +1,27 @@
...
+
+        options:
+            bam_sort: ""
+            bam_markdups: ""
```
BrunoGrandePhD: While you're updating the `utils` module, could you add `--overflow-list-size 600000` here? This should help address the issue raised in #58.

modules/picard_qc/1.0/picard_qc.smk
lchong: Would this solution mean that a separate intervals file is generated for each BAM?

whelena: no I don't think so, it's only generated once for each bed file, since the interval files are created in a `reference` directory which is outside of the `result` directory

modules/picard_qc/1.0/picard_qc.smk
```diff
@@ -0,0 +1,326 @@
...
+        """)
+
+# mRNA metrics
+rule _picard_qc_rrna_int:
+    input:
+        bam = rules._picard_qc_input_bam.output.sample_bam,
+        fasta = reference_files("genomes/{genome_build}/genome_fasta/genome.fa"),
+        rrna_int = reference_files("genomes/{genome_build}/rrna_intervals/rRNA_int_gencode-33.txt")
+    output:
+        rrna_int = CFG["dirs"]["metrics"] + "{seq_type}--{genome_build}/{sample_id}/rrna_int_list"
```
lchong: Does this generate an rRNA intervals file for every individual sample? Don't we just need to generate this once? Or is this a different file?

modules/picard_qc/1.0/picard_qc.smk
lchong: I think this is a good way to get around the issues with character limits. I wonder if it would be worth discussing an even more general strategy that could be implemented as python function, possibly in oncopipe? Something that takes in a list of files to merge, and has options for which lines to skip, which lines to keep, etc.? Maybe the 3 of us can discuss further.

modules/picard_qc/1.0/picard_qc.smk
```diff
@@ -0,0 +1,326 @@
...
+    else:
+        m = base
+
+    targets = expand(rules._picard_qc_merged_output.output.metrics, **wildcards, metrics = m)
+
+    return targets
+
+ 
+rule _picard_qc_merged_dispatch:
+    input:
```
lchong, line 306: Can you use a function call in an expand statement? If so could you call _get_picard_qc_files in the "_picard_qc_all" rule below and avoid having this extra rule?

modules/picard_qc/CHANGELOG.md
```diff
@@ -0,0 +1,23 @@
...
+    - capture: alignment summary,insert size, and hs metrics.
+    - genome: alignment summary,insert size and wgs metrics.
+- Flagstats are also calculated via samtools for each sample.
+- The final outputs are merged metrics from all samples within a `seq_type`. 
+- By default, the `_picard_qc_merge_metrics` rule takes `config["lcr-modules"]["_shared"]["samples"]`. To modify, set `CFG["samples]` in the project Snakefile using:
+    ```python
+    config["lcr-modules"]["picard_qc"]["samples"] = SAMPLES
+    ```
+- Users are required to specify `CFG['inputs']['intervals']`, which refers to a text file containing the capture regions used in sequencing.
+- Reference files for CollectRNaSeq rRNA interval list and refFlat files) are specified in `CFG["inputs"]`. By default it refers to the files generated by `reference_files` subworkflow.
```
lchong: A couple of small typos in here (should be "CollectRnaSeqMetrics" and you have an extra ")" in the sentence).

modules/utils/2.0/utils.smk
```diff
@@ -0,0 +1,149 @@
...
+import sys, os
+from os.path import join
+
+# Import package with useful functions for developing analysis modules
+import oncopipe as op
+
+_UTILS = config["lcr-modules"]["utils"]
+LOG = "/logs/" + op._session.launched_fmt
+
+wildcard_constraints:
```
lchong: What does this constraint do?

modules/picard_qc/1.0/picard_qc.smk
whelena: oh I just realize the variable naming is misleading. The `input.rrna_int` is a reference generated only once while the `output.rrna_int` is the rrna intervals found in the sample bams. So theoutput has to be generated for each sample to run picard RNAseq metrics

modules/picard_qc/1.0/picard_qc.smk
whelena: hmm ya that could be useful, are there any other cases where we would want to merge a file that looks like this? Because I think for vcf files there are tools out there that does it for you

modules/picard_qc/1.0/picard_qc.smk
whelena: I think the issue is because the metrics called are different for each `seq_type` it would need to use the wildcard values. Since the target rule `picard_qc_all` doesn't have an output, it doesn't have a wildcard. There's probably a way to do it using the samples dataframe but this seems more transparent

modules/picard_qc/CHANGELOG.md
whelena: okay thank you!

modules/utils/2.0/utils.smk
whelena: it makes sure that `out_dir` has the format `results/<module_name>-1.0` and `prefix` starts with the subdirectories `02-picard_qc/...`. This is done because `utils` doesn't rely on `op.setup_module` and so we can insert the `LOG` path to create the log directories.

modules/picard_qc/1.0/picard_qc.smk
lchong: My other ideas would have been mutation calls (although we maybe should just use a vcf/maf merge tool as you mentioned) or even copy number calls. But I guess right now the only immediately obvious application is the QC files.

whelena: I'm  not sure if it would work for multiple file formats but if we need to merge other stuff in the future we can think about converting it into an oncopipe function instead

modules/utils/2.0/utils.smk
```diff
@@ -0,0 +1,149 @@
...
+    prefix = "[0-9]{2}-.*"
+
+
+##### RULES #####
+# _utils_bam_sort: Sort a BAM file using coordinates
+rule:
```
lchong: Why do these rules not have rule names?

modules/utils/2.0/utils.smk
```diff
@@ -0,0 +1,149 @@
...
+    input:
+        bam = "{out_dir}/{prefix}/{suffix}.bam"
+    output:
+        bam = "{out_dir}/{prefix}/{suffix}.mdups.bam"
+    log:
+        stdout = "{out_dir}" + LOG + "/{prefix}/{suffix}_mark_dups.stdout.log",
+        stderr = "{out_dir}" + LOG + "/{prefix}/{suffix}_mark_dups.stderr.log"
+    wildcard_constraints:
+        prefix = ".*(mark_dups).*"
```
lchong: Maybe I don't understand the purpose of this constraint, but should it be "mdups" instead of "mark_dups" since the output name has the ".mdups" addition?

modules/utils/2.0/utils.smk
```diff
@@ -0,0 +1,149 @@
...
+    params:
+        #logs = _get_log_dirs,
+        opts = _UTILS["options"]["bam_markdups"],
+        prefix = "{out_dir}/{prefix}/{suffix}"
```
lchong: Are you using the prefix param in this rule?

modules/utils/2.0/utils.smk
```diff
@@ -0,0 +1,149 @@
...
+        bam = "{out_dir}/{prefix}/{suffix}.bam"
+    output:
+        bam = "{out_dir}/{prefix}/{suffix}.bam.bai"
+    log:
+        stdout = "{out_dir}" + LOG + "/{prefix}/{suffix}_index.stdout.log",
+        stderr = "{out_dir}" + LOG + "/{prefix}/{suffix}_index.stderr.log"
+    params:
+        #logs = _get_log_dirs,
+        opts = _UTILS["options"]["bam_index"],
+        prefix = "{out_dir}/{prefix}/{suffix}"
```
lchong: Are you using the prefix param in this rule?

modules/utils/2.0/utils.smk
whelena: I think it was to prevent multiple rules having the same name when the utils module is included multiple times. I'm not sure if that would be a problem anymore. I'll try adding the names and see if it causes problems

modules/utils/2.0/utils.smk
whelena: I think it'll make more sense if you review this part with the `bwa_mem` module. it's because the subdirectory is named `02-mark_dups`

lchong:
Tested with the demo data and it looks good (/projects/clc/clc_scratch/projects/lchong/modules_testing/pull_requests/picard_qc-1.0/lcr-modules/demo).

I also tested with some of my own sample data -- 1 genome, 1 transcriptome, and 1 paired capture sample (run is in: /projects/clc/clc_scratch/projects/lchong/modules_testing/pull_requests/picard_qc-1.0). At first I tried to run it with multiple capture samples with different genomes, which didn't work (they seemed to get mixed up in some rules), but Helena and I have discussed and I don't think this is a use case we need to deal with, since having one seq_type with multiple genome builds is probably a rare case. However, if this does seem like a use case that will come up we may need to address it.

Also having an issue with the size of the rRNA interval files. In my old GRCh38 there were ~500 rRNA regions, whereas in the new version there are only ~50. Any idea why that would happen @BrunoGrandePhD?

whelena:
I addressed using multiple genome_build for capture metrics in the latest commit. I didn't adjust it for different intervals of the same genome_build, since that's probably even rarer.

whelena:
I'm having some issues trying to support multiple kits for a genome_build. For some reason I can't create an input function that outputs `op.switch_on_column`

whelena:
I pushed a commit with versions of the input function. The first function I was trying to replace empty values in the `capture_kit_id` column with `_default` in `CFG["samples"]` in the beginning of the module. 

Then I also tried having an input function that uses the wildcards values to subset `CFG["samples"] and allow us to work with a smaller dataframe. However this means that the function is run for every samples which can be redundant.

In the end I think we'll always need to use a function that calls `op.switch_on_function` because we need to specify the `genome_build` and `capture_kit_id` for each sample

BrunoGrandePhD:
Here's what I had in mind. I haven't tested this, so there might be bugs in it, but it should convey the idea. I'm assuming that this function will only be run for capture samples since the HS metrics rule should only run on those, right?

The key here is that `op.switch_on_column()` returns a function for Snakemake to run. Here, we are already in the input function, so we can call it right away with `wildcards`. 

```python
def _picard_qc_get_intervals_file(wildcards):

    CFG = config["lcr-modules"]["picard_qc"]
    options = CFG["switches"]["capture_intervals"]
    
    if "capture_kit_id" in CFG["samples"]:
        switch = op.switch_on_column("capture_kit_id", CFG["samples"], options)
        result = switch(wildcards)
    else:
        assert "_default" in switch, (
            "The '_default' key is missing for the 'capture_intervals' switch "
            "in the configuration for the 'picard_qc' module."
        )
        result = switch["_default"]

    return result
```

whelena:
I will try that but I have a feeling that op.switch_on_column will return something like `<function switch_on_column.<locals>._switch_on_column at 0x7f582bfbb680>`

UPDATE:
When I add `print(switch)`, it prints `<function switch_on_column.<locals>._switch_on_column at 0x7f35465fd050>` and triggers an `InputFunctionException`

BrunoGrandePhD:
`switch` is supposed to be a function. You need to call it manually here. `switch(wildcards)` should return the appropriate value from `capture_intervals`. What happens if you print `result` before returning it?

BrunoGrandePhD:
To clarify, `op.switch_on_column()` and `op.switch_on_wildcard()` return input/param functions. Normally, these are called by Snakemake when then rule is run. Here, Snakemake called the `_picard_qc_get_intervals_file` function instead. We can still use the `op.switch_on_column()` within this input file function, but we need to call it manually.

whelena:
Okay I get it. I ran it as is initially but the return value is an empty string.

BrunoGrandePhD:
@whelena Did you get it to work?

whelena:
@BrunoGrandePhD No I keep getting an empty string when I call the switch function. But I separated it into 1.0 and 1.1 because initially I was thinking of using it in my demo.

BrunoGrandePhD:
@whelena: Could you resolve the current conflicts so you don't have to change the PR after the review is done?

rdmorin:
I thought I resolved all the conflicts.  I don't understand why it still says Snakefile and config.yaml are still conflicting. These two files don't show up ad modified or conflicted after I merged.

rdmorin:
@BrunoGrandePhD It looks like we can't merge this without your help. Has Helena addressed your issue?

workflows/reference_files/2.4/reference_files_header.smk
```diff
@@ -255,6 +262,12 @@ def hardlink_same_provider(wildcards):
 
     for r in matching_rules:
         # The provider must be among the ones we can convert from
+<<<<<<< HEAD
```
BrunoGrandePhD: Unresolved merge.

workflows/reference_files/2.4/reference_files_header.smk
```diff
@@ -23,8 +23,12 @@ localrules: download_genome_fasta,
             get_genome_fasta_download, 
             index_genome_fasta,
             get_main_chromosomes_download, 
+<<<<<<< HEAD
+            get_gencode_download
+=======
             get_gencode_download,
             download_af_only_gnomad_vcf
+>>>>>>> master
```
BrunoGrandePhD: Unresolved merge.

workflows/reference_files/2.4/reference_files_header.smk
```diff
@@ -188,6 +192,8 @@ rule download_dbsnp_vcf:
         """)
 
 
+<<<<<<< HEAD
```
BrunoGrandePhD: Unresolved merge.

workflows/reference_files/2.4/reference_files.smk
```diff
@@ -173,6 +176,69 @@ rule get_dbsnp_download:
         tabix {output.vcf}
         """)
 
+<<<<<<< HEAD
```
BrunoGrandePhD: Unresolved merge.

workflows/reference_files/2.4/reference_files.smk
```diff
@@ -43,6 +43,8 @@ rule create_bwa_index:
         "bwa index -p {output.prefix} {input.fasta} > {log} 2>&1"
 
 
+<<<<<<< HEAD
```
BrunoGrandePhD: Unresolved merge.

BrunoGrandePhD (dismissed):
@rdmorin: There are unresolved merge conflicts that made it into the PR. These will break lcr-modules, so this PR isn't ready for merging. I've labelled those that I was able to find in this review. 

Note that if you mark a merge conflict as resolved (usually with `git add`), even though you didn't remove the `<<</---/>>>`, Git will assume that you've dealt with the conflicts. That's what I'm assuming happened here. 

Once these unresolved conflicts are fixed, I can take a look if anything else needs to happen to merge the PR.

workflows/reference_files/2.4/reference_files_header.smk
```diff
@@ -23,12 +23,9 @@ localrules: download_genome_fasta,
             get_genome_fasta_download, 
             index_genome_fasta,
             get_main_chromosomes_download, 
-<<<<<<< HEAD
             get_gencode_download
-=======
             get_gencode_download,
             download_af_only_gnomad_vcf
->>>>>>> master
```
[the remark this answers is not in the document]
rdmorin: Sorry about this. I won't have time to fix this any time soon.

oncopipe/oncopipe/__init__.py
rdmorin: I don't know how this even happened but I don't see why we can't just roll it back if those are the only changes. How can we figure out the cause of this?

rdmorin (changes requested):
Oncopipe __init__ has a bunch of strange changes (missing whitespace). Can this be rolled back to the version in Master? I don't see any other conflicts. I suppose someone fixed those.

rdmorin (approved):

rdmorin:
@whelena can you confirm that you implemented all Bruno's suggested changes?

rdmorin:
@whelena @BrunoGrandePhD I'm not sure what to do with this PR. I can't tell if Bruno's changes have been addressed.

whelena:
@rdmorin Sorry for the delay, I have addressed the changes required

rdmorin:
I mistakenly (because I thought gridss shouldn't be in the demo) removed the gridss lines from the demo Snakefile. 
I don't know how to revert this. Because there were other conflicts I had to resolve, I wonder if @BrunoGrandePhD or @lkhilton can look over the last commit at some point before this gets approved for merge. There were several conflicts and I'm worried the changes will have unintended consequences and at the very least we need gridss to go back into the demo.

BrunoGrandePhD:
Here's how I recommend you address this:

1. Download the latest branches from GitHub (_i.e._ `origin/`). 

    ```shell
    git fetch
    ```

2. Identify the differences between `master` and this branch (`module/picard_qc/1.0`) affecting the demo project. I'm using GitHub's branches (hence the `origin/` prefix) to avoid having to update local branches with two rounds of `git checkout` and `git pull`. This assumes that all changes in your local repo are on GitHub. 

    ```shell
    git diff origin/master origin/module/picard_qc/1.0 -- demo/
    ```

3. From the (ideally coloured) output of `git diff`, you can see which lines were changed or omitted. Here's [my coloured RTF output](https://drive.google.com/file/d/1STI1boj0CuuvP2aJAqMGqVAUPjhU76i7/view) for quick reference (open in TextEdit for better colours/fonts). The ones that catch my attention are the deleted `include` statement for the `gridss` module (in `demo/Snakefile`), the deleted `_gridss_all` entry under the `all` rule (in `demo/Snakefile`), and the switch from the newer `unmatched_normal_ids` to the older `pairing_config` approach (in `demo/config.yaml`). 

4. You can make the appropriate adjustments in a separate commit on this branch (`module/picard_qc/1.0`). This can either be done on GitHub or in a local repo. Because you can't edit more than one file per commit using the GitHub interface, I would make the changes in a local repo and push them to GitHub.

rdmorin:
How can I tell if this is merge-able now?

BrunoGrandePhD:
Were you able to confirm that the demo project in this branch works as expected?

rdmorin:
@Kdreval you ran this successfully, right?

Kdreval:
oh, not yet. I need to find out the strandness for the demo sample

rdmorin:
It's the same strandedness as the BLGSP RNA-seq libraries and all other stranded GSC libraries. I believe that is "positive" but you could confirm from Bruno's paper or ask around.

BrunoGrandePhD:
The stranded GSC libraries are `ISR` in [Salmon parlance](https://salmon.readthedocs.io/en/latest/library_type.html). In other words, read 1 is from the reverse/negative/anti-sense strand. I haven't confirmed myself whether the TCRB test RNA-seq BAM file follows the same protocol. Unfortunately, they only describe their WGS and WES protocols in the [paper](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4755126/).

rdmorin:
In the LIMS the libraries are coded as stranded = NA or stranded = 'positive'. @lkhilton  and I had been discussing (hoping) that meant plus strand. I guess it does not. I can confirm based on looking at the demo bam in IGV with the reads coloured based on strand that the strandedness is the same as BLGSP bams.

Kdreval:
I have started the demo run and will update on how it goes

Kdreval:
I have run the demo project on this branch successfully upon completion on cluster.

modules/utils/2.0/utils.smk
```diff
@@ -48,7 +48,7 @@ rule _utils_bam_sort:
     shell:
         op.as_one_line("""
         samtools sort 
-        -@ {threads} -m $(({resources.mem_mb} / {threads}))M
+        -@ {threads} -m $(({resources.mem_mb} / {threads} / 2))M
```
rdmorin: Does this always return an integer even if resources.mem_mb/threads yields an odd number?

rdmorin (approved):
Just one comment/question on the division by 2

modules/utils/2.0/utils.smk
Kdreval: I don't think it will always return an integer. If the threads will be odd number, this will return float, and probably exit with error. Let me try to do that  in resources, so I can convert any number to integer, and test how star runs with such modification.

rdmorin: I think in @lkhilton's GRIDSS or GRIDSS-PURPLE-LINX module she had a way of doing this with Python to ensure it returns an integer

BrunoGrandePhD: You can use `//` in Python to drop the remainder, _i.e._ ensure an integer ([source](https://stackoverflow.com/a/1535601)).

Kdreval: Thanks for all the suggestions! I don't know how it even worked in current configuration, because (48000/12/2) returns `2000.0`, and not `2000`. The `//` is good in dry run. Let me just make sure real run completes, and does not exit with any error.

lkhilton: I'm using `int()` to return an integer.

Kdreval: I have ensured the integer to be returned in the memory resource for the bam_sort rule in most recent commit

rdmorin (approved):

PR #60, opened 2020-06-30
Module/mixcr/1.0

adossantos1:
# Pull Request Checklists

### Required

- 

- [x]  I used the cookiecutter template and updated the placeholder rules.

  ```bash
  ```

### If applicable

To be completed.


modules/mixcr/1.0/config/default.yaml
```diff
@@ -0,0 +1,37 @@
...
+        options:
+            mixcr_run: 
+                mrna:
+                    --starting-material rna
+                genome:
+                    --starting-material dna
+
+        # TODO: Add the conda environment once MiXCR is available again - using the jar for now.
+        conda_envs:
+            mixcr: "{MODSDIR}/envs/mixcr"
```
rdmorin: This file name should end in .yml

modules/mixcr/1.0/config/default.yaml
```diff
@@ -0,0 +1,37 @@
...
+            
+        threads:
+            mixcr_run: 4
+
+        mem_mb:
+            mixcr_run: 2000
```
rdmorin: Does it really not use more than this amount of RAM ever?

modules/mixcr/1.0/mixcr.smk
```diff
@@ -0,0 +1,130 @@
...
+
+# Installs latest MiXCR release from github if the mixcr folder is not present yet
+rule _install_mixcr:
+    params:
+        mixcr = CFG["lcr-modules"] + CFG["conda_envs"]["mixcr"]
+    output: 
+        complete = CFG["lcr-modules"] + CFG["conda_envs"]["mixcr"] + "/mixcr_dependencies_installed.success"
+    shell:
+        '''
+        latest_tag=$(curl --silent "https://api.github.com/repos/milaboratory/mixcr/releases/latest" |  grep '"tag_name":'| sed -E "s/v//" | sed -E 's/.*\"([^\"]+)\".*/\\1/');
```
rdmorin: This seems convoluted. Can't you just use the latest as-is without knowing the tag name?

adossantos1: You're right, I've made some changes

modules/mixcr/1.0/config/default.yaml
adossantos1: I'm running the test data again with the `time -v` command Bruno showed yesterday and will update the value once the run is done.

modules/mixcr/1.0/config/default.yaml
adossantos1: MiXCR comes as a jar with an executable script for the time being which I've set to install in `envs\mixcr`. I will remove the TODO for clarity and maybe move the path to the mixcr script under a different tag.

rdmorin: If its not a conda environment then you should move this to "params" or, probably better yet, "inputs"

modules/mixcr/1.0/config/default.yaml
```diff
@@ -0,0 +1,34 @@
...
+                mrna:
+                    --starting-material rna
+                genome:
+                    --starting-material dna
+            
+        threads:
+            mixcr_run: 4
+
+        mem_mb:
+            mixcr_run: 5300
```
rdmorin: I would just set this to the next highest round number, e.g. 6000

rdmorin:
Are you still working on changes, @adossantos1 ?

adossantos1:
No, I've made all the suggested changes and run it on the test mRNA fastqs.

modules/mixcr/1.0/config/default.yaml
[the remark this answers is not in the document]
adossantos1: The above yaml is outdated - I have since changed it to 6000 (based on /bin/time/ -v <command> output) which hasn't given me problems yet

rdmorin (approved):
Weird that I couldn't see your more recent changes when I tried to review the PR

adossantos1:
Yeah, they just changed for me as well

rdmorin (approved):

PR #62, opened 2020-07-01
Module/lofreq/1.0

BrunoGrandePhD:
# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable

To be completed.


BrunoGrandePhD:
This is still a draft. I was able to run the module successfully on the test data. I need to update the CHANGELOG and test on some other data. This PR should only be merged once #41 is merged.

demo/dry-run.sh
```diff
@@ -1,3 +1,6 @@
 #!/bin/bash
 
-snakemake --dryrun --cores 1 --printshellcmds --reason --use-conda all
+# Default to all targets
+TARGETS=${@:-all}
+
+snakemake --dryrun --cores 1 --printshellcmds --reason --use-conda $TARGETS
```
rdmorin: Good call!

modules/lofreq/1.0/lofreq.smk
```diff
@@ -0,0 +1,115 @@
...
+    resources:
+        mem_mb = op.retry(CFG["mem_mb"]["lofreq"])
+    shell:
+        op.as_one_line("""
+        lofreq somatic {params.opts} --threads {threads} -t {input.tumour_bam} -n {input.normal_bam}
+        -f {input.fasta} -o $(dirname {output.vcf_snvs})/ -d {input.dbsnp} {params.regions} 
+        > {log.stdout} 2> {log.stderr}
+        """)
+
+
```
rdmorin, line 117: Should we incorporate vcf2maf into this (and all variant callers) directly? It's not clear to me if that will be optional and, if so, how.

modules/sequenza/1.0/config/default.yaml
```diff
@@ -0,0 +1,48 @@
...
+    sequenza:
+
+        inputs:
+            # Available wildcards: {seq_type} {genome_build} {sample_id}
+            sample_bam: "__UPDATE__"
+            sample_bai: "__UPDATE__"
+            merge_seqz: "{MODSDIR}/src/bash/merge_seqz.sh"
+            filter_seqz: "{MODSDIR}/src/bash/filter_seqz.sh"
+            run_sequenza: "{MODSDIR}/src/R/run_sequenza.R"
+            cnv2igv: "{SCRIPTSDIR}/cnv2igv/1.0/cnv2igv.py"
```
rdmorin: Why not point to the latest version of cnv2igv?

modules/sequenza/1.0/sequenza.smk
```diff
@@ -0,0 +1,240 @@
+#!/usr/bin/env snakemake
+
+
+##### ATTRIBUTION #####
+
+
+# Original Author:  Ryan Morin
+# Module Author:    Ryan Morin
+# Contributors:     N/A
```
rdmorin: Contributors: Bruno Grande

modules/sequenza/1.0/src/python/cnv2igv.py
```diff
@@ -0,0 +1,232 @@
+import argparse
+import re
```
rdmorin: I think we can remove this file now that it is in lcr-scripts, right?

oncopipe/oncopipe/__init__.py
```diff
@@ -291,7 +291,43 @@ def list_files(directory, file_ext):
     return files_all
```
rdmorin: I can't really comment meaningfully on this file

rdmorin:
What's going on with this one, @BrunoGrandePhD ? Should someone else take it on?

BrunoGrandePhD:
Yeah, any spare time I have should go towards MBB 343 and oncopipe.

rdmorin:
@Kdreval can you take over and try to finalize this tool?

Kdreval:
Yes, I will be able to do it. From what I see in this conversation, it has already been tested on some, but not many samples? I will also go through comments.

Kdreval:
> I don't understand why there all these modifications to Sequenza in this PR. Was that on purpose? There is at least one modification that will break the current version (the auto-installation). Kostia has already fixed that in the latest version.

This all changes are to 1.0 version. As the current version is 1.4, I think that sequenza module will not be affected by this changes. Also, some of your comments here were already addressed (e.g. cnv2igv script now refers to lcr-scripts)

modules/lofreq/1.0/lofreq.smk
Kdreval: Are you thinking about incorporating the whole module here? I think that might not be necessary, as we can combine them together in the project snakefile - similar to what is in demo project. Alternatively, I can do an if statement based on config, and that if will also include all rule - so the final output will be the maf file.

Kdreval:
> This is still a draft. I was able to run the module successfully on the test data. I need to update the CHANGELOG and test on some other data. This PR should only be merged once #41 is merged.

I will test it by running on all samples available in gambl. That must address testing on both genome and capture seq_types.

rdmorin:
I understand. Thanks for clarifying.

modules/lofreq/1.0/lofreq.smk
```diff
@@ -0,0 +1,115 @@
...
+        stderr = CFG["logs"]["lofreq"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/lofreq.stderr.log"
+    params:
+        opts = CFG["options"]["lofreq"],
+        regions = op.switch_on_wildcard("seq_type", CFG["switches"]["regions_bed"])
+    conda:
+        CFG["conda_envs"]["lofreq"]
+    threads:
+        CFG["threads"]["lofreq"]
+    resources:
+        mem_mb = op.retry(CFG["mem_mb"]["lofreq"])
```
Kdreval: Bruno, doesn't this `op.retry` need to take 3 arguments instead of 1 given here? I am judging by [this](https://lcr-modules.readthedocs.io/en/stable/oncopipe/oncopipe.html#oncopipe.retry). In the current setup, it does not retry if killed because of memory excess. I think It is because multiplier and max.value needs to be added.

Kdreval:
I have tested on capture sample from demo workflow and genomes. I made sure it runs on cluster, the recent master is merged, demo workflow completes dry run, and the changelog is updated.

modules/lofreq/1.0/lofreq.smk
```diff
@@ -67,7 +67,8 @@ rule _lofreq_run:
     threads:
         CFG["threads"]["lofreq"]
     resources:
-        mem_mb = op.retry(CFG["mem_mb"]["lofreq"])
+        mem_mb = CFG["mem_mb"]["lofreq"],
```
rdmorin: Was this a bug?

modules/lofreq/1.0/lofreq.smk
lkhilton: Since Ryan made this comment before he'd made a vcf2maf module, I'd say this is no longer relevant.

modules/lofreq/1.0/lofreq.smk
```diff
@@ -0,0 +1,116 @@
...
+        stdout = CFG["logs"]["lofreq"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/lofreq.stdout.log",
+        stderr = CFG["logs"]["lofreq"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/lofreq.stderr.log"
+    params:
+        opts = CFG["options"]["lofreq"],
+        regions = op.switch_on_wildcard("seq_type", CFG["switches"]["regions_bed"])
+    conda:
+        CFG["conda_envs"]["lofreq"]
+    threads:
+        CFG["threads"]["lofreq"]
+    resources:
```
lkhilton, line 72: Can you change this to use the resource unpacking instead? e.g. `**CFG["resources"]["lofreq"]`

modules/lofreq/1.0/config/default.yaml
```diff
@@ -0,0 +1,41 @@
...
+                # Set to "" if you have no capture data
+                capture: "-l __UPDATE__"
+
+        conda_envs:
+            lofreq: "{MODSDIR}/envs/lofreq-2.1.5.yml"
+            
+        threads:
+            lofreq: 48
+
+        mem_mb:
```
lkhilton: See my comment below about resource unpacking. This would need to be changed to: 
```
resources: 
    lofreq: 
        mem_mb = 100000
        bam = 1
```

modules/lofreq/CHANGELOG.md
```diff
@@ -0,0 +1,21 @@
+# Changelog
+
+All notable changes to the `lofreq` module will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [1.0] - 2020-10-15
```
lkhilton, line 8: Kostia I think you could collapse your part of the changelog with Bruno's.

modules/lofreq/1.0/lofreq.smk
Kdreval: I don't think it is a bug. I think it did not work for all the samples as it was originally thought. I found that it will multiply whatever is in `mem_mb` by 1.5, and this will be used as a resource for the job scheduling. Some samples were completed successfully, but others - exited being killed by slurm for exceeding memory limit. On the other hand, the same samples were completed if I specify memory directly, without a function. Therefore, I changed it here.

modules/lofreq/1.0/lofreq.smk
```diff
@@ -0,0 +1,116 @@
...
+        bai = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{sample_id}.bam.bai"
+    run:
+        op.relative_symlink(input.bam, output.bam)
+        op.relative_symlink(input.bai, output.bai)
+
+
+# Run LoFreq in somatic variant calling mode
+rule _lofreq_run:
+    input:
+        tumour_bam = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{tumour_id}.bam",
```
lkhilton, line 52: For the Strelka module we decided that we would always merge the snv and indel vcf files and delete the individual variant type files. Could you do the same here? 

Do the VCF files need any filtering? And are there any intermediate files output by lofreq or is it just the vcf files?

lkhilton (changes requested):

modules/lofreq/1.0/lofreq.smk
rdmorin: Agree with Laura. Obsolete comment

rdmorin (approved):

modules/lofreq/1.0/lofreq.smk
Kdreval: In regards to filtering, I found that there is a default filter in lofreq in and it says `Predicted variants are already filtered using default parameters (which include coverage, strand-bias, SNV-quality etc). There will usually be no need for you to filter the output again`. I looked further on what are those default parameters, and see they are mentioned here `the main filtering step is working on the variant qualities (which are converted p-values) and it's by default based on Bonferroni correction and a significance threshold of 0.01`. Examining the output, I do see for several files that I checked that in the output file, there are already only snvs with PASS in the filter column. So, I think that the filtering for these vcfs is not needed.
For the intermediate files, there are no temp files created. Regarding output, it produces total 26 files (including logs and indices), of which we are really interested in 4 (those are that got symlinked into the 99-outputs). Are you thinking about clearing those files? I think I would keep them, since it includes the calls for both normal and tumour samples.

rdmorin: Do you have an example loFreq output directory I could have a look at?

rdmorin: I suspect everything else here that I haven't indicated with a comment should be wrapped in temp() so it is deleted after.

```
normal_relaxed.log                  
normal_stringent.snvs.vcf.gz.tbi             
somatic_raw.snvs.vcf.gz.tbi        
tumor_stringent.snvs.vcf.gz
normal_relaxed.vcf.gz               
somatic_final.indels.vcf.gz                  #merge and output as full mutation VCF
somatic_final.snvs.vcf.gz                  #merge and output as full mutation VCF, regenerate index
tumor_relaxed.log                  
tumor_stringent.snvs.vcf.gz.tbi
normal_relaxed.vcf.gz.tbi           
somatic_final.indels.vcf.gz.tbi              
somatic_final.snvs.vcf.gz.tbi              
tumor_relaxed.vcf.gz
normal_stringent.indels.vcf.gz      
somatic_raw.indels.vcf.gz                  
tumor_relaxed.vcf.gz.tbi
normal_stringent.indels.vcf.gz.tbi  
somatic_final_minus-dbsnp.indels.vcf.gz.tbi  
somatic_final_minus-dbsnp.indels.vcf.gz      #merge and output as snp-filtered vcf
somatic_final_minus-dbsnp.snvs.vcf.gz.tbi  
somatic_final_minus-dbsnp.snvs.vcf.gz      #merge and output as snp-filtered vcf, regenerate index
somatic_raw.indels.vcf.gz.tbi              
tumor_stringent.indels.vcf.gz
normal_stringent.snvs.vcf.gz        
somatic_raw.snvs.vcf.gz                    
tumor_stringent.indels.vcf.gz.tbi
```

lkhilton: Kostia and I discussed that since most of these files are very small we could keep them for the sake of understanding the calls loFreq makes, especially if we discover things are quite different from Strelka or Mutect.

rdmorin: Fair enough. I suggest at the very least we delete normal_relaxed.vcf.gz

Kdreval:
I have addressed comments about unpacking resources, combining snvs and indels into single vcf, and deleting the large file. I also tested this modification on both exome and genome sample.

rdmorin:
@lkhilton if you are satisfied with his changes can you please approve ASAP? I would be grateful to be able to start deploying this on GAMBL (in particular, for BLGSP).

modules/lofreq/1.0/lofreq.smk
```diff
@@ -54,51 +54,89 @@ rule _lofreq_run:
...
-        bam =1
+        **CFG["resources"]["lofreq"]
     shell:
         op.as_one_line("""
         lofreq somatic {params.opts} --threads {threads} -t {input.tumour_bam} -n {input.normal_bam}
-        -f {input.fasta} -o $(dirname {output.vcf_snvs})/ -d {input.dbsnp} {params.regions} 
+        -f {input.fasta} -o $(dirname {output.vcf_snvs_filtered})/ -d {input.dbsnp} {params.regions} 
         > {log.stdout} 2> {log.stderr}
+            &&
+        rm {params.rm_files}
```
lkhilton, line 80: Kostia, what's the rationale for an explicit deletion command, rather than marking the file as a temp output file?

Kdreval: I did in that way, because I thought it will make easier to add here other files, too, if we will decide to clear other output files later. Otherwise, I was thinking I will over complicate output of the rule.

lkhilton (approved):

PR #63, opened 2020-07-03
This is the first draft of liftover module

Kdreval:
# Pull Request Checklists

### Required

  ```bash
  ```

To be completed.


BrunoGrandePhD:
I'm busy for the next two weeks before I can tackle a new PR. Please bear with us as we work our way through the backlog, which is a good problem to have! I'm glad everyone is contributing!

rdmorin:
I suggested Kostia submit this so I could do some testing. I plan to get to this when I'm back.

modules/liftover/1.0/liftover.smk
```diff
@@ -0,0 +1,157 @@
...
+CFG = op.setup_module(
+    name = "liftover",
+    version = "1.0",
+    subdirectories = ["inputs", "seghg38tobedhg38", "bedhg38tobedhg19", "bedhg19toseghg19", "outputs"])
+
+# Define rules to be run locally when using a compute cluster
+localrules:
+    _liftover_input_seg,
+    _hg38seg_2_hg38bed,
+    _hg38bed_2_hg19bed,
```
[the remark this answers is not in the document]
Kdreval: I think now that this whole module will not require the cluster nodes, and I will remove the resource requirements for these rules

modules/liftover/1.0/config.yaml
```diff
@@ -0,0 +1,6 @@
+lcr-modules:
```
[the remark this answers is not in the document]
Kdreval: I do not need this config and will remove this file

modules/liftover/1.0/config/default.yaml
```diff
@@ -0,0 +1,29 @@
+lcr-modules:
+
+    liftover:
+        inputs:
+            sample_seg: "data/{sample_id}_subclones.hg38.igv.seg"
+        scratch_subdirectories: "scratch/"
+        options:
+            seghg38tobedhg38: "{MODSDIR}/src/convert_for_liftover.py"
+            bedhg38tobedhg19: "{MODSDIR}/src/chains/hg38ToHg19.over.chain"
```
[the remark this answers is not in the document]
Kdreval: Did you mean the global reference directory that is created by the utils module before running star? Or it will be better to create the "{MODSDIR}/reference"?

demo/config.yaml
```diff
@@ -20,3 +20,31 @@ lcr-modules:
...
+            bedhg38tobedhg19: "{MODSDIR}/src/chains/hg38ToHg19.over.chain"
+            bedhg19toseghg19: "{MODSDIR}/src/convert_for_liftover.py"
+        conda_envs:
+            liftover-366: "{MODSDIR}/envs/liftover-366.yaml"
+        threads:
+            seghg38tobedhg38: 12
+            bedhg38tobedhg19: 12
+            bedhg19toseghg19: 12
+        mem_mb:
+            seghg38tobedhg38: 100000
```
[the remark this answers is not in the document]
Kdreval: It does not require that much memory. Even if I limit to only 1 core, it finishes running the test file in less than 5 sec, so even if I do `| htop` , I still can't see how much memory exactly is used. I think that once it will be tested on a large number of files,  it will be better to estimate memory usage and then specify `mem_mb` value.

modules/liftover/1.0/liftover.smk
```diff
@@ -0,0 +1,157 @@
...
+
+
+##### RULES #####
+
+# Symlinks the input files into the module results directory (under '00-inputs/')
+rule _liftover_input_seg:
+    input:
+        seg = CFG["inputs"]["sample_seg"]
+    output:
+        seg = CFG["dirs"]["inputs"] + "seg/{sample_id}_subclones.hg38.igv.seg"
```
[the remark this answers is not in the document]
Kdreval: > Can you address these suggestions before I get into testing this?

I have addressed all these suggestions, tested on several files, and ready to push changes.

demo/config.yaml
```diff
@@ -23,28 +23,4 @@ lcr-modules:
...
-        pairing_config:
-            genome:
-                run_paired_tumours: True
-                run_unpaired_tumours_with: null
-                run_paired_tumours_as_unpaired: False
-            capture:
-                run_paired_tumours: True
-                run_unpaired_tumours_with: null
-                run_paired_tumours_as_unpaired: False
+            sample_seg: "data/{tumour_sample_id}_subclones.hg38.igv.seg"
```
rdmorin: This is the demo config. You shouldn't be committing edits to this file

lkhilton: Isn't it helpful to demo how to use new modules as they're introduced?

rdmorin: Sorry, my mistake. I see that you have retained the part of the config for the demo. This file needs a newline added at the end.

Kdreval: Also, I see now that the new modules, such as sequenza and strelka were added to this file. I will push the updated version of this file that includes these modules, and a new line at the end

modules/liftover/1.0/config/default.yaml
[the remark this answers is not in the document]
rdmorin: Yes, in essence, the references workflow would eventually need to be updated to pull these files down from UCSC for each reference build.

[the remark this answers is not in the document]
rdmorin: If the file is a modest size I'm fine with it being here instead

modules/liftover/1.0/liftover.smk
```diff
@@ -23,46 +23,35 @@ CFG = op.setup_module(
...
-    _liftover_all,
-
 
 ##### RULES #####
 
 # Symlinks the input files into the module results directory (under '00-inputs/')
 rule _liftover_input_seg:
     input:
-        seg = CFG["inputs"]["sample_seg"]
+        seg = expand(CFG["inputs"]["sample_seg"], tumour_sample_id=CFG["runs"]["tumour_sample_id"], normal_sample_id=CFG["runs"]["normal_sample_id"], tool=CFG["tool"])
```
rdmorin: If the user is setting this tool as "battenberg" (as per the documentation in the config) then how will this ever match the file naming produced by battenberg? Is the user expected to rename all those files first?

demo/Snakefile
```diff
@@ -44,14 +45,16 @@ config["lcr-modules"]["_shared"]["samples"] = SAMPLES
...
 ##### TARGETS ######
 
 
 rule all:
     input:
-        rules._manta_all.input,
-        rules._star_all.input
+        #rules._manta_all.input,
+        #rules._star_all.input,
+        rules._liftover_all.input
```
rdmorin, line 67: If you are adding liftover to the demo you will need to include a demo input file in data or this will fail. Is your plan to include a Sequenza output from the demo exome?

modules/liftover/1.0/liftover.smk
Kdreval: No, there will be no need to rename the files before using with this module. I specified these wildcards so they will be recognized from input files, otherwise, snakemake gives an error that it cannot determine wildcards of output files from the inputs. I tested it with the file input named 08-20595T_subclones.hg38.igv.seg, which will be the output of battenberg

demo/Snakefile
Kdreval: I did not add the demo files to be committed to the demo/data directory. I will add files to test both battenberg and sequenza

modules/liftover/1.0/liftover.smk
rdmorin: I don't understand. 
tool: __UPDATE__ # Specify the name of the tool that produced the .seg files. For example, "battenberg" 

Where does the value of this variable get used and how does it allow the input files to be recognized? I must be missing something here.

Kdreval: It will be used to retain the name of the tool that produced seg files in the naming of final files that are produced by liftover. It is used as a wildcard {tool} in the liftover.smk file, and it is the sole purpose of this config option. I used it so when there are final lifted files, it will be possible to recognize which CNV caller was used producing original seg files. It will allow the input files to be recognized as belonging to a certain tool. Instead, in the config file, both "tool" and "sample_seg" needs to be updated.

modules/liftover/1.0/config/default.yaml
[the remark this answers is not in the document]
Kdreval: The size of the chain file is 3.9Mb for one reference build. I think that it can be here for now, but with the next update of reference files downloading liftover chains can be included, and then I will update the link to the location of reference chain in this config

modules/liftover/1.0/liftover.smk
rdmorin: input_seg: "results/gambl/{tool}_0.0/{genome_build}/{tumour_sample_id}--{normal_sample_id}/{tumour_sample_id}_subclones_1.igv.seg"

Isn't this what your config would look like?

Kdreval: > To clarify, it seems to me this would only work if you started with input files that are named something along these lines:
> "seg/20595T--20595N.battenberg.hg38.igv.seg"

I got it now. No, it will work based on what is in the config. I am reading the file from what is specified in config, regardless of its naming pattern - what the user provides in the filed "sample_seg" of the input in the config/default.yaml . But, within this rule and during symlinking, the output (the symlink in the 00-inputs folder) will have a different naming pattern - this is where I am taking advantage of bringing everything to the same consistent naming pattern. Now that I am writing this I also realized that the sample will be recognized only if it is listed in the samples table, and only if it is paired.

Kdreval: > input_seg: "results/gambl/{tool}_0.0/{genome_build}/{tumour_sample_id}--{normal_sample_id}/{tumour_sample_id}_subclones_1.igv.seg"
> 
> Isn't this what your config would look like?

let me test a path like this and naming scheme like this

Kdreval: shouldn't it be with 99-outputs?
`input_seg: "results/gambl/{tool}_0.0/99-outputs/{genome_build}/{tumour_sample_id}--{normal_sample_id}/{tumour_sample_id}_subclones_1.igv.seg"`
I tested with the path you mentioned and it does not recognize wild card genome build in the configuration I pushed. I thought this wildcard is "build-in" into lcr-modules and I would not need to specify it. If I specify the genome_build wildcard, it is working. This is the output for the first rule:
```
rule _liftover_input_seg:
    input: results/gambl/battenberg_0.0/hg38/08-20595T--08-20595N/08-20595T_subclones_1.igv.seg
    output: results/liftover-1.0/00-inputs/seg/08-20595T--08-20595N.battenberg.igv.seg
    jobid: 5
    reason: Missing output files: results/liftover-1.0/00-inputs/seg/08-20595T--08-20595N.battenberg.igv.seg
    wildcards: tumour_sample_id=08-20595T, normal_sample_id=08-20595N, tool=battenberg
```

rdmorin:
I am not confident this structure will behave as desired. To test this I have added a second sample pair to samples.tsv and in dry run it seems to only ever process the first pair in the file. Can you please confirm this by pulling this file and also do the following:
make a copy of the test seg file and give it a name that matches the tumour ID of the new fake sample. 

e.g. 
```cp 08-20595T_subclones.hg38.igv.seg 99-20595T_subclones.hg38.igv.seg```

Do you see the same behaviour?

PR #64, opened 2020-07-06
Module/bam2fastq/1.0

whelena:
# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable

To be completed.

### DIFF
--- a/demo/Snakefile
+++ b/demo/Snakefile
@@ -25,6 +25,7 @@ subworkflow reference_files:
 
 
 # Load module-specific configuration
+configfile: "../modules/bam2fastq/1.0/config/default.yaml"
 configfile: "../modules/star/1.1/config/default.yaml"
 configfile: "../modules/manta/2.1/config/default.yaml"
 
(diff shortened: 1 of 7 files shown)


lkhilton:
I realize this is a draft still, but I'm being nosy because @adossantos1 wants to try out her MiXCR module on GAMBL. We're going to have space issues if she tries to convert all the GAMBL RNAseq BAM files to FASTQ simultaneously, is there a mechanism that can be implemented here to limit the number of fastq files that can exist at a time? In the context of large projects like GAMBL unfortunately fastqs will always have to be temporary.

whelena:
That's a good point. When writing this module I assumed that users wants to keep the fastq files. I'm wondering how often would you want to keep the fastq files vs havinh it temporarily. 

I could also write an additional rule in `utils` that is essentially bam2fastq, so that users can run it as part of a module without keeping the fastq files

lkhilton:
That could work nicely.

PR #68, opened 2020-07-27
Modules/manta/2.3

rdmorin:
# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable

I added one line to Manta.smk to allow a users to limit the resources used by specifying "resources bam=N" where N is the maximum number of concurrent Manta jobs.


rdmorin:
Is @whelena able to approve this please? It is a minor change that I need for GAMBL

Issue #69, opened 2020-07-27
STAR index building fails


rdmorin:
I get this error when running the demo workflow. It looks like the STAR command might be garbled (after the 2> part)

```
Error in rule create_star_index:
    jobid: 2
    output: genomes/grch37/star_index/star-2.7.3a/gencode-33/overhang-99
    log: genomes/grch37/star_index/star-2.7.3a/gencode-33/overhang-99.log (check log file(s) for error message)
    conda-env: /home/rmorin/project/rmorin/software/conda_environments/478bc34b
    shell:
        mkdir -p genomes/grch37/star_index/star-2.7.3a/gencode-33/overhang-99 && STAR --runThreadN 12 --runMode genomeGenerate --genomeDir genomes/grch37/star_index/star-2.7.3a/gencode-33/overhang-99 --genomeFastaFiles genomes/grch37/genome_fasta/genome.fa --sjdbOverhang 99 --sjdbGTFfile genomes/grch37/downloads/gencode-33/gencode.annotation.grch37.gtf --outTmpDir genomes/grch37/star_index/star-2.7.3a/gencode-33/overhang-99/_STARtmp > genomes/grch37/star_index/star-2.7.3a/gencode-33/overhang-99.log 2>&1 --outFileNamePrefix genomes/grch37/star_index/star-2.7.3a/gencode-33/overhang-99/

**Comment by rdmorin** [comment]
The output file prefix is not being passed in a way that is ever likely to work. I'm surprised this didn't come up as an issue sooner. 
```
shell:
        op.as_one_line("""
        mkdir -p {output.index}
            &&
        STAR --runThreadN {threads} --runMode genomeGenerate --genomeDir {output.index}
        --genomeFastaFiles {input.fasta} --sjdbOverhang {wildcards.star_overhang}
        --sjdbGTFfile {input.gtf} --outTmpDir {output.index}/_STARtmp > {log} 2>&1
        --outFileNamePrefix {output.index}/
        """)

lkhilton:
I ran this on Friday last week, the STAR indices all generated as expected. 
``` 
cd /projects/rmorin_scratch/lcr-modules-references/genomes && 
ls */star_index/star-2.7.3a/gencode-33/overhang-*
```

lkhilton:
To be clear, I created those STAR indices by running the following: 
```
cd /projects/rmorin_scratch/lcr-modules-references
[...]

PR #71, opened 2020-07-27
fixed mem_mb to allow demanding indexing jobs to complete

rdmorin:
I have fixed reference_files to run on a cluster by adding mem_mb in a few places. I suggest we consider keeping the version naming the same here because the outputs (when successful) should be identical.

### DIFF
--- a/demo/Snakefile
+++ b/demo/Snakefile
@@ -16,9 +16,9 @@ subworkflow reference_files:
     workdir:
         "reference/"
     snakefile:
-        "../workflows/reference_files/2.2/reference_files.smk"
+        "../workflows/reference_files/2.3/reference_files.smk"
     configfile:
-        "../workflows/reference_files/2.2/config/default.yaml"
+        "../workflows/reference_files/2.3/config/default.yaml"
 
 
 ##### CONFIGURATION FILES #####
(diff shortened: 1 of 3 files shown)


workflows/reference_files/2.3/reference_files.smk
```diff
@@ -51,14 +53,16 @@ rule create_star_index:
...
     threads: 12
+    resources:
+        mem_mb = 42000
     shell:
         op.as_one_line("""
         mkdir -p {output.index}
             &&
         STAR --runThreadN {threads} --runMode genomeGenerate --genomeDir {output.index}
         --genomeFastaFiles {input.fasta} --sjdbOverhang {wildcards.star_overhang}
-        --sjdbGTFfile {input.gtf} --outTmpDir {output.index}/_STARtmp > {log} 2>&1
```
rdmorin, line 60: Can someone please confirm this was a bug?

demo/Snakefile
```diff
@@ -16,9 +16,9 @@ subworkflow reference_files:
     workdir:
```
rdmorin, line 16: Now in sync with the current version

rdmorin:
This patch resolves #69

workflows/reference_files/2.3/reference_files.smk
lkhilton: I don't think it is a bug. That should redirect both STDOUT and STDERR to {log}. I can confirm at least that stdout is redirected, since I didn't have any errors when I generated the files I can't confirm re: STDERR. But I think you were able to see errors in this file when you were having memory issues on Cedar, right?

lkhilton: Anyway, moving it to the end of the STAR command is fine.

PR #81, opened 2020-08-21
Module/cellranger/1.0

whelena:
# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable

To be completed.


Kdreval:
@whelena what is the status of this module? Seems like it hasn't been reviewed? Do you have any test data to run through it?

Kdreval:
gonna close this PR. We can always restart from this branch whenever data is available

Issue #70, opened 2020-07-27
Provide other common resources for more restrictive schedulers


rdmorin:
Some schedulers impose a default time constraint on jobs that can be quite short. For example, my jobs on Cedar (Westgrid) seem to be given a limit of 60 minutes. We should either provide a wrapper that requests a much longer/realistic time for jobs or incorporate this into the rules for all non-local jobs. We may have to address this for other limits also.

BrunoGrandePhD:
Currently, resources (mainly `mem_mb` for now) are listed individually in the module snakefile ([example](https://github.com/LCR-BCCRC/lcr-modules/blob/master/modules/manta/2.3/manta.smk#L139-L141)). Because of this, custom resources need to be added to the snakefiles. 

In order words, there is no way of adding custom resources on a per-rule basis by simply editing the YAML configuration file. That said, I think this can easily be implemented. 

The key here is to leverage the ability for snakemake to expand python dictionaries to take in key-value pairs. This functionality is best demonstrated with [`unpack()`](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#input-functions-and-unpack). However, when no functions are involved, you can simply use the Python unpacking operator, `**`. I was able to confirm that this works for the `resources` directive. 

Here was my test snakefile:
```python
custom_resources = {"mem_mb": 1500, "walltime": 24}

rule test_rule:
    output: "output.txt"
    threads: 2
    resources: **custom_resources
    # Thanks to unpacking, the above line is equivalent to:
    # resources:
    #     mem_mb: 1500
    #     walltime: 24
    shell: "echo Threads: {threads} / Resources: {list(resources.keys())} > {output}"
```

After running the rule, `output.txt` contained `Threads: 2 / Resources: 2 1 1500 24`. Normally, you would refer to a specific resource (_e.g._ `{resource.mem_mg}`), but here, I wanted to see everything tracked as a resource. Note that the first two numbers are internally tracked resources, `_cores` and `_nodes`, which we can ignore here. 

Given this functionality, I propose that future modules stop using the `mem_mb` field in the configuration YAML file and start using a `resources` field. Under this new field, there would be keys for each rule (like `mem_mb`), but the values paired with these keys would be a set of key-value pairs instead of integers. 

Let's take the manta-2.3 module as an example. The default configuration file would change as follows. Note that I took the liberty to move the `bam` resource from the Snakefile to the YAML file since this new format can now handle an arbitrary number of resources. 

```
[...]

PR #83, opened 2020-08-24
Module/mutect2/1.0

ppararaj:
# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable


whelena:
is this module ready for review?

workflows/reference_files/2.4/reference_files_header.smk
```diff
@@ -0,0 +1,366 @@
...
+##### CONFIG #####
+localrules: download_genome_fasta,
+            download_main_chromosomes, 
+            download_gencode_annotation,
+            hardlink_download, 
+            update_contig_names,
+            get_genome_fasta_download, 
+            index_genome_fasta,
+            get_main_chromosomes_download, 
+            get_gencode_download
```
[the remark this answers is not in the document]
ppararaj: Added `download_af_only_gnomad_vcf` as a local rule.

modules/mutect2/1.0/mutect2.smk
```diff
@@ -0,0 +1,185 @@
...
+    shell:
+        "samtools view -H {input.tumour_bam} | grep '^@RG' | "
+        r"sed 's/.*SM:\([^\t]*\).*/\1/g'"" | uniq > {output.tumour_sm} 2> {log.stderr}"
+        " && "
+        "samtools view -H {input.normal_bam} | grep '^@RG' | "
+        r"sed 's/.*SM:\([^\t]*\).*/\1/g'"" | uniq > {output.normal_sm} 2> {log.stderr}"
+
+
+# Launces Mutect2 in paired mode
+rule _mutect2_run:
```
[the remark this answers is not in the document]
ppararaj: Added parallelization by chromosome

modules/mutect2/1.0/mutect2.smk
```diff
@@ -95,18 +107,100 @@ rule _mutect2_run:
...
+        CFG["dirs"]["mutect2"] + "{{seq_type}}--{{genome_build}}/{{tumour_id}}--{{normal_id}}--{{pair_status}}/chromosomes/{chrom}.output.vcf.gz.tbi",
+        chrom = chrs
+    )
+    return(tbis)
+
+
+# Merge chromosome mutect2 VCFs from the same sample
+rule _mutect2_merge_vcfs:
+    input:
+        vcf = _mutect2_get_chr_vcfs,
```
rdmorin, line 171: To avoid code duplication, couldn't you forego the redundant function to retrieve tbi files and instead just modify each string in the list returned by the first function using a lambda function that concatenates ".tbi". In fact, why are we even listing the tbi as an input for this?

ppararaj: I was observing that when using scratch subdirectories the indices were not appearing alongside the VCF files unless they were inputs in the merge step, and the indices are needed for bcftools concat to work. I tried outputting both VCF and .tbi files from the same input function as a dictionary and using `unpack()`, but I was getting an error stating that the `vcf` key did not exist.

ppararaj: I went back to see if it was possible to use a single input function and it seems to work now. I probably used a different key name than `vcf`.

modules/mutect2/1.0/mutect2.smk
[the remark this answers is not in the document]
ppararaj: Added no normal rules

modules/mutect2/1.0/mutect2.smk
```diff
@@ -0,0 +1,296 @@
+#!/usr/bin/env snakemake
```
lkhilton, line 1: I have one last request. Per #70, could you please change all instances of 
```
resources: 
    mem_mb = CFG["rulename"]["mem_mb"]
```

to the new unpacking style? It lets users feed in additional resource parameters.

PR #84, opened 2020-08-26
Cram bam2fastq

rdmorin:

I have updated the changelog to describe my changes. This is a minor but important update to allow handling of cram files.


rdmorin:
@whelena can you check that this won't break functionality of other tools that rely on its output? Since we specify the file names in the config I hope the "fastq.gz" naming scheme will be fine.

rdmorin:
@whelena are you finished modifying this version? If so, can you approve the PR so it can be merged?

PR #85, opened 2020-08-27
Module/controlfreec/1.0

Jwong684:
# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable

To be completed.


Jwong684:
It works on a lowpass-WGS dataset, but currently only has one output directory (all output dependent on the config_WGS.txt file).

Bedtools, samtools, sambamba are currently dependent on the user's local versions of these programs. (i.e. when generating the initial config_WGS.txt file, I use "which bedtools" , "which samtools", and "which sambamba" to define the path to these programs in the config.

demo/Snakefile
```diff
@@ -8,13 +8,14 @@ import oncopipe as op
...
 SAMPLES = op.load_samples("data/samples.tsv")
 
-
 ##### REFERENCE_FILES WORKFLOW #####
 
 
 subworkflow reference_files:
     workdir:
-        "reference/"
+        # "/projects/bgrande/reference_files/"
```
rdmorin: The demo Snakefile shouldn't have any lines commented out and should have all paths restored to how they are set in the Master branch. Remember your changes eventually get merged to become the new demo.

demo/config.yaml
```diff
@@ -1,10 +1,10 @@
 lcr-modules:
 
     _shared:
-        lcr-modules: "../"
-        lcr-scripts: "../../lcr-scripts"
+        lcr-modules: "/projects/rmorin/projects/tumour_contam/lcr-modules/"
```
rdmorin: Restore all paths to default. If this branch is relatively up to date with Master the defaults should all work. Not sure why you changed these.

demo/config.yaml
```diff
@@ -1,10 +1,10 @@
...
+        lcr-scripts: "/projects/rmorin/projects/tumour_contam/lcr-scripts/"
         root_output_dir: "results/"
-        scratch_directory: "scratch/"
+        scratch_directory: "/projects/rmorin_scratch/jawong_tmp/"
```
rdmorin: It would be easier to just symlink "/projects/rmorin_scratch/jawong_tmp/" to ./scratch so you don't have to change the config

demo/config.yaml
```diff
@@ -33,3 +33,11 @@ lcr-modules:
             # if using manta output, use vcf file in the 99-outputs subdirectory and ensure manta version corresponds to the loaded module
             candidate_small_indels: "results/manta-2.2/99-outputs/vcf/{seq_type}--{genome_build}/candidateSmallIndels/{tumour_id}--{normal_id}--{pair_status}.candidateSmallIndels.vcf"
 
+    controlfreec:
+        inputs:
+            sample_bam: "data/{sample_id}.bam"
+            sample_bai: "data/{sample_id}.bam.bai"
+        logs:
+            controlfreec: "logs/"
```
rdmorin: Doesn't LCR modules have a place for logs that is tool-specific (in the output directories)?

demo/data/samples.tsv
```diff
@@ -3,3 +3,4 @@ TCRBOA7-N-WEX	capture	TCRBOA7	normal	grch37
 TCRBOA7-T-WEX	capture	TCRBOA7	tumour	grch37
 TCRBOA7-ALT-T-WEX	capture	TCRBOA7-ALT	tumour	grch37
 TCRBOA7-T-RNA	mrna	TCRBOA7	tumour	grch37
+05-28495-B82177	genome	05-28495-B82177	tumour	grch37
```
rdmorin: I believe Bruno has a genome that can go into the demo. We can't include one of ours. Let's discuss this afternoon.

modules/controlfreec/1.0/config/default.yaml
```diff
@@ -0,0 +1,72 @@
+lcr-modules:
+    
+    controlfreec:
+
+        inputs:
+            # Available wildcards: {seq_type} {genome_build} {sample_id}
+            # sample_bam: "__UPDATE__"
+            # sample_bai: "__UPDATE__"
+            sample_bam: "/projects/rmorin/projects/tumour_contam/lcr-modules/demo/data/{sample_id}.bam"
```
rdmorin: Change these back to the UPDATE lines (i.e. comment these and uncomment the ones above. You also don't need the full path and should just start with "data/". Then the example will be more generic and this will be more portable.

modules/controlfreec/1.0/controlfreec.smk
```diff
@@ -0,0 +1,157 @@
...
+        CFG["conda_envs"]["sambamba"]
+    params:
+        config = CFG["options"]["configFile"],
+        outdir = CFG["dirs"]["run"] + "{seq_type}--{genome_build}/{sample_id}/",
+        window = CFG["options"]["window"],
+        ploidy = CFG["options"]["ploidy"],
+        chrLen = CFG["options"]["chrLenFile"],
+        chrFiles = CFG["options"]["chrFiles"],
+        reference = CFG["options"]["gemMappabilityFile"]
+    shell:
```
rdmorin, line 135: Are all these tools available in a single conda environment? If not, you need to make a conda environment that has them all available and supply it with this module.

modules/controlfreec/1.0/controlfreec.smk
```diff
@@ -0,0 +1,157 @@
...
+        "sed \"s|referenceFile|{params.reference}|g\" > {output}"
+
+rule _controlfreec_run:
+    input:
+        CFG["dirs"]["run"] + "{seq_type}--{genome_build}/{sample_id}/config_WGS.txt"
+    output:
+        CFG["dirs"]["run"] + "{seq_type}--{genome_build}/{sample_id}/{sample_id}.bam_info.txt",
+        CFG["dirs"]["run"] + "{seq_type}--{genome_build}/{sample_id}/{sample_id}.bam_ratio.txt",
+        CFG["dirs"]["run"] + "{seq_type}--{genome_build}/{sample_id}/{sample_id}.bam_CNVs"
+    conda: CFG["conda_envs"]["controlfreec"]
```
rdmorin, line 173: I assume this doesn't have sambamba, samtools etc available. Please confirm.

modules/controlfreec/1.0/controlfreec.smk
```diff
@@ -0,0 +1,157 @@
...
+        CFG["dirs"]["run"] + "{seq_type}--{genome_build}/{sample_id}/{sample_id}.bam_CNVs.p.value.txt"
+    params:
+        calc_sig = CFG["software"]["FREEC_sig"]
+    threads: CFG["threads"]["calc_sig"]
+    resources: mem_mb = CFG["mem_mb"]["calc_sig"]
+    conda: CFG["conda_envs"]["controlfreec"]
+    log:         
+        stdout = CFG["logs"]["run"] + "{seq_type}--{genome_build}/{sample_id}/calc_sig.stdout.log",
+        stderr = CFG["logs"]["run"] + "{seq_type}--{genome_build}/{sample_id}/calc_sig.stderr.log"
+    shell:
```
rdmorin, line 197: If this is an R environment should it be separate from the controlfreec executable's environment?

modules/controlfreec/1.0/controlfreec.smk
```diff
@@ -0,0 +1,157 @@
...
+    conda: CFG["conda_envs"]["controlfreec"]
+    threads: CFG["threads"]["controlfreec_run"]
+    resources: mem_mb = CFG["mem_mb"]["controlfreec_run"]
+    log:
+        stdout = CFG["logs"]["run"] + "{seq_type}--{genome_build}/{sample_id}/run.stdout.log",
+        stderr = CFG["logs"]["run"] + "{seq_type}--{genome_build}/{sample_id}/run.stderr.log"
+    shell:
+        "freec -conf {input} > {log.stdout} 2> {log.stderr} "
+
+rule _controlfreec_calc_sig:
```
rdmorin, line 183: I don't see how this rule is ever run. It's outputs don't come up in latter rules (or the all rule). What am I missing?

modules/controlfreec/1.0/etc/scripts/makeGraph.R~
```diff
@@ -0,0 +1,134 @@
+args <- commandArgs()
+
+dataTable <-read.table(args[5], header=TRUE);
```
rdmorin: I don't think this file needs to be here.

modules/controlfreec/1.0/controlfreec.smk
Jwong684: Ah you're right. I'll need to add the output into the all rule

modules/controlfreec/1.0/etc/scripts/makeGraph.R~
Jwong684: Deleted

modules/controlfreec/1.0/controlfreec.smk
Jwong684: Compiled all environments into one (includes samtools, sambamba, bedtools with freec)

modules/controlfreec/1.0/controlfreec.smk
Jwong684: Compiled all envs into one

modules/controlfreec/1.0/config/default.yaml
Jwong684: Changed

demo/config.yaml
Jwong684: Removed

demo/config.yaml
Jwong684: Reverted back

demo/config.yaml
Jwong684: Reverted back

demo/Snakefile
Jwong684: Uncommented all

modules/controlfreec/1.0/controlfreec.smk
Jwong684: I haven't ran into any issues with it so far. Controlfreec automatically includes the R environment when being installed

Jwong684:
Updated reference files in module.

It will now download/extract .gem mappability files, split genome.fa into the individual chr components, and also generate a chromLength txt file from the genome.fa.fai

Performed test-runs to completion for WGS files under grch37 and grch38

rdmorin:
@Jwong684 can you try to address the conflicts?

Jwong684:
I fixed the conflicts for the three files and pushed them to my branch.

I think I pulled everything from master into my branch (I used git checkout master, then git pull -ff-only) so all additional files were then staged into my latest commit. I don't think it had any effect other than writing a bunch of files from master into my local repo.

PR #98, opened 2020-09-04
Module/salmon/1.0

Kdreval:
# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable

- [X] I ensured dry dun of demo setup executes without errors.

- [X] I checked the master branch and merged the latest master into the development branch.


demo/data/samples.tsv
```diff
@@ -1,8 +1,4 @@
 sample_id	seq_type	patient_id	tissue_status	genome_build
-99-20595T	capture	99-20595	tumour	hg38
-99-20595N	capture	99-20595	normal	hg38
-08-20595T	capture	08-20595	tumour	hg38
-08-20595N	capture	08-20595	normal	hg38
-TCRBOA7-N-WEX 	capture 	TCRBOA7 	normal 	grch37
-TCRBOA7-T-WEX 	capture 	TCRBOA7 	tumour 	grch37
```
rdmorin, line 7: Why are the exome files now missing from this?

modules/salmon/1.0/config/default.yaml
```diff
@@ -0,0 +1,26 @@
...
+    
+    salmon:
+
+        inputs:
+            # Available wildcards: {seq_type} {genome_build} {sample_id}
+            sample_fastq_1: "__UPDATE__"
+            sample_fastq_2: "__UPDATE__"
+
+        options:
+            quant: "-l A"
```
rdmorin, line 12: Can you add a comment to explain what this option does? What about library type (stranded, unstranded)?

modules/salmon/1.0/salmon.smk
```diff
@@ -0,0 +1,103 @@
...
+##### RULES #####
+
+
+# Symlinks the input files into the module results directory (under '00-inputs/')
+rule _salmon_input_fastq:
+    input:
+        fastq_1 = CFG["inputs"]["sample_fastq_1"],
+        fastq_2 = CFG["inputs"]["sample_fastq_2"]
+    output:
+        fastq_1 = CFG["dirs"]["inputs"] + "fastq/{seq_type}--{genome_build}/{sample_id}.R1.fastq",
```
rdmorin: Aren't we planning on keeping fastq gzipped? My recent change to bam2fastq should output them as gzipped. Shouldn't the symlinks be named "fastq.gz"?

rdmorin (changes requested):
I've left some comments. I also can't see what you changed in reference_files.smk. Can you clarify what the differences are between this and Master?

demo/data/samples.tsv
Kdreval: I looked on what is the status of this table in branch demo repairs, and thought to match it. I will add the exomes back into this file.

modules/salmon/1.0/config/default.yaml
Kdreval: when `-l A' is added, it tells salmon that it should automatically determine the library type of the sequencing reads (e.g. stranded vs. unstranded etc.).

Kdreval:
The changes in `reference_files.smk` that are in this PR were introduced by Helena in [this](https://github.com/LCR-BCCRC/lcr-modules/pull/98/commits/8e4d6fccecc04aca9d425e9c0470529330bf9285#diff-7c197596982fc7adb950689a408ee08c) commit

modules/salmon/1.0/config/default.yaml
rdmorin: Great. Maybe add a comment below in the config to say that's what it does?

modules/salmon/1.0/salmon.smk
rdmorin: Great

rdmorin:
> The changes in `reference_files.smk` that are in this PR were introduced by Helena in [this](https://github.com/LCR-BCCRC/lcr-modules/pull/98/commits/8e4d6fccecc04aca9d425e9c0470529330bf9285#diff-7c197596982fc7adb950689a408ee08c) commit

Can you confirm they're compatible with your Salmon module and generate the reference index that you need?

demo/data/samples.tsv
Kdreval: wait, I think there is an interface issue. There are currently 3 samples in the in the table: 2 exomes and 1 mrna from the same tumour. These capture samples that you referred to (lines 6 and 7) are shown as removed because when I was testing dry run for demo, it again gave that strange issue of completion without executing any jobs. Therefore, I removed the hidden symbols from these lines and made sure they are tab-separated.

rdmorin:
Also, does this require the (pending) changes to reference_files version 2.4 to work?

Kdreval:
>Can you confirm they're compatible with your Salmon module and generate the reference index that you need?
> 
Yes.  I have tested these transcriptomes today by checking proper transcripts by BLAT, and also used these to run mrna samples (some cell lines for which fastqs were available in GAMBL). I also ensured that Salmon outputs after using these transcriptomes are compatible with `tximport`.

Kdreval:
> Also, does this require the (pending) changes to reference_files version 2.4 to work?

No, it does not. I think that branch can be deleted. This PR does include changes Prasath did for reference files 2.4 for his module. I don't know if other people edited the 2.4 version too, but this PR is merged with current master (`git merge` output is `Already up to date`), so I _think_ everyone else's changes should not be affected.

demo/data/samples.tsv
rdmorin: The file only has four lines:

https://github.com/LCR-BCCRC/lcr-modules/blob/afe21b19ceb683a454fffeeb8fcc19ea874c8d5e/demo/data/samples.tsv

rdmorin: Wasn't there another sample for the liftOver part or has it been renamed to match the TCRB0A7 one?

Kdreval: Oh, ok. I will return back the 4 hg38 exome samples I deleted from this table. For the liftOver, yes, I renamed the sample to match the TCRB0A7.

rdmorin: I think the (fake) hg38 exomes are important because liftOver should only run on hg38 outputs, right?

Kdreval: I returned these samples back

modules/salmon/1.0/config/default.yaml
Kdreval: I have added a comment in config

modules/salmon/1.0/salmon.smk
Kdreval: I have updated it to indicate fastq compression

rdmorin (approved):

PR #106, opened 2020-09-12
Cram bam2fastq

Kdreval:
Changes introduced include handling of genome build wildcard, and marking target files under rule output. These changes also required adjustment of salmon module, to ensure these 2 modules can work together. Also, added minor adjustments for running it on cluster - increased `mem_mb` because some files require much more memory than initially was allocated, and added `bam=1` to enable reducing number of concurrent jobs.


modules/bam2fastq/1.1/bam2fastq.smk
```diff
@@ -33,35 +33,46 @@ localrules:
 ##### RULES #####
 
 
-# Symlinks the input files into the module results directory (under '00-inputs/')
+def get_genome_build(wildcards):
```
rdmorin, line 36: Why does this need to be done?

modules/bam2fastq/1.1/bam2fastq.smk
```diff
@@ -33,35 +33,46 @@ localrules:
...
-        op.relative_symlink(input.bam, output.bam)
+        op.relative_symlink(input, output.bam)
 
 
 # Conditional rules depending on whether or not fastq outputs will be temporary
 if CFG["temp_outputs"] == True:
     rule _bam2fastq_run:
         input:
             bam = str(rules._bam2fastq_input_bam.output.bam),
-            genome = reference_files("genomes/{genome_build}/genome_fasta/genome.fa")
```
rdmorin, line 51: I'm struggling understanding the rationale or need for this change

rdmorin: Is it somehow a way to avoid having genome-build end up in the output file names?

modules/bam2fastq/1.1/config/default.yaml
```diff
@@ -17,7 +17,7 @@ lcr-modules:
             bam2fastq: 4
 
         mem_mb:
-            bam2fastq: 5000
+            bam2fastq: 100000
```
rdmorin, line 20: This alone might be justification for switching to the samtools equivalent of this instead of using Picard. I suspect the limitation with using samtools is that the bam would need to be re-sorted on read name first.

Kdreval: I agree. I tested samtools on handful of bams that were giving errors with picard, and samtools worked much faster and did not need that much memory. I did not need to re-sort, however that might not be the case for other samples. As I we are going to use reference fasta file from the header, it might be easier to use samtools instead of picard

rdmorin: How can samtools export read 1 and read 2 in order if the bam is not readsorted? I thought Picard did this in memory (hence the huge RAM burden)

Kdreval: Yes, samtools need bam files to be sorted. I thought that the bam files we have are already sorted.

rdmorin: There are two ways to sort a bam file. Read name or position. All ours are sorted on position. I suspect they would need to be re-sorted on read name

Kdreval: It is true, they will need to be resorted by position. I think I will modify the whole bam2fastq_run rule and use samtools instead, with read-sorting them in advance.

rdmorin: They would need to be sorted by read name. If you tackle this I suggest you and I chat about how to get samtools to sort efficiently. The default produces a lot of small tempfiles and causes a lot of I/O

rdmorin: I am also slightly concerned there's a fundamental reason why samtools wasn't the first choice. Maybe we should ask Bruno on Monday

rdmorin (commented):
Let's hold off re-implementing this with samtools for now

rdmorin (approved):

PR #114, opened 2020-10-01
Module/shimmer/1.0

Jwong684:
# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable

- [] I added rules to the `reference_files` workflow to generate any new reference files.

To be completed.

### DIFF
--- a/modules/shimmer/1.0/shimmer.smk
+++ b/modules/shimmer/1.0/shimmer.smk
@@ -1,0 +1,125 @@
+#!/usr/bin/env snakemake
+
+
+##### ATTRIBUTION #####
+
+
+# Original Author:  N/A
+# Module Author:    Jasper
+# Contributors:     N/A
+
+
+##### SETUP #####
+
+
+# Import package with useful functions for developing analysis modules
+import oncopipe as op
+
+# Setup module and store module-specific configuration in `CFG`
+# `CFG` is a shortcut to `config["lcr-modules"]["shimmer"]`
+CFG = op.setup_module(
+    name = "shimmer",
+    version = "1.0",
+    # TODO: If applicable, add more granular output subdirectories
+    subdirectories = ["inputs", "shimmer", "outputs"],
+)
+
+# Define rules to be run locally when using a compute cluster
+# TODO: Replace with actual rules once you change the rule names
+localrules:
+    _shimmer_input_bam,
+    _shimmer_step_2,
+    _shimmer_output_vcf,
+    _shimmer_all,
+
+
+##### RULES #####
+
+
+# Symlinks the input files into the module results directory (under '00-inputs/')
+# TODO: If applicable, add an input rule for each input file used by the module
+rule _shimmer_input_bam:
+    input:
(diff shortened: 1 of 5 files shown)


rdmorin (commented):
Where's the changelog?

Jwong684:
Updated with Changelog and missing files (/etc/, environment yaml, config yaml)

modules/shimmer/1.0/envs/shimmer.env.yaml
```diff
@@ -0,0 +1,142 @@
+name: null
```
lkhilton, line 1: This yaml file should be in `lcr-modules/envs/shimmer/shimmer.yaml` and there should be a symlink from `modules/shimmer/1.0/envs` pointing to the actual yaml file.

modules/shimmer/1.0/etc/merge_vcfs.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
lkhilton, line 1: Could you remove this script? I see that the per-chromosome idea failed (rats!) but you shouldn't merge vcf files like this anyway. You should always use `bcftools concat` followed by `bcftools sort` and `bcftools index`, or else downstream tools will complain/break.

modules/shimmer/1.0/etc/merge_vs.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
lkhilton, line 1: I see no harm in keeping this one though, since it's not a vcf and might be handy if the --regions flag is ever fixed in future releases.

modules/shimmer/1.0/shimmer.smk
```diff
@@ -0,0 +1,123 @@
...
+        bai = CFG["inputs"]["sample_bai"]
+    output:
+        bam = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{sample_id}.bam",
+        bai = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{sample_id}.bai"
+    run:
+        op.relative_symlink(input.bam, output.bam)
+        op.relative_symlink(input.bai, output.bai)
+
+
+rule _shimmer_on:
```
lkhilton, line 50: Is it possible to include a rule to build/make Shimmer from the source code you have in `etc`?

lkhilton (changes requested):
See comments.

modules/shimmer/1.0/shimmer.smk
rdmorin: That should totally be possible.

modules/shimmer/1.0/etc/merge_vcfs.sh
rdmorin: What was the problem with running this per-chromosome?

Kdreval:
Is there still a need for this module? If so, Jasper is this ready to go?

PR #115, opened 2020-10-09
Mutect2 2.0

lkhilton:
# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable

To be completed.


lkhilton:
*IMPORTANT* Before merging this PR I will be moving these files to a v2.0 of the mutect2 module and v 2.5 of the reference file workflow, so please don't merge after reviewing/approving. 

This version uses the latest mutect2 LearnReadOrientModel functionality for artifact filtering. I've also updated the reference file workflow to download from the Broad Google Cloud bucket because they will be deprecating the FTP download site.

rdmorin:
@ppararaj are you reviewing this for Laura, Please? We need this module for GAMBL and BLGSP

ppararaj (approved):
Looks good

lkhilton:
Ready for final approval!

modules/vcf2maf/1.1/vcf2maf.smk
```diff
@@ -76,6 +77,7 @@ rule _vcf2maf_run:
         mem_mb = CFG["mem_mb"]["vcf2maf"]
     shell:
         op.as_one_line("""
+        if [[ -e {output.maf} ]]; then rm -f {output.maf}; fi;
```
rdmorin, line 80: Should we add this same logic for the intermediate vep file? If it is not deleted by the rule then it actually prevents VEP from re-running. VCF2maf just assumes that it's been run and uses the old vep.vcf as input.

workflows/reference_files/2.4/reference_files_header.smk
```diff
@@ -226,19 +226,78 @@ rule download_af_only_gnomad_vcf:
...
+        file = lambda w: {
+            "grch37": "gs://gatk-best-practices/somatic-b37/small_exac_common_3.vcf", 
+            "grch38": "gs://gatk-best-practices/somatic-hg38/small_exac_common_3.hg38.vcf.gz"
+        }[w.version]
+    conda: CONDA_ENVS["gsutil"]
+    shell:
+        op.as_one_line("""
+        if [[ {params.file} == *".gz" ]]; then 
+            gsutil cp {params.file} - 2> {log}
             |
```
rdmorin, line 296: I'm not telling you to not do this but some of us cannot understand what these awk commands do, so I'm simply taking your word for it that this works. Since I do stuff in Perl here and there I know I'm not in a good position to complain.

rdmorin (approved):
Thanks for those last few tweaks.

rdmorin:
@lkhilton I leave it to you to merge this

PR #119, opened 2020-10-20
Improvements to the bam2fastq-star-utils pipeline

lkhilton:
These changes have been tested on the LSARP RNAseq data and/or capture data. Version updates have been indicated. 

### star module changes (will be migrated to v1.4): 
- Read length is obtained from the samples table so that STAR can vary the index and `sjdboverhang` parameters per sample. The samples table MUST contain a `read_length` column. If all samples have the same read length, a default value can be set by following the instructions in the [LCR modules documentation](https://lcr-modules.readthedocs.io/en/latest/for_users.html#adding-and-transforming-columns).
- Includes placeholder fastq inputs to prevent premature deletion of fastq files marked as temp in the `bam2fastq` module. 
- Touches an output file on completion of `_star_run` so that killed jobs will re-run properly.
- Enabled job grouping in a secondary Snakefile. Works with bam2fastq for efficient temp file deletion.  

### bam2fastq module changes (will be migrated to v1.2): 
- In 1.1 the sample input filename format was inflexible and designed for the GAMBL project. The function that drops the genome build has been moved to the `_bam2fastq_run` rule so that the input sample filenames don't have to adhere to a specific format. 
- Enabled job grouping in a secondary Snakefile. Can be used with STAR and/or Salmon to encourage efficient temp file deletion. 

### utils module changes (will be migrated to 2.1): 
- In 2.0, if the rule `_utils_bam_sort` was interrupted, it would error on re-run because `samtools sort` doesn't delete temporary bam files on interruption. This version writes temp bam files from `_utils_bam_sort` to a temporary directory which facilitates automatic removal of the temp sort files. 

### bwa_mem module changes (will be migrated to 1.1): 
- Added wildcard constraints, default utils-2.0 module, resource unpacking, and a placeholder "fastq_real" input to prevent premature deletion of temp fastq files. 
- Enabled job grouping as for other modules. 

### salmon module changes (will be migrated to 1.1): 
- Enable job grouping and resource unpacking.


modules/bam2fastq/1.1/bam2fastq_grouped.smk
```diff
@@ -0,0 +1,149 @@
...
+        params:
+            opts = CFG["options"]["bam2fastq"]
+        conda:
+            CFG["conda_envs"]["picard"]
+        threads:
+            CFG["threads"]["bam2fastq"]
+        resources:
+            **CFG["resources"]["bam2fastq"]
+        group: 
+            CFG["group"]["bam2fastq"]
```
Kdreval: Does the grouping also needs to be added if fastq files are not marked as temporary?

modules/star/1.3/star.smk
```diff
@@ -48,38 +52,56 @@ rule _star_input_fastq:
...
-        fastq_2 = CFG["dirs"]["inputs"] + "fastq/{seq_type}--{genome_build}/{sample_id}.R2.fastq.gz",
+        fastq_2 = CFG["dirs"]["inputs"] + "fastq/{seq_type}--{genome_build}/{sample_id}.R2.fastq.gz"
     run:
         op.relative_symlink(input.fastq_1, output.fastq_1)
         op.relative_symlink(input.fastq_2, output.fastq_2)
 
+# Function to retrieve read length from sample table
+def get_overhang(wildcards,build = False):
+    tbl = config["lcr-modules"]["star"]["samples"]
+    # read_length = tbl[(tbl.sample_id == wildcards.sample_id) & (tbl.seq_type == wildcards.seq_type)]["read_length"].values()[0]
```
Kdreval: this line is probably temporary?

schemas/read_length/read_length-1.0.yaml
```diff
@@ -0,0 +1,11 @@
...
+
+description: Required fields for each sample row in the samples data frame
+
+properties:
+    read_length:
+        type: integer
+        description: RNAseq read length (usually 50, 75, 100, or 150 bp)
+
+required:
+    - read_length
```
Kdreval: it is shown here as no new line

Kdreval:
I left few minor questions. There are others general, too:
1. I thought we were going to add also a column of library strandness to be implemented in STAR. Not sure if you wanted to add it here, or that is independent modification we will do later.
2. When migrating to the new version, would you also please include all these changes to demo workflow?

modules/bam2fastq/1.1/bam2fastq_grouped.smk
lkhilton: I can't really envision a scenario when it would be needed, but this still gives users the flexibility.

modules/star/1.3/star.smk
lkhilton: Yes, thanks I'll delete that line when I update this to the new version.

lkhilton:
Re: the strand field, I don't think STAR needs strand information unless the library is unstranded and you want to later run cufflinks or cuffdiff. Our current module doesn't make any assumptions about strandedness. It can be added as a parameter to other modules that need it though - I think Ryan said Picard QC needs it, and you maybe mentioned Kallisto? Individual field requirements are added via yaml files in the `schemas` directory, and the module won't run if that information isn't made available, so I think we should avoid adding unnecessary schemas to modules. 

I will incorporate these into the demo workflow.

rdmorin:
Is this ready for re-review?

lkhilton:
I just need to finish updating the versions and the demo. Hopefully today. Do you need it?

rdmorin:
I don't. Just wondering if it had stalled somewhere on the review end.

lkhilton:
@Kdreval this is ready for re-review please! Everything is in the demo and besides the mysterious `touch()` problem I'm having it works.

demo/config.yaml
```diff
@@ -79,7 +77,7 @@ lcr-modules:
     utils:
         inputs:
             bed: 
-                grch37: "__UPDATE__" # make sure this corresponds with config["lcr-modules"]["picard_qc"]["inputs"]["intervals"]
+                grch37: "/projects/dscott_prj/CCSRI_1500/exomes/ref/agilent/hg19/target_regions.nochr.bed" # make sure this corresponds with config["lcr-modules"]["picard_qc"]["inputs"]["intervals"]
```
Kdreval: I was using `__UPDATE__` here so that if tested outside GSC and not updated to the user-specific file, the oncopipe will exit pointing out to this. Otherwise, it will be showing error of missing file, and it might give hard time outside users to see where this missing file is coming from. Do you think it should be reverted to `__UPDATE__` when testing is done? Otherwise, I think it might be helpful to add another comment here for outside users.

Kdreval (approved):

demo/config.yaml
lkhilton: OK I think the actual solution here is to include the bed file in the repo. Otherwise the demo isn't really runnable, and there's nothing in the readme to help external users find this file. I'll put a copy there before I merge this. Thanks for the review!

lkhilton:
@Kostia after squashing what seems like a never-ending number of bugs, this is ready for final review and merging :scienceparrot:

Kdreval (approved):

PR #124, opened 2020-10-28
Module/lofreq/1.0

Kdreval:
# Pull Request Checklists

- [X] I updated the `CHANGELOG.md` file associated with the module and explained which modifications and why they were made.

- [X] I updated the demo project according to the modifications I introduced to the module.

- [X] I updated tested the new modifications both in dry run and in snakemake run of demo sample.

- [X] I merged most recent master branch and checked to resolve any possible conflicts.


modules/lofreq/1.0/lofreq.smk
```diff
@@ -143,7 +162,8 @@ rule _lofreq_all:
             genome_build=CFG["runs"]["tumour_genome_build"],
             tumour_id=CFG["runs"]["tumour_sample_id"],
             normal_id=CFG["runs"]["normal_sample_id"],
-            pair_status=CFG["runs"]["pair_status"])
+            pair_status=CFG["runs"]["pair_status"],
+            tool="lofreq")
```
rdmorin: I would be surprised if this worked for more than one sample. For Strelka I had to use this syntax (you should double-check):

tool = "lofreq" * len(CFG["runs"])

modules/lofreq/1.0/src/bash/lofreq_filter.sh
```diff
@@ -0,0 +1,12 @@
...
+# This script will filter out the non-standard chromosomes and non-ACTG characters from
+# vcf files produced by lofreq. It accepts .qz-compressed files.
+# Usage: 
+#   lofreq_filter.sh input.vcf.gz | gzip > output.vcf.gz
+
+INPUT_FILE="$1"
+
+zcat "${INPUT_FILE}" \
+	| awk '($4=="A" || $4 == "C" || $4=="T" || $4=="G" || /\#/)' \
+	| grep -E -v "v1|v2|GL0"
```
rdmorin: This is a bit hacky. Shouldn't we really be checking for exact match of the first column to one of the chromosomes in the main chromosome list?

rdmorin (changes requested):
One thing I think is rather major (in the expand statement at the end). The rest is probably minor and could be improved later.

modules/lofreq/1.0/lofreq.smk
Kdreval: yes, you are right. It works only for one sample.

lkhilton: I don't think it's necessary to have it as a wildcard here anyway. It can be called as a wildcard in downstream tools without it being specified here.

rdmorin: That's a really good point. I had done it this way in the liftover tool because it made sense to have it as a wildcard there. Here it could just be explicitly "strelka" throughout. That said, I suppose there's no harm in this as long as it is expanded out to be repeated the right number of times. The only (minor) benefit of doing it this way is the changes you make to the code will be identical in the other variant callers. Aside: if the config stores the tool name then it would be even more elegant to use it from the config instead of hard-coding it in the expand statement. That would make the code even more copy-paste-able for modifying Mutect, Strelka etc.

Kdreval: Yes, I agree this wildcard is not explicitly needed here. I thoght more of a "placeholder" here, since we might later want to include the version of the module as well. For that purpose, I will move this to config, so it will be easily modifiable in the future.

modules/lofreq/1.0/src/bash/lofreq_filter.sh
Kdreval: Sound like a time for me to learn details of regex in awk. I changed it to the following `awk -e '$1 ~ /^#|^(chr)*[1-9]$|^(chr)*[1-2][0-9]$|^(chr)*X$/ {print$0}'` and tested on both grch37 and hg38 files. Currently running demo project to make sure it is working in the whole run.

rdmorin: Nice!
If I understand correctly, this prints any header line and any line in which the first column matches the pattern chr (optional) plus numeric or X. Dropping chrY is intentional, right?

rdmorin: Notably, it's a lot more succinct if we do this with Perl:
```
perl -ne 'print if /^#/ || /^(chr)*\d+\s.+/'
```

modules/lofreq/1.0/config/default.yaml
```diff
@@ -8,6 +8,8 @@ lcr-modules:
             sample_bai: "__UPDATE__"
             lofreq_filter: "{MODSDIR}/src/bash/lofreq_filter.sh"
 
+        tool: "lofreq" # included into the naming of output files, cand be updated to contain the module version
```
rdmorin: Isn't the tool name (i.e. "lofreq") already in the config at a higher level? I could be wrong.

Kdreval: It is, yes. I can simply do `CFG` here, but that would not be customizable. I think I created more confusion than the benefits by including this wildcard. I thought we were considering adding the tool name (and, in the future, possibly tool's version), so if the files get moved out of the directory, this information will follow with the file. My intention here was to create a placeholder, and if we will decide later to include the module version, it can be easily done here. I think at this point, it will be better if I explicitly say the tool name in the file naming, and we can modify this to be a wildcard at a later point, if needed.

rdmorin (approved):

Issue #128, opened 2020-10-31
LoFreq VCF misuse of AF


rdmorin:
The Lofreq somatic outputs use AF to denote variant allele frequency in the tumour. This is very problematic if we ever want to use the Gnomad VCF with AF information to annotate these VCFs. The (silent) behaviour is for 'bcftools annotate' to substitute the VAF (mis-named AF) in the lofreq VCF with the Gnomad AF. For any variants not in Gnomad, they keep the original value. In other words, common and rare variants have the population frequency stored in AF and the rest have VAF stored there. This is obviously hugely problematic. I propose an easy fix. The following perl command in a pipe from (and to) bcftools view should replace the AF with VAF in the header and every line. 

```
 perl -ne 's/AF=/VAF=/g;s/ID=AF/ID=VAF/;print;'
```

rdmorin:
Another solution using bcftools is to drop the AF completely but there are clearly drawbacks to this. I will point out that the DP and DP4 columns can be used to regenerate the VAF if needed. My bcftools skills are not strong but I suspect there is a way to recalculate this with bcftools and add it back as a new (e.g. VAF) annotation. 

```
bcftools annotate -x INFO/AF lofreq.vcf.gz | bcftools view -O z -o lofreq.noAF.vcf.gz
bcftools index -t lofreq.noAF.vcf.gz
```

Kdreval:
I think the first solution of renaming AF to VAF is an option to go. I will implement this together with the fix for the killed runs

Kdreval:
This issue has been addressed with PR [# 129](https://github.com/LCR-BCCRC/lcr-modules/pull/129) and can be closed.

PR #129, opened 2020-11-03
Lofreq patches

Kdreval:
I have added the line to filtering script that substitutes `AF` for `VAF` in vcf files (header + each variant) to address the issue [# 128](https://github.com/LCR-BCCRC/lcr-modules/issues/128). Most importantly, I modified `_lofreq_run` rule to ensure that killed runs are being relaunched. I have tested the following conditions:
1. Starting the run on empty directory - works as expected and runs to completion
2. Starting the run when partial output files are present, like `normal_*` and/or `tumor_*`  - deletes the partial output files (removing all files in directory, but keeping the parental directory), works as expected, and runs to completion
3. Starting run on the completed outputs - run is not executed, Snakemake shows `Nothing to be done` message, and all output files are in place.


rdmorin (changes requested):
Does the demo have the rm_files parameter used? If not, have you documented in the config what this can be set to? 0 or 1?

Kdreval:
No, the demo does not have `rm_files` used. It is not specified in config. The `{params.rm_files}` is the large output file that is deleted at the end of lofreq run. I used it here as one of the "checkpoints" to access whether the run is completed or not, because it is one of the first to be created when lofreq is running. Also, since we are deleting it at the end of the run, if it is still present in the directory, it indicates lofreq run was not completed sucessfully.

rdmorin:
Why not put the rm_files in the demo? Is it because it includes a full path to the output? Could you at least put an example (commented out) in the config so users know what to put there?

Kdreval:
My intention was to handle automatically the check of whether run was completed, and not to require user input. It is not in either of demo config, or module config for now. I can move it to config, but I thought that can potentially create a situation where the user will delete actual outputs. I will move the `rm_files` to config, with specifying the target file as it is now in snakemake, and will not use `__UPDATE__` there, so it can be run without modification, but will give a flexibility for the user, and I will add comments to document it.

lkhilton:
Wait, wait. `params.rm_files` is just the path to the large normal_relaxed.vcf.gz file that we agreed needed to be cleaned up. I don't think it needs to be user-configurable or in the config. Kostia is just using it as a check on whether the lofreq run has started, and if it has but the final target file hasn't been generated yet then the output directory is deleted.

rdmorin:
I still don't follow. How is Kostia specifying that param? Where does it come from?

Kdreval:
It currently comes from parameters of the same rule. It would be line 67 that is `rm_files = CFG["dirs"]["lofreq"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/normal_relaxed.vcf.gz"`

rdmorin:
Oh. I missed that. Sorry for the confusion. This is fine as-is

rdmorin (approved):

PR #131, opened 2020-11-03
Make relative symlinks truly relative and add absolute_symlink function

lkhilton:
The purpose of this change is to allow `op.relative_symlink()` to make symlinks truly relative. Previously [this line](https://github.com/LCR-BCCRC/lcr-modules/blob/cb2b9cc61ad1e3b704aa446caff4cc87b05b646d/oncopipe/oncopipe/__init__.py#L225) used `os.path.realpath()` to determine the destination directory. If the destination directory had a symlink anywhere in its path, the relative symlink would track back to the last common directory, usually outside of the project directory and definitely outside of the module results directory. Removing that line allows the symlinks to be made truly relative. No changes to the cookiecutter module were required because the condition `if not os.path.isabs(src)` was already true for the non-absolute paths used by lcr-modules. I still added a `force_relative` condition so that symlinks can be made relative if the source path is absolute.

I've tested this on a couple of lcr-modules by deleting and initiating regeneration of the symlinks in `99-outputs` and I can confirm that it now generates truly relative paths without the need for changes to any of the module Snakemake code.

### DIFF
--- a/oncopipe/oncopipe/__init__.py
+++ b/oncopipe/oncopipe/__init__.py
@@ -163,7 +163,7 @@ def set_value(value, *keys):
 
 # UTILITIES
 
-def prepare_symlinks(src, dest):
+def prepare_symlink(src, dest):
     # Coerce length-1 NamedList instances to strings
     def coerce_namedlist_to_string(obj):
         if isinstance(obj, smk.io.Namedlist) and len(obj) == 1:
@@ -192,19 +192,8 @@ def prepare_symlinks(src, dest):
     else:
         dest_dir, dest_file = os.path.split(dest)
     os.makedirs(dest_dir, exist_ok=True)
-
     dest = os.path.join(dest_dir, dest_file)
-    if os.path.lexists(dest) and os.path.islink(dest):
-        if os.path.realpath(src) == os.path.realpath(dest):
-            return
-        elif overwrite:
-            os.remove(dest)
-    assert not os.path.exists(dest), (
-        "Symbolic link already exists but points elsewhere: \n"
-        f"    Current: {dest} -> {os.path.realpath(dest)} \n"
-        f"    Attempted: {dest} -> {os.path.realpath(src)}"
-    )
-
+    
     return [src, dest]
 
 def absolute_symlink(src, dest, overwrite=True): 
@@ -222,9 +211,23 @@ def absolute_symlink(src, dest, overwrite=True):
         Whether to overwrite the destination file if it exists.
     """
     # Prepare source and destination file paths 
-    src, dest = prepare_symlinks(src, dest)
+    src, dest = prepare_symlink(src, dest)
+
+    # Handle destination file if it already exists
+    if os.path.lexists(dest) and os.path.islink(dest):
+        if os.path.realpath(src) == os.path.realpath(dest) and os.path.isabs(os.readlink(dest)):
+            return
+        elif overwrite:
+            os.remove(dest)
+    assert not os.path.exists(dest), (
+        "Symbolic link already exists but points elsewhere: \n"
+        f"    Current: {dest} -> {os.readlink(dest)} \n"
+        f"    Attempted: {dest} -> {src}"
+    )
+
     # Retrieve the absolute file path for the source file
     src = os.path.abspath(src)
+
     # Symlink the source file to the destination
     os.symlink(src, dest)
 
@@ -243,13 +246,24 @@ def relative_symlink(src, dest, overwrite=True):
         Whether to overwrite the destination file if it exists.
     """
     # Prepare source and destination file paths 
-    src, dest = prepare_symlinks(src, dest)
+    src, dest = prepare_symlink(src, dest)
+
     # Retrieve the relative file path for the source file
-    print(dest)
     dest_dir = os.path.split(dest)[0]
-    print(dest_dir)
-    print(src)
     src = os.path.relpath(src, dest_dir)
+
+    # Handle destination file if it already exists
+    if os.path.lexists(dest) and os.path.islink(dest):
+        if os.path.realpath(src) == os.path.realpath(dest) and not os.path.isabs(os.readlink(dest)):
+            return
+        elif overwrite:
+            os.remove(dest)
+    assert not os.path.exists(dest), (
+        "Symbolic link already exists but points elsewhere: \n"
+        f"    Current: {dest} -> {os.readlink(dest)} \n"
+        f"    Attempted: {dest} -> {src}"
+    )
+
     # Symlink the source file to the destination
     os.symlink(src, dest)
 
--- a/oncopipe/oncopipe/__init__.py
+++ b/oncopipe/oncopipe/__init__.py
@@ -194,7 +194,23 @@ def prepare_symlink(src, dest):
     os.makedirs(dest_dir, exist_ok=True)
     dest = os.path.join(dest_dir, dest_file)
     
-    return [src, dest]
+    return src, dest
+
+def compare_links(src, dest, overwrite):
+    # Check if the destination file exists and is a symlink
+    if os.path.lexists(dest) and os.path.islink(dest):
+        # If the destination link exists and has the same src, exit
+        if src == os.readlink(dest):
+            exit(0) 
+        elif overwrite:
+            os.remove(dest)
+    assert not os.path.exists(dest), (
+        "Symbolic link already exists but points elsewhere: \n"
+        f"    Current: {dest} -> {os.readlink(dest)} \n"
+        f"    Attempted: {dest} -> {src}"
+    )
+
+    
 
 def absolute_symlink(src, dest, overwrite=True): 
     """Creates an absolute symlink from any working directory.
@@ -212,24 +228,16 @@ def absolute_symlink(src, dest, overwrite=True):
     """
     # Prepare source and destination file paths 
     src, dest = prepare_symlink(src, dest)
-
-    # Handle destination file if it already exists
-    if os.path.lexists(dest) and os.path.islink(dest):
-        if os.path.realpath(src) == os.path.realpath(dest) and os.path.isabs(os.readlink(dest)):
-            return
-        elif overwrite:
-            os.remove(dest)
-    assert not os.path.exists(dest), (
-        "Symbolic link already exists but points elsewhere: \n"
-        f"    Current: {dest} -> {os.readlink(dest)} \n"
-        f"    Attempted: {dest} -> {src}"
-    )
-
     # Retrieve the absolute file path for the source file
     src = os.path.abspath(src)
 
+    # Check if destination file already exists and error if it does
+    compare_links(src, dest, overwrite)
+
     # Symlink the source file to the destination
     os.symlink(src, dest)
+
+    
 
 def relative_symlink(src, dest, overwrite=True):
     """Creates a relative symlink from any working directory.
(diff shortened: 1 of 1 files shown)


BrunoGrandePhD:
Is the plan to set `force_relative=True` for all calls to `op.relative_symlink()` in all rules other than perhaps the input rules?

lkhilton:
It actually turned out to be unnecessary, because the lcr-modules all use relative paths already, which means `not os.path.isabs(src)` is already satisfied. This makes sense because all the symlinks being generated by `op.relative_symlink()` looked like this: 

```
results/gambl/strelka-1.1/99-outputs/vcf/genome--grch37/00-20702T--00-20702N--matched_combined.passed.vcf.gz -> ../../../../../../../../rmorin/projects/gambl-repos/gambl-test/results/gambl/strelka-1.1/03-filtered/genome--grch37/00-20702T--00-20702N--matched/combined.passed.vcf.gz
```

So it was trying to make the symlink relative, but the line I removed was forcing the destination path to the real path.

lkhilton:
Maybe we should consider adding a `os.absolute_symlink` function for symlinks that we know will point outside the module or project?

rdmorin:
That sounds very useful.

lkhilton:
I've consolidated the work of preparing the paths for symlinks into a single function `prepare_symlinks`, and added an `absolute_symlink` function. Now `relative_symlink` will always make symlinks relative to the user-specified file paths rather than the destination file real path, and `absolute_symlink` can take relative paths and generate an absolute symlink. The later will be very useful in the context of reference files, input files, etc.

BrunoGrandePhD:
I took the day off of work and was away from my computer. I'll take a look at this within the next day or two.

oncopipe/oncopipe/__init__.py
```diff
@@ -163,22 +163,7 @@ def set_value(value, *keys):
...
-        The source file or directory path.
-    dest : str
-        The destination file path. This can also be a destination
-        directory, and the destination symlink name will be identical
-        to the source file name (unless directory).
-    overwrite : boolean
-        Whether to overwrite the destination file if it exists.
-    """
-
+def prepare_symlink(src, dest):
```
BrunoGrandePhD, line 166: Good idea on abstracting this shared logic between both symlinking functions!

oncopipe/oncopipe/__init__.py
```diff
@@ -207,23 +192,79 @@ def coerce_namedlist_to_string(obj):
     else:
         dest_dir, dest_file = os.path.split(dest)
     os.makedirs(dest_dir, exist_ok=True)
-
     dest = os.path.join(dest_dir, dest_file)
+    
+    return [src, dest]
```
BrunoGrandePhD: For future reference, you don't need parentheses or brackets when returning multiple values in Python. By default, a tuple (_i.e._ an immutable list) will be returned. You don't need to change this here though because this works fine. 

I've included an example below that demonstrates how this works and how you can use unpacking in Python to assign both values to separate variables in one go. 

```
>>> def foo():
...     return 1, 2
...
>>> foo()
(1, 2)

>>> a, b = foo()
>>> a
1
>>> b
2
```

oncopipe/oncopipe/__init__.py
```diff
@@ -207,23 +192,79 @@ def coerce_namedlist_to_string(obj):
...
+
+    # Handle destination file if it already exists
+    if os.path.lexists(dest) and os.path.islink(dest):
+        if os.path.realpath(src) == os.path.realpath(dest) and os.path.isabs(os.readlink(dest)):
+            return
+        elif overwrite:
+            os.remove(dest)
+    assert not os.path.exists(dest), (
+        "Symbolic link already exists but points elsewhere: \n"
+        f"    Current: {dest} -> {os.readlink(dest)} \n"
```
BrunoGrandePhD: Nice catch changing from `os.path.realpath()` to `os.readlink()`!

oncopipe/oncopipe/__init__.py
```diff
@@ -207,23 +192,79 @@ def coerce_namedlist_to_string(obj):
...
+        f"    Attempted: {dest} -> {src}"
+    )
```
BrunoGrandePhD: I'm trying to think of a way of eliminating this near-duplicate code. What if you moved the call to `os.path.abspath()` that comes after this (line 229) to before this block and then change the calls to `os.path.realpath()` to `os.readlink()` like you did in the error message? This would imply that the link is absolute, negating the need for the extra conditional that prevents this code block from being identical to what's in the other function. 

If this approach works, then you can wrap this block as a two-argument function that can be used in both `relative_symlink()` and `absolute_symlink()`. I can explain further if I'm not making sense.

BrunoGrandePhD (changes requested):
I have one suggestion that I would like your thoughts on, although I don't require it to be implemented, just acknowledged.

oncopipe/oncopipe/__init__.py
lkhilton: Thanks!

oncopipe/oncopipe/__init__.py
lkhilton: I think I see what you're saying and will give it a shot.

oncopipe/oncopipe/__init__.py
```diff
@@ -194,7 +194,23 @@ def coerce_namedlist_to_string(obj):
...
+            exit(0) 
+        elif overwrite:
+            os.remove(dest)
+    assert not os.path.exists(dest), (
+        "Symbolic link already exists but points elsewhere: \n"
+        f"    Current: {dest} -> {os.readlink(dest)} \n"
+        f"    Attempted: {dest} -> {src}"
+    )
+
+
```
lkhilton: @BrunoGrandePhD I've implemented your suggestion here

oncopipe/oncopipe/__init__.py
```diff
@@ -207,24 +192,80 @@ def coerce_namedlist_to_string(obj):
...
+def compare_links(src, dest, overwrite):
+    # Check if the destination file exists and is a symlink
     if os.path.lexists(dest) and os.path.islink(dest):
-        if os.path.realpath(src) == os.path.realpath(dest):
-            return
+        # If the destination link exists and has the same src, exit
+        if src == os.readlink(dest):
+            exit(0) 
         elif overwrite:
             os.remove(dest)
```
BrunoGrandePhD: You probably don't want to use `exit(0)`, which (~I think~ I confirmed this) would cause the Python process to terminate (rather than just exiting the function). In other words, I expect this to cause the Snakefile to stop running rather than skipping the symlinking. 

Essentially, this conditional was intended to skip the call to `os.symlink()`, which raises an error if the destination already exists with the same source. Originally, I just used `return` to cause the `relative_symlink()` function to end early. However, now that this conditional is within a utility function, that's no longer possible. One workaround is to do the same as if `overwrite=True`, which is to remove the symlink knowing that the source won't change. Hence, you can simplify this if-else block to something like this: 

```suggestion
        if src == os.readlink(dest) or overwrite:
            os.remove(dest)
```

oncopipe/oncopipe/__init__.py
```diff
@@ -207,24 +192,80 @@ def coerce_namedlist_to_string(obj):
     else:
         dest_dir, dest_file = os.path.split(dest)
     os.makedirs(dest_dir, exist_ok=True)
```
BrunoGrandePhD: While doing some testing, I realized that this function call will fail if the given destination path is just a file name (_i.e._ referring to the current working directory). An easy solution is converting the `dest_dir` to an absolute path right before passing it to the function. I confirmed that running `os.path.abspath()` on an empty string returns the current working directory:

```
>>> dest_dir = ""
>>> os.path.abspath(dest_dir)
'/Users/bgrande/Desktop/testdir'
```

```suggestion
    os.makedirs(os.path.abspath(dest_dir), exist_ok=True)
```

BrunoGrandePhD (changes requested):
Two changes are required here. 

I did some preliminary testing, which indicated that the functions work when run in overwrite mode (the default). I'll need the `exit(0)` removed before I can test in non-overwrite mode.

oncopipe/oncopipe/__init__.py
lkhilton: Good catch!

oncopipe/oncopipe/__init__.py
```diff
@@ -206,25 +191,79 @@ def coerce_namedlist_to_string(obj):
...
+    # Check if the destination file exists and is a symlink
     if os.path.lexists(dest) and os.path.islink(dest):
-        if os.path.realpath(src) == os.path.realpath(dest):
-            return
-        elif overwrite:
+        # If the destination link exists and has the same src, exit
+        if src == os.readlink(dest) or overwrite:
             os.remove(dest)
     assert not os.path.exists(dest), (
         "Symbolic link already exists but points elsewhere: \n"
```
lkhilton, line 213: One issue I've run into here on testing is that if the destination file exists but is a file and not a symlink, the error message doesn't work because there's no link to read. Can you think of a solution? It still generates an error, but not a helpful one.

BrunoGrandePhD: Good catch! I would add an assertion before the existing one to handle that edge case for a more informative error message:

```python
    # Raise error if the file exists and isn't a symbolic link
    assert not (os.path.exists(dest) and not os.path.islink(dest)), (
        "Destination file already exists but isn't a symbolic link: \n"
        f"    Current: {dest} \n"
        f"    Attempted: {dest} -> {src}"
    )
    # Raise error if the file exists and is a symbolic link
    assert not (os.path.exists(dest) and os.path.islink(dest)), (
        "Symbolic link already exists but points elsewhere: \n"
        f"    Current: {dest} -> {os.readlink(dest)} \n"
        f"    Attempted: {dest} -> {src}"
    )
```

BrunoGrandePhD:
Some quick testing on my end seems to work. For context, this was done in a folder containing only `file.txt`. 

```
>>> import oncopipe as op
Warning: The oncopipe package was imported outside of a snakefile. Most functions are designed to work within a snakefile. Some unexpected behaviour/errors might occur.

>>> op.relative_symlink("file.txt", "rel_link.txt")
>>> op.absolute_symlink("file.txt", "abs_link.txt")

>>> op.relative_symlink("file.txt", "rel_link.txt", overwrite=False)
>>> op.absolute_symlink("file.txt", "abs_link.txt", overwrite=False)

>>> op.relative_symlink("file2.txt", "rel_link.txt", overwrite=False)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/bgrande/repos/lcr-modules/oncopipe/oncopipe/__init__.py", line 262, in relative_symlink
    compare_links(src, dest, overwrite)
  File "/Users/bgrande/repos/lcr-modules/oncopipe/oncopipe/__init__.py", line 205, in compare_links
    assert not os.path.exists(dest), (
AssertionError: Symbolic link already exists but points elsewhere:
    Current: rel_link.txt -> file.txt
    Attempted: rel_link.txt -> file2.txt

>>> op.absolute_symlink("file2.txt", "abs_link.txt", overwrite=False)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/bgrande/repos/lcr-modules/oncopipe/oncopipe/__init__.py", line 233, in absolute_symlink
    compare_links(src, dest, overwrite)
  File "/Users/bgrande/repos/lcr-modules/oncopipe/oncopipe/__init__.py", line 205, in compare_links
    assert not os.path.exists(dest), (
AssertionError: Symbolic link already exists but points elsewhere:
    Current: abs_link.txt -> /Users/bgrande/Desktop/testdir/file.txt
    Attempted: abs_link.txt -> /Users/bgrande/Desktop/testdir/file2.txt
```

The final state of the working directory looked like this, as expected: 
```
lrwxr-xr-x  1 bgrande  staff    39B Nov  5 14:01 abs_link.txt@ -> /Users/bgrande/Desktop/testdir/file.txt
-rw-r--r--  1 bgrande  staff     7B Nov  5 13:13 file.txt
lrwxr-xr-x  1 bgrande  staff     8B Nov  5 14:01 rel_link.txt@ -> file.txt
```

lkhilton:
Thanks for the help, Bruno! I implemented the fix you suggested and tested it locally and it works as expected.

BrunoGrandePhD (approved):
🚀

PR #133, opened 2020-11-05
Crossmap addition to vcf2maf module

Kdreval:
I have updated the vcf2maf module to include conversion of maf files between genome builds. I made sure it is working on 2 test maf files. Once approved, I will move this to v1.2 and incorporate into demo workflow. Please do not merge with master if approved.

### DIFF
--- a/modules/vcf2maf/1.1/config/default.yaml
+++ b/modules/vcf2maf/1.1/config/default.yaml
@@ -7,6 +7,7 @@ lcr-modules:
             # Available wildcards: {out_dir}, {seq_type}, {genome_build}, {tumour_id}, {normal_id}, {pair_status}, {vcf_name}, {base_name}
             vep_cache: "__UPDATE__" #example "ref/ensembl_vep_cache/"
             sample_vcf_gz: "__UPDATE__" #full path to your compressed vcfs from favourite variant caller (with wildcards. You must encapsulate everything unique to the variant caller naming scheme in {base_name})
+            convert_coord: "{SCRIPTSDIR}/crossmap/1.0/convert_maf_coords.sh"
         vcf_base_name: "__UPDATE__" #put the consistent portion of the file name here. For Strelka, this is "combined.passed"
         options:
             vcf2maf: "--filter-vcf 0 --vcf-tumor-id TUMOR --vcf-normal-id NORMAL --cache-version 86"
@@ -14,9 +15,16 @@ lcr-modules:
             #--species        Ensembl-friendly name of species (e.g. mus_musculus for mouse) [homo_sapiens]
             #--cache-version  Version of offline cache to use with VEP (e.g. 75, 84, 91) [Default: Installed version]
             species: "homo_sapiens"
+        switches:
+            chains:
+                # specify path to the chain files used for crossmap conversion
+                grch37: "__UPDATE__"
+                grch37d5: "__UPDATE__"
+                hg38: "__UPDATE__"
 
         conda_envs:
             vcf2maf: "{MODSDIR}/envs/vcf2maf-1.6.18.yaml"
+            crossmap: "{SCRIPTSDIR}/crossmap/1.0/convert_maf_coords.yaml"
             
         threads:
             vcf2maf: 12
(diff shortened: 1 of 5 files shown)


modules/vcf2maf/1.1/config/default.yaml
```diff
@@ -7,16 +7,24 @@ lcr-modules:
...
+            convert_coord: "{SCRIPTSDIR}/crossmap/1.0/convert_maf_coords.sh"
         vcf_base_name: "__UPDATE__" #put the consistent portion of the file name here. For Strelka, this is "combined.passed"
         options:
             vcf2maf: "--filter-vcf 0 --vcf-tumor-id TUMOR --vcf-normal-id NORMAL --cache-version 86"
             # --filter-vcf     A VCF for FILTER tag common_variant. Set to 0 to disable [~/.vep/ExAC_nonTCGA.r0.3.1.sites.vep.vcf.gz]
             #--species        Ensembl-friendly name of species (e.g. mus_musculus for mouse) [homo_sapiens]
             #--cache-version  Version of offline cache to use with VEP (e.g. 75, 84, 91) [Default: Installed version]
             species: "homo_sapiens"
+        switches:
+            chains:
```
rdmorin: Why not show an example here too

#example "ref/blahblah/hg19Tohg38.chain"

modules/vcf2maf/1.1/vcf2maf.smk
```diff
@@ -91,18 +92,50 @@ rule _vcf2maf_run:
...
+        {params.chain}
+        {output.maf}
+        crossmap
+        > {log.stdout} 2> {log.stderr}
+        """)
+
+
 rule _vcf2maf_output_maf:
     input:
-        maf = str(rules._vcf2maf_run.output.maf)
```
rdmorin, line 96: Why did you get rid of str() on this?

rdmorin (approved):
I left some comments you can take or leave.

modules/vcf2maf/1.1/config/default.yaml
Kdreval: come to think of it more, it worked in my test because I already pointed to these chain files, which are already available because I downloaded them before. Instead, I should have incorporated the chains into `reference_files` workflow, so they will be automatically downloaded for the user. I will do that.

modules/vcf2maf/1.1/vcf2maf.smk
Kdreval: When I added another input for converted files, snakemake was behaving somehow strange on them if wrapped with `str()`. I removed it here to keep consistent with the other input for converted files. I will return the `str()` wrapping here back. Since outputs of this rule were already available when I did the test, I did not run this rule explicitly, so it might not behave as I am thinking it would be.

modules/vcf2maf/1.1/config/default.yaml
rdmorin: Good call

modules/vcf2maf/1.1/vcf2maf.smk
rdmorin: I can't remember why we started consistently adding str() but there was a good reason and it seemed to fix an issue I was having.

modules/vcf2maf/1.1/config/default.yaml
Kdreval: I added the handling of automatic download and usage of chains in the latest commit.

modules/vcf2maf/1.1/vcf2maf.smk
```diff
@@ -92,17 +92,23 @@ rule _vcf2maf_run:
...
+        return reference_files("genomes/{genome_build}/chains/hg19ToHg38.over.chain")
+
 rule _vcf2maf_crossmap:
     input:
         maf = rules._vcf2maf_run.output.maf,
-        convert_coord = CFG["inputs"]["convert_coord"]
+        convert_coord = CFG["inputs"]["convert_coord"],
+        chains = get_chain
     output:
         maf =  CFG["dirs"]["crossmap"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/{base_name}.converted.maf"
```
rdmorin, line 108: I am worried that .converted.maf is not very clear and, if anything, the path name implies the wrong genome build for the converted maf. I get that it's implied that the .converted.maf is the opposite genome build as the directory implies. Should we discuss a separate directory that states the new/converted build? Maybe we can discuss this in the GAMBL meeting

rdmorin: interested in your thoughts, Kostia and @lkhilton

rdmorin (approved):

PR #134, opened 2020-11-05
Module/controlfreec/1.2

Jwong684:
# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable

To be completed.


modules/controlfreec/CHANGELOG.md
```diff
@@ -38,4 +38,22 @@ This release was authored by Jasper.
...
+Improved control-freec calls for FFPE genomes. Everything is now default in paired mode with BAF turned on to measure LOH/BAF. All unmatched samples will be ran with a default normal. This will now generate temporary pileups in a scratch space (time consuming). The parameters have been set to minimize noise while also maximizing the number of prominent signals across the genome. To accomplish this, I set the default ploidy to 2 to prevent control-freec from overfitting the model for unpaired samples when ploidy != 2. Importantly, I set the parameter forceGCcontentNormalization=1, which forces GC normalzation for both the sample and control read counts (RC), and then calculates the ratio of sample RC/control RC. Then I changed the intercept of the polynomial from 1 to 0 to make the valuation of sample relative to the control instead of to its own GC content. This effectively smoothed out the profiles without added segments of overamplified "noise". I also allowed control-freec to assess the relative normal contamination and allowed control-freec to set its own window size given the coverage across the genome. These parameters effectively smoothed out unnecessary noise even further.
```
Kdreval: ` All unmatched samples will be ran with a default normal.`
How about matched samples? Are they running paired with patient's normal, or the same default normal as unpaired samples?

Kdreval: `I set the default ploidy to 2 to prevent control-freec from overfitting the model for unpaired samples when ploidy != 2`
But it still detects CNVs, right? What is the rough range of ploidy for detected CNVs?

Kdreval: `allowed control-freec to set its own window size given the coverage across the genome`
Seems like in the config file the window is set to 100000, or it is a different parameter?

demo/config.yaml
```diff
@@ -107,3 +108,6 @@ lcr-modules:
         inputs:
             sample_bam: "data/{sample_id}.bam"
             sample_bai: "data/{sample_id}.bam.bai"
+        scratch_subdirectories: ["mpileup"]
```
Kdreval: I wonder if this should be `scratch/mpileup`? Otherwise, I think it is creating the mpileup in running directory?

modules/controlfreec/1.1/config/default.yaml
```diff
@@ -3,12 +3,12 @@ lcr-modules:
...
 
         inputs:
-            # Available wildcards: {seq_type} {genome_build} {sample_id}
+            # Available wildcards: {seq_type} {genome_build} {tumour_id} {normal_id}
             sample_bam: "__UPDATE__"
             sample_bai: "__UPDATE__"
 
 
-        scratch_subdirectories: []
+        scratch_subdirectories: ["mpileup"]
```
Kdreval: same here, I think it should be `scratch/mpileup`

modules/controlfreec/1.1/config/default.yaml
```diff
@@ -17,41 +17,53 @@ lcr-modules:
         options:
             configFile: "{MODSDIR}/config/freec/config_WGS.txt"
             BedGraphOutput: FALSE
-            breakPointThreshold: 0.8 #use 0.6 for more segments to get more CNVs
+            breakPointThreshold: 0.6 #use 0.6 for more segments to get more CNVs
             breakPointType: 2
```
Kdreval: can you also please add comment here to show what 2 means, and if there are other options user can set?

modules/controlfreec/1.1/config/default.yaml
```diff
@@ -17,41 +17,53 @@ lcr-modules:
...
             minimalSubclonePresence: 20 #20 for WGS; 30 for WES
             maxThreads: 24
-            noisyData: FALSE #set TRUE for exomes to avoid false positives due to non-uniform capture
-            ploidy: 2,3,4 #will select the ploidy that explains the most CNAs
+            noisyData: TRUE #set TRUE for exomes to avoid false positives due to non-uniform capture
+            ploidy: 2 #will select the ploidy that explains the most CNAs
+            shiftInQuality: 0
             #sex: XX or XY
-            step: 10000 # only if window is specified; do not use for WES
+            step: 20000 # only if window is specified; do not use for WES
```
Kdreval: In case of WES, what this parameter should be? is it 0?

modules/controlfreec/1.1/config/default.yaml
```diff
@@ -17,41 +17,53 @@ lcr-modules:
...
             controlfreec_run: 24
             calc_sig: 1
             plot: 1
+            freec2bed: 1
 
         mem_mb:
-            controlfreec_run: 8000
+            pileup: 48000
+            cat: 8000
+            controlfreec_run: 24000
```
Kdreval: You run it on Numbers, right? So I assume this is enough for it to run and not be killed.

modules/controlfreec/1.1/config/default.yaml
```diff
@@ -17,41 +17,53 @@ lcr-modules:
...
+            freec2bed: 1000
         
         pairing_config:
             genome:
-                run_paired_tumours: False
-                run_unpaired_tumours_with: "no_normal"
-                run_paired_tumours_as_unpaired: True
+                run_paired_tumours: True
+                run_unpaired_tumours_with: "unmatched_normal"
+                run_paired_tumours_as_unpaired: False
```
Kdreval: I think if we want to be running every tumour as if it does not have a normal, this pairing config should be modified to `True` here

modules/controlfreec/1.1/config/default.yaml
```diff
@@ -17,41 +17,53 @@ lcr-modules:
...
-            controlfreec_run: 8000
+            pileup: 48000
+            cat: 8000
+            controlfreec_run: 24000
             calc_sig: 1000
             plot: 1000
+            freec2bed: 1000
         
         pairing_config:
             genome:
```
Kdreval: in theory this tool works for WES, so exomes should be configured here. In current pairing config I think it will not run on exomes. You can test it on the WES sample from demo, even in dry run.

modules/controlfreec/1.1/config/freec/config_WGS.txt
```diff
@@ -13,22 +13,44 @@ gemMappabilityFile = referenceFile
 outputDir = OUTDIR
 BedGraphOutput=FALSE
 breakPointThreshold = breakPointValue
+breakPointType = breakPointTypeBe
 coefficientOfVariation = coefVar
-contamination = numCon
+#contamination = numCon
```
Kdreval: if it's commented, does this mean it should be removed?

modules/controlfreec/1.1/config/freec/config_WGS.txt
```diff
@@ -13,22 +13,44 @@ gemMappabilityFile = referenceFile
...
 forceGCcontentNormalization = forceGCvalue
+intercept = interPoly
 minCNAlength = minCNAvalue
+minMappabilityPerWindow = minMapPerWindow
 minimalSubclonePresence = minimumSubclonePresenceValue
 noisyData = booNoise
+printNA = FALSE
 ploidy = ploidyInput
-step = stepValue
+#step = stepValue
```
Kdreval: same here

modules/controlfreec/1.1/config/freec/config_WGS.txt
```diff
@@ -13,22 +13,44 @@ gemMappabilityFile = referenceFile
...
 telocentromeric = teloValue
-window = windowSize
+uniqueMatch = uniqBoo
+#window = windowSize
```
Kdreval: same

modules/controlfreec/1.1/controlfreec.smk
```diff
@@ -34,20 +35,6 @@ localrules:
...
-        bai = CFG["inputs"]["sample_bai"]
-    output:
-        bam = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{sample_id}.bam",
-        bai = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{sample_id}.bai"
-    run:
-        op.relative_symlink(input.bam, output.bam)
-        op.relative_symlink(input.bai, output.bai)
-        
-
 # generate references
```
Kdreval: Does it need to be in reference_files?

modules/controlfreec/1.1/controlfreec.smk
```diff
@@ -114,44 +100,120 @@ rule _check_chrFiles:
...
+rule _controlfreec_mpileup_per_chrom:
+    input:
+        bam = CFG["dirs"]["inputs"] + "{seq_type}--{genome_build}/{sample_id}.bam",
+        fastaFile = reference_files("genomes/{genome_build}/genome_fasta/genome.fa"),
+        bed = "references/hg38/freec/dbsnp.common_all-151.bed"
+    output: # creates a temporary file for mpileup
+        pileup = temp(CFG["dirs"]["mpileup"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}/{sample_id}.{chrom}.minipileup.pileup.gz"),
+    conda:
+        CFG["conda_envs"]["controlfreec"]
+    threads: 1
```
Kdreval: if its the speed-limiting step, do you want to give it more threads here? and it also should be moved to config, so will be easily modifiable by user

modules/controlfreec/1.1/controlfreec.smk
```diff
@@ -114,44 +100,120 @@ rule _check_chrFiles:
...
+    resources:
```
Kdreval: please add bam=1 here so we can limit number of concurrent jobs running at the same time. Also, in recent modules (like lofreq, linx) we started using resources unpacking - it might be beneficial to use it here too

modules/controlfreec/1.1/controlfreec.smk
```diff
@@ -114,44 +100,120 @@ rule _check_chrFiles:
...
+        expand(
+            CFG["dirs"]["mpileup"] + "{{seq_type}}--{{genome_build}}/{{tumour_id}}--{{normal_id}}/{{sample_id}}.{chrom}.minipileup.pileup.gz", 
+            chrom = chroms
+        )
+    output: 
+        temp(CFG["dirs"]["mpileup"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}/{sample_id}.bam_minipileup.pileup.gz")
+    threads: 1
+    resources: 
+        mem_mb = CFG["mem_mb"]["cat"], 
+        pileup = 1
```
Kdreval: what does resources.pileup does? doesn't seem to be used in shell command

modules/controlfreec/1.1/controlfreec.smk
```diff
@@ -160,89 +222,128 @@ rule _controlfreec_config:
...
+        CNV = str(rules._controlfreec_calc_sig.output.txt),
+        bed = str(rules._controlfreec_freec2bed.output.bed),
+        BAF = str(rules._controlfreec_run.output.BAF),
+        BAFgraph = str(rules._controlfreec_plot.output.bafplot),
+        ratio = str(rules._controlfreec_run.output.ratio)
     output:
-        plot = CFG["dirs"]["outputs"] + "{seq_type}--{genome_build}/plots/{sample_id}.bam_ratio.txt.png",
-        log2plot = CFG["dirs"]["outputs"] + "{seq_type}--{genome_build}/log2plots/{sample_id}.bam_ratio.txt.log2.png",
-        CNV = CFG["dirs"]["outputs"] + "{seq_type}--{genome_build}/CNV/{sample_id}.bam_CNVs.p.value.txt"
+        plot = CFG["dirs"]["outputs"] + "{seq_type}--{genome_build}/plots/{tumour_id}--{normal_id}.bam_ratio.txt.png",
```
Kdreval: I think bam in all these symlink names can be dropped

demo/config.yaml
lkhilton: The op.setup_module code makes a symlink in the control freec results directory and puts the real directory in the scratch path that's specified by the user, so it doesn't need to be `scratch/mpileup`. 

However I do recommend leaving this blank and placing a comment there recommending to users which directories should be scratch. Otherwise you get unexpected scratch directory usage and it creates many, many issues with existing symlink errors if the user needs to re-run.

modules/controlfreec/1.1/config/default.yaml
lkhilton: This should only be set as a scratch directory if the demo is setup to use scratch... is it?

modules/controlfreec/1.1/config/default.yaml
lkhilton: Agree with Kostia, there are a few other parameters that could use a comment so users know what the parameter does and if/when it should be changed.

modules/controlfreec/1.1/config/default.yaml
```diff
@@ -17,41 +17,53 @@ lcr-modules:
...
+            contaminationAdjustment: TRUE
+            degree: '3\&4' #3&4 for WGS; 1 for WES
+            forceGCcontentNormalization: 1 #0 for WGS; 1 for WES # 0 forces control-base normalization, 1 forces GC
+            intercept: 0 # 0 for control-based (paired) ; 1 for GC-content (unpaired)
+            minCNAlength: 8 #1 for WGS; 3 for WES
+            minimalCoveragePerPosition: 10 # default is 0
+            minMappabilityPerWindow: 0.9
+            minQualityPerPosition: 20
             minimalSubclonePresence: 20 #20 for WGS; 30 for WES
             maxThreads: 24
```
lkhilton: This should be set by the {threads} parameter.

modules/controlfreec/1.1/config/default.yaml
lkhilton: Please change to use resource unpacking. Restructure as follows: 
```
resources: 
    controlfreec_run: 
        mem_mb: 8000
    pileup: 
        mem_mb: 48000
    ...
```

Then, this gets called in the Snakefile as follows: 
```
resources: 
  **CFG['resources']['controlfreec_run']
```

This method gives users flexibility to add additional resources, e.g. `pileup: 1`, to prevent too much memory hogging.

modules/controlfreec/1.1/config/default.yaml
```diff
@@ -17,41 +17,53 @@ lcr-modules:
         options:
             configFile: "{MODSDIR}/config/freec/config_WGS.txt"
```
lkhilton: Given how many of these parameters are different for WES vs. genomes, maybe some of them need to use `switch_on_wildcard()`?

modules/controlfreec/1.1/config/freec/config_WGS.txt
lkhilton: I'm not sure why any of the values are commented, won't that prevent them from being ever used?

modules/controlfreec/1.1/config/freec/config_WGS.txt
```diff
@@ -13,22 +13,44 @@ gemMappabilityFile = referenceFile
 outputDir = OUTDIR
```
lkhilton: Should there be a separate `config_WES.txt` file for exomes?

modules/controlfreec/1.1/controlfreec.smk
lkhilton: Yes please, also I don't think that you can assume that the user has `references/{genome_build}` configured as the location for their reference files. For modules I've written where I need to get specific, small reference files, I put them in the module `00-inputs` directory. That can be a stopgap until the rest is incorporated in the reference workflow.

modules/controlfreec/1.1/controlfreec.smk
```diff
@@ -78,7 +65,6 @@ chromRegions = ("1", "2", "3", "4", "5", "6", "7", "8",
         "17", "18", "19", "20", "21", "22", "X", "Y")
```
lkhilton: I'm not sure how these are getting "chr" added for hg19 or hg38. 

There's a file in the reference file workflow `genomes/{genome_build}/genome_fasta/main_chromosomes.txt`. You could check out this [checkpoint rule](https://github.com/LCR-BCCRC/lcr-modules/blob/5569794beb005874a67e767fb970f9b8eda3b5ab/modules/mutect2/2.0/mutect2.smk#L51) and [this function](https://github.com/LCR-BCCRC/lcr-modules/blob/5569794beb005874a67e767fb970f9b8eda3b5ab/modules/mutect2/2.0/mutect2.smk#L153) from the mutect module for how to use that for parallelizing by chromosome.

modules/controlfreec/CHANGELOG.md
Jwong684: Patient's normals. I set the config to be:

`        pairing_config:
            genome:
                run_paired_tumours: True
                run_unpaired_tumours_with: "unmatched_normal"
                run_paired_tumours_as_unpaired: False`

modules/controlfreec/1.1/controlfreec.smk
```diff
@@ -114,44 +100,120 @@ rule _check_chrFiles:
         touch("references/{genome_build}/freec/chr/.all_done")
 
 
+rule _dbsnp_to_bed:
```
lkhilton: Can you please set all rules to adhere to the convention `_controlfreec_..`

modules/controlfreec/1.1/controlfreec.smk
lkhilton: Good point. Jasper, we've been trying to standardize our output file names. So here, for example, this could be `"{seq_type}--{genome_build}/plots/{tumour_id}--{normal_id}--{pair_status}.controlfree.ratio.png"`

modules/controlfreec/1.1/controlfreec.smk
```diff
@@ -160,89 +222,128 @@ rule _controlfreec_config:
...
+        "sed \"s|minCovPerPos|{params.minimalCoveragePerPosition}|g\" | "
         "sed \"s|booNoise|{params.noisyData}|g\" | "
         "sed \"s|stepValue|{params.step}|g\" | "
         "sed \"s|teloValue|{params.telocentromeric}|g\" | "
+        "sed \"s|uniqBoo|{params.uniqBoo}|g\" | "
         "sed \"s|numThreads|{params.threads}|g\" | "
         "sed \"s|referenceFile|{input.reference}|g\" > {output}"
 
 
 rule _controlfreec_run:
```
lkhilton: I'm unsure how this is going to work in unpaired mode, e.g. without a normal. I think you might need a separate rule.

modules/controlfreec/1.1/controlfreec.smk
```diff
@@ -114,44 +100,120 @@ rule _check_chrFiles:
...
+    resources: 
+        mem_mb = CFG["mem_mb"]["cat"], 
+        pileup = 1
+    group: "controlfreec"
+    shell: 
+        "cat {input} > {output} "
+
+
 # set-up controlfreec
 rule _controlfreec_config:
```
lkhilton: Beginning from this rule could you please add `pair_status` as a wildcard? It should always be `{tumour_id}--{normal_id}--{pair_status}`.

modules/controlfreec/CHANGELOG.md
Jwong684: Yes, it detects all CNVs relative to ploidy = 2. Before, when I set the range to 2,3,4, it seems like there would be some over-fitting done, where certain samples would be assigned something like ploidy = 3 and then show every chromosome as a CN loss. It felt like, especially with unpaired samples, that setting a default to 2 would give the most meaningful results

Jwong684: In the config file I set it to 100000, but I just commented out the actual windows option inside the config_WGS.txt file, so it wouldn't change anything

lkhilton (changes requested):
Like Kostia I'm requesting changes that will make this more generalizable and enable it to be run with different pairing configs.

demo/config.yaml
Jwong684: In my test demo run, it was generating the scratch folder inside the scratch space, so it seemed like it was working

modules/controlfreec/1.1/config/default.yaml
Jwong684: Okay!

modules/controlfreec/1.1/config/default.yaml
Jwong684: Yes, for WES it is 0 and window will also be 0 since it uses the exons as dynamic windows

modules/controlfreec/1.1/config/default.yaml
Jwong684: Laura: Okay that sounds good Laura.

Kostia: I had some issues before with memory dumping, so I gave it an upper range just to be safe. I can change it back to 8000

Kdreval: It's opposite, I was thinking it looks a bit low. Please leave it as is if that's what you tested.

modules/controlfreec/1.1/config/default.yaml
Jwong684: Yes, my current demo repo looks like this:

scratch -> /projects/rmorin_scratch/jawong_tmp/

results/controlfreec-1.1
├── 00-inputs
│   └── genome--hg38
├── 01-mpileup -> ../../scratch/controlfreec-1.1/01-mpileup
├── 02-run
│   └── genome--hg38
├── 99-outputs
│   └── genome--hg38
└── logs
    ├── launched-2020-11-05-at-00-06-02
    ├── launched-2020-11-05-at-09-59-48
    ├── launched-2020-11-05-at-10-00-11
    ├── launched-2020-11-05-at-10-38-58
    ├── launched-2020-11-05-at-10-39-52
    ├── launched-2020-11-05-at-10-40-00
    ├── launched-2020-11-05-at-10-43-04
    ├── launched-2020-11-05-at-10-46-34
    ├── launched-2020-11-05-at-10-47-43
    ├── launched-2020-11-05-at-10-51-01
    └── launched-2020-11-05-at-10-51-39

Jwong684: The scratch 01-mpileup symlink was automatically generated when doing my test run

modules/controlfreec/1.1/config/default.yaml
Jwong684: Yes, it should work on WES, but it has to be in paired mode. I have not tried testing it extensively yet though. The current configuration requires a separate config_WGS.txt file (to include a "target_regions" section for the exons/target sites).

modules/controlfreec/1.1/config/freec/config_WGS.txt
Jwong684: Yes those values are commented out to give way for the other parameters with lower priority. In this case, coefficientOfVariation would dictate window size, and step size relies on window size. If either of those have values, then coefficientOfVariation is overruled, so both windowSize and Step are commented out.

In the case of contamination, if contamination contains a value, it will override contaminationAdjustment's auto-correction. Unless we know the contamination by normal in a given sample, I think letting control-freec decide is the best choice

modules/controlfreec/1.1/controlfreec.smk
Jwong684: Hm I see, so just move these references as a subdirectory under 00-input?

ex. CFG["dirs"]["inputs"] + "references/{genome_build}/[...]"

modules/controlfreec/1.1/controlfreec.smk
Jwong684: The concatenation part should be fairly fast - it just cat's all the separate chrom.gz files into one. I'm not sure if supplying more threads will speed it up.

I will add it to the config though

modules/controlfreec/1.1/controlfreec.smk
lkhilton: Yep!

modules/controlfreec/1.1/controlfreec.smk
Jwong684: This was supposed to just limit the number of mpileups that is generated per iteration so it doesn't overfill the scratch space with mpileups

modules/controlfreec/1.1/controlfreec.smk
lkhilton: I actually don't think this particular call to `threads` needs to be configurable because `samtools mpileup` is single-threaded, isn't it?

modules/controlfreec/1.1/controlfreec.smk
lkhilton: This is comparable to `bam = 1`

modules/controlfreec/CHANGELOG.md
lkhilton: Ideally, since control freec is able to run without a normal, the module would be user-configurable to run without one.

modules/controlfreec/1.1/controlfreec.smk
Jwong684: Okay I can drop the ".bam" - I haven't integrated pair_status as a wildcard here yet, but I guess it would be a good idea to include it

modules/controlfreec/1.1/config/freec/config_WGS.txt
Jwong684: Yes there will be. I need to think about the best way to implement exomes, since the default parameters and config file would be separate. I think switch_on_wildcards should be the easiest way to go at it. I'll likely get to the WES part in a future update since I need to test default WES parameters first (probably on the LLMPP exome data)

modules/controlfreec/1.1/controlfreec.smk
Jwong684: The chromRegions variable is only used in the final _check_chrFiles rule, which generates a touch(.output.done) file to make sure all individual {chrom}.fa files have been accounted for. It has no bearing on the contents of the files.

My _generate_chrFasta rule is just my quick and dirty way to grep each chromosome (with or without "chr") depending on the genome_build

modules/controlfreec/1.1/controlfreec.smk
Jwong684: Okay!

modules/controlfreec/1.1/controlfreec.smk
Jwong684: Yeah I think the simplest method would be to have a bunch of different config_WGS.txt files for different contexts (ex. paired, unpaired, WES, WGS). And a separate rule for unpaired vs paired to account for the control.

Jwong684: At the moment, I made control-freec run in paired mode always and to match unpaired tumours since it gives the cleanest results.

modules/controlfreec/1.1/controlfreec.smk
Jwong684: Okay!

modules/controlfreec/CHANGELOG.md
rdmorin: Agree. Either we should add a comment to the default config file to explain what options should be used with an unmatched normal or the module should be able to set the appropriate combination of parameters depending on whether its being run in inpaired, unmatched or matched mode.

rdmorin: Again, it would be good to leave a short comment in the default config that indicates a range can be provided

rdmorin: That's going to cause a lot of confusion to the users, don't you think? Why is this even in the config if it's not being used?

demo/config.yaml
rdmorin: Even if this "works", as a general rule, I think we are leaving this empty (i.e. []) in default configs along with a comment to explain which rules might need scratch space. i.e. what Laura said.

modules/controlfreec/1.1/config/freec/config_WGS.txt
rdmorin: Can you document that briefly in the config? ie. a line below that says something to the effect of:
#uncomment the above line if you want to specify tumour purity, otherwise allow ControlFREE to determine this automatically (recommended)

modules/controlfreec/1.1/controlfreec.smk
rdmorin: Laura is correct.

modules/controlfreec/1.1/controlfreec.smk
rdmorin: We probably want to change to bam=1, right? Just for consistency across tools?

modules/controlfreec/1.1/controlfreec.smk
rdmorin: It's definitely preferable to use the existing chromosome name files rather than hard-coding anything

modules/controlfreec/1.1/config/freec/config_WGS.txt
rdmorin: Let's just document that this version is not recommended for exomes and make sure capture data isn't in the default config.

modules/controlfreec/1.1/controlfreec.smk
lkhilton: I like having it user-configurable. That way if you're running two or more modules in the same Snakefile, you can give each module a different allocation of resources.

rdmorin: Not sure I understand what you mean by user-configurable. You mean having that set by the config?

lkhilton: Yes, it's better to use resource unpacking so the user can decide what keyword to use to limit pileup generation.

modules/controlfreec/CHANGELOG.md
Jwong684: I left it in the config so that if people want to modify the window/step/contamination values, they could. The actual config_WGS.txt file that control-freec uses already has those options commented out. 

I just commented these options out for now and left a comment to tell users to remove the comment in config_WGS.txt if they wish to use these options.

modules/controlfreec/1.1/controlfreec.smk
```diff
@@ -52,59 +52,72 @@ rule _get_map_refs:
...
+        chrs = CFG["dirs"]["inputs"] + "references/{genome_build}/main_chromosomes_withY.txt"
+    run:
+        op.relative_symlink(input.chrs, output.chrs)
+
+
+def _controlfreec_get_chr_fastas(wildcards):
+    CFG = config["lcr-modules"]["controlfreec"]
+    chrs = checkpoints._controlfreec_input_chrs.get(**wildcards).output.chrs
+    with open(chrs) as file:
+        chrs = file.read().rstrip("\n").split("\n")
```
rdmorin: You really should not clobber your chrs variable here. You're creating a new variable (chromosome) later. You should create it here instead.

Jwong684: Updated to chromosome

Jwong684:
@Kdreval @lkhilton I've updated the module. Also included unmatched/matched rule separation for running controlfreec.

modules/controlfreec/1.1/controlfreec.smk
```diff
@@ -160,89 +250,157 @@ rule _controlfreec_config:
...
-        log2plot = CFG["dirs"]["outputs"] + "{seq_type}--{genome_build}/log2plots/{sample_id}.bam_ratio.txt.log2.png",
-        CNV = CFG["dirs"]["outputs"] + "{seq_type}--{genome_build}/CNV/{sample_id}.bam_CNVs.p.value.txt"
+        plot = CFG["dirs"]["outputs"] + "{seq_type}--{genome_build}/plots/{tumour_id}--{normal_id}--{pair_status}.ratio.txt.png",
+        log2plot = CFG["dirs"]["outputs"] + "{seq_type}--{genome_build}/log2plots/{tumour_id}--{normal_id}--{pair_status}.ratio.txt.log2.png",
+        CNV = CFG["dirs"]["outputs"] + "{seq_type}--{genome_build}/CNV/{tumour_id}--{normal_id}--{pair_status}.CNVs.p.value.txt",
+        bed = CFG["dirs"]["outputs"] + "{seq_type}--{genome_build}/bed/{tumour_id}--{normal_id}--{pair_status}.CNVs.bed",
+        BAF = CFG["dirs"]["outputs"] + "{seq_type}--{genome_build}/BAF/{tumour_id}--{normal_id}--{pair_status}.BAF.txt",
+        BAFgraph = CFG["dirs"]["outputs"] + "{seq_type}--{genome_build}/BAFplot/{tumour_id}--{normal_id}--{pair_status}.BAF.txt.png",
+        ratio = CFG["dirs"]["outputs"] + "{seq_type}--{genome_build}/ratio/{tumour_id}--{normal_id}--{pair_status}.ratio.txt"
```
Kdreval: My only suggestion would be to further clean up the names of output files by dropping some intermediate extensions, like this:
```
        plot = CFG["dirs"]["outputs"] + "{seq_type}--{genome_build}/plots/{tumour_id}--{normal_id}--{pair_status}.ratio.png",
        log2plot = CFG["dirs"]["outputs"] + "{seq_type}--{genome_build}/log2plots/{tumour_id}--{normal_id}--{pair_status}.log2_ratio.png",
        CNV = CFG["dirs"]["outputs"] + "{seq_type}--{genome_build}/CNV/{tumour_id}--{normal_id}--{pair_status}.CNVs.txt",
        bed = CFG["dirs"]["outputs"] + "{seq_type}--{genome_build}/bed/{tumour_id}--{normal_id}--{pair_status}.CNVs.bed",
        BAF = CFG["dirs"]["outputs"] + "{seq_type}--{genome_build}/BAF/{tumour_id}--{normal_id}--{pair_status}.BAF.txt",
        BAFgraph = CFG["dirs"]["outputs"] + "{seq_type}--{genome_build}/BAFplot/{tumour_id}--{normal_id}--{pair_status}.BAF.png",
        ratio = CFG["dirs"]["outputs"] + "{seq_type}--{genome_build}/ratio/{tumour_id}--{normal_id}--{pair_status}.ratio.txt"
```

demo/config.yaml
```diff
@@ -108,6 +108,6 @@ lcr-modules:
         inputs:
             sample_bam: "data/{sample_id}.bam"
             sample_bai: "data/{sample_id}.bam.bai"
-        scratch_subdirectories: ["mpileup"]
-        unmatched_normal_ids:
-            genome--hg38: "BLGSP-71-23-00361-14A-01E"
+        scratch_subdirectories: [] # mpileup should be in scratch space
+        unmatched_normal_ids: 
+            genome--hg38: "BLGSP-71-23-00361-14A-01E" ## FFPE normal
```
lkhilton: Can you delete these two lines please? Usually the unmatched normal ID is set in the shared lcr-modules config, and users can set module-specific unmatched_normal_ids in their project config, too. We don't want this in the module since users at other sites won't have this normal.

modules/controlfreec/1.1/controlfreec.smk
```diff
@@ -33,13 +37,17 @@ localrules:
     _controlfreec_all
 
 
+wildcard_constraints:
```
lkhilton: Is there a particular rule/rules that need this wildcard constraint? I've had TONS of issues setting global constraints in my Snakefiles and I've reverted to using them only on a per-rule basis because of it. If you can set this just in one or two rules that need it it might save a ton of headaches down the road.

modules/controlfreec/1.1/controlfreec.smk
```diff
@@ -233,86 +262,114 @@ rule _controlfreec_config:
...
-        BAF = CFG["dirs"]["run"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}/{tumour_id}.bam_minipileup.pileup.gz_BAF.txt"
+        info = CFG["dirs"]["run"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/{tumour_id}.bam_minipileup.pileup.gz_info.txt",
+        ratio = CFG["dirs"]["run"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/{tumour_id}.bam_minipileup.pileup.gz_ratio.txt",
+        CNV = CFG["dirs"]["run"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/{tumour_id}.bam_minipileup.pileup.gz_CNVs",
+        BAF = CFG["dirs"]["run"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/{tumour_id}.bam_minipileup.pileup.gz_BAF.txt"
     conda: CFG["conda_envs"]["controlfreec"]
     threads: CFG["threads"]["controlfreec_run"]
-    resources: mem_mb = CFG["mem_mb"]["controlfreec_run"]
+    resources: **CFG["resources"]["controlfreec_run"]
+    wildcard_constraints:
```
lkhilton: Is there anything different about the matched vs. unmatched rules? I don't think you actually need two rules, the output rule with the runs table should take care of matching the right matched vs. unmatched tumour/normal pairs.

lkhilton (commented):
@Jwong684 there are a few outstanding questions/comments. Almost there!

demo/config.yaml
Jwong684: Will do!

modules/controlfreec/1.1/controlfreec.smk
Jwong684: I based this off your gridss snakemake where you had it as a global constraint. I didn't have any issues with this wildcard since my test set only had unmatched and matched. I'll remove it for now since it doesn't seem like it has any impact at all.

modules/controlfreec/1.1/controlfreec.smk
Jwong684: I was thinking it may be an issue if I had multiple runs that tries to read from the same unmatched normal at the same time, so I made a separate rule specifically for the unmatched group in case that ever needs to be configured, like set bam=1.

modules/controlfreec/1.1/controlfreec.smk
lkhilton: Yes, I've only recently learned of the problems that these global constraints cause. Thanks for removing.

modules/controlfreec/1.1/controlfreec.smk
lkhilton: I'm not too worried about it - control FREEC doesn't take long to run if the pileups are pre-generated, and you can just set a resource limit that prevents it from launching too many matched or unmatched jobs simultaneously. I think you would only need a separate unmatched normal rule if you were creating group jobs. Please do delete the redundant rule and the wildcard constraints.

Jwong684: Okay, the redundant rule and wildcard constraint are removed.

lkhilton (approved):

Jwong684:
Ah shoot - I just made a tiny edit to my snakemake and it cancelled the merge approval. Could someone approve it so I could merge my updated version into master?

lkhilton (dismissed):

lkhilton (approved):

PR #136, opened 2020-11-10
Fix and improve crossmap in vcf2maf module

Kdreval:
With addition of crossmap to `vcf2maf` module, my latest commit introduced a bug to the `reference_files` workflow in the rule that downloaded the liftOver chains from UCSC. This caused issue preventing snakemake to run. This PR fixes the bug by including required wildcards `version` and `params.provider` into the rules I added. 
In addition, this PR improves naming of output symlinks from `vcf2maf` module by specifying in file name the chain used for conversion of genomic coordinates. 
This implementation has been tested and run on all current strelka outputs.

### DIFF
--- a/modules/vcf2maf/1.1/vcf2maf.smk
+++ b/modules/vcf2maf/1.1/vcf2maf.smk
@@ -117,7 +117,7 @@ rule _vcf2maf_crossmap:
         mem_mb = CFG["mem_mb"]["vcf2maf"]
     params:
         out_name = CFG["dirs"]["crossmap"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/{base_name}.converted_",
-        chain = lambda w: "hg38ToHg19" if "38" in {w.genome_build} else "hg19ToHg38",
+        chain = lambda w: "hg38ToHg19" if "38" in str({w.genome_build}) else "hg19ToHg38",
         file = ".maf"
     shell:
         op.as_one_line("""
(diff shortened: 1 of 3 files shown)


modules/vcf2maf/1.1/vcf2maf.smk
```diff
@@ -95,53 +95,58 @@ rule _vcf2maf_run:
...
+        return reference_files("genomes/{genome_build}/chains/grch37/hg19ToHg38.over.chain")
 
 rule _vcf2maf_crossmap:
     input:
         maf = rules._vcf2maf_run.output.maf,
         convert_coord = CFG["inputs"]["convert_coord"],
         chains = get_chain
     output:
-        maf =  CFG["dirs"]["crossmap"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/{base_name}.converted.maf"
+        maf =  CFG["dirs"]["crossmap"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/{base_name}.converted"
```
lkhilton: This is just an empty target file, right? Could you please re-name it e.g. `converted = CFG...`? I think it would help avoid confusion on debugging.

lkhilton (commented):
This is a nice addition to the vcf2maf module. One teensy request before I will approve.

modules/vcf2maf/1.1/vcf2maf.smk
rdmorin: That empty file was my suggestion as a sentinel to get around the unpredictable naming of the real output. There's a long Slack thread on this from last night if you want to scroll through the last bit of it for context.

lkhilton: The pipeline makes perfect sense, I'm just suggesting it shouldn't be called `maf = ...` since it's not the actual maf file.

Kdreval: Yes, this is an empty file. I rename it to `dispatch` so that it will be more indicative of what the file really is.

rdmorin: Oh, I see. You just meant the variable name. Sorry.

Kdreval: I renamed it to be `dispatch` and it is ready to be merged if approved.

lkhilton (approved):

PR #139, opened 2020-11-16
make lofreq compatible with vcf2maf

Kdreval:
The Lofreq outputs vcf file in the format that is not compatible with vcf2maf to populate `t_depth`, `t_ref_counts`, `t_alt_counts`. The reason is that `DP` and `DP4` columns need to be in the `FORMAT` field instead of `INFO`. I have followed suggestions from Lofreq developers in [this](https://github.com/CSB5/lofreq/issues/57) issue and vcf2maf developers in [this](https://github.com/mskcc/vcf2maf/issues/167) issue to modify the Lofreq outputs. By implementing changes to the filtering script, I made sure vcf files will be compatible with vcf2maf.

### DIFF
--- a/modules/lofreq/1.0/src/bash/lofreq_filter.sh
+++ b/modules/lofreq/1.0/src/bash/lofreq_filter.sh
@@ -28,9 +28,9 @@ zcat "${INPUT_FILE}" \
                   s/$&/$&\tFORMAT\tTUMOR/;
                 };
                 print;
-              };
+              }
               # For each feature, add FORMAT column with descriptors and populate TUMOUR column with depth, reads counts
-              if(/^(chr)*[\dX]+\s.+/){
+              else{
                 my @data = map { chomp; [ split /=|;/ ] } $_;
                 $NEW_ROW = "$_\tDP:DP4\t$data[0][1]:$data[0][7]\n";
                 print $NEW_ROW;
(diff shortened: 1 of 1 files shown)


lkhilton:
I am not qualified to review something so Perl-y! @rdmorin?

rdmorin:
I'm not even sure I am. I hope we can simplify this. I'm seeing some code that could be reduced and simplified. @Kdreval can you send me an example of the Lofreq header before and after this manipulation is applied?

Kdreval:
I have updated the language in the latest commit, please see the updated version

modules/lofreq/1.0/src/bash/lofreq_filter.sh
```diff
@@ -8,6 +8,20 @@
...
+  | perl -ne '
+              # Add 2 new rows to the description and 2 new columns in the header
+              if(/^#/){
+                if(/##INFO=<ID=DP,.+\n/){$DP=$&;}; $DP =~ s/INFO/FORMAT/; print $DP if /min_dp/;
+                if(/##INFO=<ID=DP4,.+\n/){$DP4=$&;}; $DP4 =~ s/INFO/FORMAT/; print $DP4 if /min_dp/;
+                if(/^#CHROM.+/){s/$&/$&\tFORMAT\tTUMOR/;};
+                print;
+                };
+              # For each feature, add FORMAT column with descriptors and populate TUMOUR column with depth, reads counts
+              if(/^(chr)*[\dX]+\s.+/){
```
rdmorin: I wonder if this could just forego the regex and just be:
else{
...
}

rdmorin (changes requested):
I made some edits. Can you test out this version?

rdmorin (approved):

PR #140, opened 2020-11-18
Starfish

lkhilton:
# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable


rdmorin (commented):
Is there any way to change this to show the differences in the files vs the previous version? Or is this basically a complete rewrite of my crappy original version?

lkhilton:
I started from scratch 😬

lkhilton:
Forgot to mention I dropped the bed file conversion part because it was unnecessary for mutect inputs. I can add something to a vcf_utils module for gnomAD filtering and vcf to bed conversions.

modules/starfish/2.0/config/default.yaml
```diff
@@ -0,0 +1,51 @@
...
+
+        dirs: 
+            
+            _parent: "__UPDATE__"
+
+        inputs:
+            # Available wildcards: {seq_type} {genome_build} {tumour_id} {normal_id} {pair_status}
+            # Each key:value pair is the name of the variant caller and the path to its VCF file
+            # Each key should be unique in this set. 
+            vcf:
```
rdmorin, line 12: Doesn't this tool need at least two variant callers to work? I guess it doesn't matter if they'll get clobbered by the user config but should we put at least two here, e.g. Strelka and Lofreq?

rdmorin: In fact, why not list all the ones we currently have available and tell the user to delete/comment out any they are not using?

modules/starfish/2.0/starfish2.smk
```diff
@@ -0,0 +1,247 @@
...
+
+# All variant callers from config
+callers = list(CFG["inputs"]["vcf"].keys())
+callers = [caller.lower() for caller in callers]
+
+# The name of the union VCF file
+union_vcf = copy.deepcopy(callers)
+union_vcf.append("union")
+union_vcf = "_".join(union_vcf)
+not_union_vcf = "!" + union_vcf
```
rdmorin: wut?

workflows/reference_files/2.4/reference_files_header.smk
```diff
@@ -315,6 +315,23 @@ rule download_liftover_chains:
         gzip -dc > {output.chains}
         """)
 
+rule download_sdf:
```
rdmorin, line 333: good find getting them online. I thought we had to make our own sdf references!

modules/starfish/2.0/config/default.yaml
```diff
@@ -0,0 +1,51 @@
+lcr-modules:
+    
+    starfish:
+
+        dirs: 
+            
+            _parent: "__UPDATE__"
```
rdmorin: Is this documented? Do you want to provide a best practices example? I might not be asking this if this was adopted in the demo.

modules/starfish/2.0/starfish2.smk
```diff
@@ -0,0 +1,247 @@
...
+        stdout = CFG["logs"]["starfish"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/starfish_run.stdout.log",
+        stderr = CFG["logs"]["starfish"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/starfish_run.stderr.log"
+    params:
+        opts = CFG["options"]["starfish_run"]
+    conda:
+        CFG["conda_envs"]["starfish"]
+    threads:
+        CFG["threads"]["starfish_run"]
+    resources:
+        **CFG["resources"]["starfish_run"]
```
rdmorin: Is how this works documented anywhere?

modules/starfish/2.0/starfish2.smk
```diff
@@ -0,0 +1,247 @@
...
+        complete = CFG["dirs"]["starfish"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/starfish.complete"
+    output: 
+        renamed = touch(CFG["dirs"]["starfish"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/starfish.renamed")
+    run: 
+        # Get the starfish letter names for the variant callers
+        names = list(string.ascii_uppercase[0:(len(callers))])
+        # Generate a dictionary of letter names and variant callers
+        name_dict = {names[i]: callers[i] for i in range(len(callers))}
+        outdir = os.path.dirname(input.complete)
+        for src in glob.glob(outdir + "/[A-Z]*.vcf.gz*"):
```
rdmorin: Nice way to handle this. Starfish should do this itself since it goes to all the effort of asking for the tool names.

modules/starfish/2.0/starfish2.smk
```diff
@@ -0,0 +1,247 @@
...
+        name_dict = {names[i]: callers[i] for i in range(len(callers))}
+        outdir = os.path.dirname(input.complete)
+        for src in glob.glob(outdir + "/[A-Z]*.vcf.gz*"): 
+            bname = os.path.basename(src)
+            lhs, rhs = bname.split(".", 1)
+            lhs = "_".join(lhs)
+            for key in list(name_dict.keys()): 
+                lhs = lhs.replace(key, name_dict[key])
+            dest = os.path.join(outdir, ".".join([lhs, rhs]))
+            os.rename(src, dest)
```
rdmorin: If you are still getting that tabix index error. This might be a good place to just "touch" all the .tbi files to make sure they are newer than their VCF

modules/starfish/2.0/starfish2.smk
```diff
@@ -0,0 +1,247 @@
...
+            for key in list(name_dict.keys()): 
+                lhs = lhs.replace(key, name_dict[key])
+            dest = os.path.join(outdir, ".".join([lhs, rhs]))
+            os.rename(src, dest) 
+
+
+def _starfish_get_output(wildcards): 
+    # Get the checkpoint output directory
+    CFG = config["lcr-modules"]["starfish"]
+    checkpoint_output = os.path.dirname(checkpoints._starfish_rename_output.get(**wildcards).output.renamed)
```
rdmorin: Can you please document what this line does because all documentation on this is so vague I have no clue. Maybe this can be the one place in the world that has it eloquently described.

rdmorin: For example, why **wildcards? Why get? What is any of this doing?

modules/starfish/2.0/starfish2.smk
```diff
@@ -0,0 +1,247 @@
...
+    # Rename each VCF with the variant caller names
+    vcfs = expand(
+        CFG['dirs']['starfish'] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/{vcf_name}.vcf.gz",
+        seq_type = wildcards.seq_type,
+        genome_build = wildcards.genome_build, 
+        tumour_id = wildcards.tumour_id, 
+        normal_id = wildcards.normal_id, 
+        pair_status = wildcards.pair_status, 
+        vcf_name = glob_wildcards(os.path.join(checkpoint_output, "{vcf_name}.vcf.gz")).vcf_name
```
rdmorin: I also don't understand what this does

modules/starfish/2.0/starfish2.smk
```diff
@@ -0,0 +1,247 @@
+#!/usr/bin/env snakemake
```
rdmorin: Why is this snakefile named starfish2.smk ?

modules/starfish/2.0/config/default.yaml
```diff
@@ -0,0 +1,51 @@
...
+
+        scratch_subdirectories: []
+
+        options:
+            starfish_run: "--sample ALT --squash-ploidy --vennout-format pdf"
+            starfish_union: "-m both"
+
+        conda_envs:
+            bcftools: "{MODSDIR}/envs/bcftools-1.10.2.yaml"
+            starfish: "{MODSDIR}/envs/starfish.yaml"
```
rdmorin, line 31: I'm not complaining, merely curious. What is our policy on when version numbers are and are not needed in these?

modules/starfish/2.0/starfish2.smk
```diff
@@ -0,0 +1,247 @@
...
+    _starfish_all,
+
+##### GLOBAL VARIABLES #####
+
+# All variant callers from config
+callers = list(CFG["inputs"]["vcf"].keys())
+callers = [caller.lower() for caller in callers]
+
+# The name of the union VCF file
+union_vcf = copy.deepcopy(callers)
```
rdmorin: Wouldn't it be easier to just call .keys twice? If you're worried about getting them in the same order I doubt that's a problem. You don't have to change it. Just thought I would point this out.

modules/starfish/2.0/starfish2.smk
```diff
@@ -0,0 +1,247 @@
...
+
+# Symlinks the final output files into the module results directory (under '99-outputs/')
+rule _starfish_output_vcf:
+    input:
+        vcf = CFG['dirs']['starfish'] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/{vcf_name}.vcf.gz"
+    output:
+        vcf = CFG["dirs"]["outputs"] + "vcf/{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}.{vcf_name}.vcf.gz", 
+        tbi = CFG["dirs"]["outputs"] + "vcf/{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}.{vcf_name}.vcf.gz.tbi"
+    wildcard_constraints: 
+        vcf_name = "\S+(?<!union)"
```
rdmorin: What does this actually do?

modules/starfish/2.0/starfish2.smk
```diff
@@ -0,0 +1,247 @@
...
+        op.relative_symlink(input.tbi, output.tbi)
+
+rule _starfish_output_venn: 
+    input: 
+        venn = str(rules._starfish_run.output.venn)
+    output: 
+        venn = CFG["dirs"]["outputs"] + "venn/{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}.venn.pdf"
+    run: 
+        op.relative_symlink(input.venn, output.venn)
+
```
rdmorin: Can you please add one or two comment lines explaining what these complex functions do?

rdmorin: To clarify... "get the checkpoint output directory" means literally nothing to me

modules/starfish/2.0/starfish2.smk
```diff
@@ -0,0 +1,247 @@
...
+    checkpoint_output = os.path.dirname(checkpoints._starfish_rename_output.get(**wildcards).output.renamed)
+    # Rename each VCF with the variant caller names
+    vcfs = expand(
+        CFG["dirs"]["outputs"] + "vcf/{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}.{vcf_name}.vcf.gz",
+        seq_type = wildcards.seq_type,
+        genome_build = wildcards.genome_build, 
+        tumour_id = wildcards.tumour_id, 
+        normal_id = wildcards.normal_id, 
+        pair_status = wildcards.pair_status, 
+        vcf_name = glob_wildcards(os.path.join(checkpoint_output, "{vcf_name}.vcf.gz")).vcf_name
```
rdmorin: I can't follow this but trust it works

modules/starfish/2.0/starfish2.smk
```diff
@@ -0,0 +1,247 @@
...
+        )
+    return vcfs
+
+
+rule _starfish_dispatch:
```
rdmorin: Why (or do) we need a dispatch if the existence of the union file depends on all the others completing? I am not saying it should change but I'm just wondering.

workflows/reference_files/2.4/reference_files.smk
```diff
@@ -91,6 +91,16 @@ rule get_liftover_chains:
     shell:
         "ln -srf {input.chains} {output.chains}"
 
+rule get_sdf_refs: 
+    input: 
+        sdf = rules.download_sdf.output.sdf
+    output: 
+        sdf = directory("genomes/{genome_build}/sdf")
+    wildcard_constraints: 
+        genome_build = "hg38|hg19|grch37"
```
rdmorin: How are we going to deal with ICGC genomes?

rdmorin (changes requested):
Some minor stuff. Mostly questions. Is it too much to ask for this to be incorporated into the demo workflow to show how it works with the available callers?

modules/starfish/2.0/config/default.yaml
```diff
@@ -11,9 +11,8 @@ lcr-modules:
             # Each key:value pair is the name of the variant caller and the path to its VCF file
             # Each key should be unique in this set. 
             vcf: 
-                strelka: "results/gambl/strelka-1.1/99-outputs/vcf/{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}.strelka.combined.vcf.gz"
-                lofreq: "results/gambl/lofreq-1.0/99-outputs/{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}.lofreq.snvs.vcf.gz"
```
rdmorin: I kinda liked it with all of these listed but if they go in the demo that's even better.

modules/starfish/2.0/config/default.yaml
```diff
@@ -0,0 +1,51 @@
...
+            bcftools: "{MODSDIR}/envs/bcftools-1.10.2.yaml"
+            starfish: "{MODSDIR}/envs/starfish.yaml"
+            
+        threads:
+            starfish_run: 4
+            starfish_union: 4
+
+        resources:
+            starfish_run: 
+                mem_mb: 15000
```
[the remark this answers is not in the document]
lkhilton: It was for 90% of the files, but for the others I had to up it to 100000. I'm wary of making such a high number the default because it can slow its progression on the cluster. But maybe I should add the attempt feature here.

modules/starfish/2.0/config/default.yaml
lkhilton: I was _trying_ to find a way to have them blank or commented out in the default config so users don't have to copy the whole default.yaml, but if there wasn't a placeholder here it wouldn't get overwritten by the values in my GAMBL config. I also wanted the `__UPDATE__` value here because it throws a helpful error if the user forgets to set this value. You don't think having one example is good enough?

modules/starfish/2.0/starfish2.smk
lkhilton: Oh yeah I can delete the not_union_vcf variable. The rest combines the names of the variant callers and appends "union" to it, e.g. "strelka_lofreq_mutect_union"

workflows/reference_files/2.4/reference_files_header.smk
lkhilton: Supposedly the online ones have additional annotations that you can't get from making your own. I don't know how much benefit it would be for Starfish, but 🤷

modules/starfish/2.0/config/default.yaml
lkhilton: Ah yes I will add a comment here, thanks.

modules/starfish/2.0/starfish2.smk
lkhilton: RIGHT!?

modules/starfish/2.0/starfish2.smk
lkhilton: Pending the addition of `op.prepare_resources()` to the oncopipe package...

modules/starfish/2.0/config/default.yaml
rdmorin: It's probably good enough, but if the minimum will always be 2 why not put two? Notably, what happens if someone decides to use this tool with two or three variant callers that _aren't_ strelka? Does this produce an error that would require them to copy the config anyway?

modules/starfish/2.0/config/default.yaml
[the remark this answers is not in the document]
rdmorin: I 100% think the attempt feature should be used here. This is what got me thinking last night about the current conflict between resource unpacking and that feature.

modules/starfish/2.0/starfish2.smk
rdmorin: I get that. I just didn't understand why the not_union one existed. Was this an attempt to construct a regex?

workflows/reference_files/2.4/reference_files.smk
lkhilton: Good question. I can probably just symlink the grch37 build to the hs37d5 directory, they're equivalent enough for these purposes.

modules/starfish/2.0/starfish2.smk
lkhilton: It was the only way to generate the symlinks once per VCF.

modules/starfish/2.0/starfish2.smk
lkhilton: This is the replacement for the "not_union_vcf" variable. It's a negative lookbehind assertion, meaning this value is restricted to those that don't end with "union"

rdmorin: I figured. Seems overly complex. Why can't the same rule output the union and non-union VCFs? I'm sure there's a good reason but I can't glean it from this, mostly because I don't understand much of the other stuff going on within checkpoints.

workflows/reference_files/2.4/reference_files.smk
rdmorin: good call

modules/starfish/2.0/src/starfish.py
```diff
@@ -0,0 +1,467 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
lkhilton: No, I added a function to throw an error if the packages for venn diagram drawing aren't available. Otherwise the script runs happily with no explanation as to why the venn wasn't produced.

modules/starfish/2.0/starfish2.smk
lkhilton: Ooh, I did that to make sure the cookie cutter didn't overwrite the exisiting Starfish 1.0 snakefile. I'll drop the 2.

modules/starfish/2.0/src/starfish.py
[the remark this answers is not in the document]
rdmorin: Can you document that change in the code if you haven't already?

modules/starfish/2.0/starfish2.smk
lkhilton: If I call `.keys()` again I also have to call the `.lower` function again, so I think this is more efficient purely for that reason. 

But seriously one of the most annoying things about Python is that you have to explicitly copy a list to a new variable.

rdmorin: Perl is so much better for this kind of crap. It's a lazy language for lazy coders. 
Is this call to "lower" because you don't trust the user to use the lowercase name? Isn't your config already lowercase? I must be missing something here.

modules/starfish/2.0/config/default.yaml
lkhilton: Added some comments to the config to clarify.

modules/starfish/2.0/config/default.yaml
lkhilton: I think they're usually named for the version of the core tool. Since the starfish one is just dependencies and Starfish itself is neither installed via Conda nor versioned it doesn't apply here.

modules/starfish/2.0/config/default.yaml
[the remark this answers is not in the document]
lkhilton: I am going to add the attempt feature in a future version bump when we can implement it more widely.

modules/starfish/2.0/src/starfish.py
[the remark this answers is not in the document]
lkhilton: Documented in changelog.

modules/starfish/2.0/config/default.yaml
lkhilton: The demo should clarify how the config should be set up.

modules/starfish/2.0/starfish2.smk
lkhilton: I don't want to assume the user has used all lowercase letters in the keys in the config. Since my function to replace the generic Starfish output vcf names with tool names depends on case-sensitive find-and-replace, I wanted to guarantee that there were no upper case letters in these keys that might inadvertently get replaced and garble the output file names.

lkhilton:
@rdmorin this is ready for re-review.

oncopipe/oncopipe/__version__.py
```diff
@@ -6,7 +6,7 @@
```
rdmorin, line 6: Were these modifications intentionally included in this PR?

oncopipe/oncopipe/__init__.py
```diff
@@ -244,7 +244,7 @@ def absolute_symlink(src, dest, overwrite=True):
```
rdmorin, line 244: Was this change intentionally included in this PR?

modules/starfish/2.0/starfish2.smk
rdmorin: Thank you for adding detailed in-line comments in the code. It is much more clear now.

modules/starfish/2.0/starfish2.smk
rdmorin: Thank you for explaining how wildcard unpacking is being used here.

rdmorin (approved):
I wasn't sure if you wanted the oncopipe updates merged into Master as a part of this PR or if those were brought in unintentionally. In any case, I haven't reviewed those. I seem to recall Bruno offered to review those updates. It is probably better to have his eyes on it because I'm not familiar with that code.

oncopipe/oncopipe/__version__.py
lkhilton: Yeah, this is the solution I came up with to fix the broken relative_symlink function, which was approved by Bruno in a previous PR. This retains the original relative_symlink functionality for backwards compatibility, but adds an option for making within-module symlinks truly relative. Hopefully a best-of-both-worlds solution.

lkhilton: Also I forgot to increase the version when I changed oncopipe previously, so this fixes that.

PR #141, opened 2020-11-20
Module/battenberg/1.0

rdmorin:
# Pull Request Checklists

### Required

Can we document what the purpose of this is? I haven't done it and am wary of breaking the module. Feel free to request this change in the PR. 

 There are a few of my own TODO comments but more notes for future iterations/improvements. 

### If applicable

See Changelog. This will be more difficult than for most tools. 


modules/battenberg/1.0/battenberg.smk
```diff
@@ -0,0 +1,187 @@
...
+    log:
+        stdout = CFG["logs"]["battenberg"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}/{tumour_id}_battenberg.stdout.log",
+        stderr = CFG["logs"]["battenberg"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}/{tumour_id}_battenberg.stderr.log"
+    params:
+        reference_path = lambda w: _battenberg_CFG["reference_path"][w.genome_build],
+        script = CFG["inputs"]["battenberg_script"],
+        calc_sex_status = CFG["inputs"]["calc_sex_status"],
+        x_chrom = lambda w: _battenberg_CFG["options"]["x_chrom"][w.genome_build],
+        y_chrom = lambda w: _battenberg_CFG["options"]["y_chrom"][w.genome_build],
+        chr_prefixed = lambda w: _battenberg_CFG["options"]["chr_prefixed_reference"][w.genome_build],
```
Kdreval, line 96: I might not be correct, but I think these lambda functions would not handle the hs37d5 reference. Maybe, we can try `op.switch_on_wildcard()` here? In this case, the lambda function here will become, for example, `op.switch_on_wildcard("genome_build", config["switches"]["x_chrom"])`, and in the config file the `options` will be changed to `switches`.

modules/battenberg/1.0/config/default.yaml
```diff
@@ -0,0 +1,50 @@
...
+                grch37: " "
+            x_chrom: 
+                grch37: "X"
+                hg38: "chrX"
+            y_chrom: 
+                grch37: "Y" 
+                hg38: "chrY"
+
+        conda_envs:
+            battenberg: "{MODSDIR}/envs/battenberg_fresh.yml"
```
Kdreval, line 31: I think with this, there are 2 extra yml files in this PR that I think are not refereed to, `battenberg_fresh.yaml` (seems to be in the parent directory) and `modules/battenberg/1.0/envs/battenberg-1.0.yaml`

modules/battenberg/CHANGELOG.md
```diff
@@ -0,0 +1,17 @@
...
+
+This installation is currently accomplished by a _install_battenberg rule but there is probably a better way (subworkflow?).  
+Obtaining and setting up reference directories for Battenberg is CURRENTLY UP TO THE USER. The forked repository (https://github.com/morinlab/battenberg) describes how this can be achieved. I decided to forego auto-installation of references due to some finnicky steps required (e.g. creating a file that includes the full path to each reference file on the local filesystem). I am open to suggestions for how to elegantly handle this annoyance. 
+
+In contrast to the Sequenza module, I haven't embedded cnv2igv.py in the module. Instead, the user must specify the path to their LCR-scripts repository. Similar to my original version of Sequenza, I relied on the calc_sex_status.sh script because it was working. Suggestions for more elegant solutions are welcome.
```
Kdreval, line 17: I think cnv2igv has been added since then, and is used now in the `_battenberg_to_igv_seg` rule

modules/battenberg/CHANGELOG.md
```diff
@@ -0,0 +1,17 @@
...
+
+## [1.0] - 2020-06-17
+
+This release was authored by Ryan Morin.
+
+I modified the Battenberg source code to take an optional "chr_prefixed_genome" argument and some modifications to enable battenberg to work with genomes that have "chr" prefixes in the chromosome names. The installation uses conda to create an R installation with all required prerequisites with the exception of ASCAT and Battenberg, which are installed from github. The modified R script that runs the pipeline is bundled with this tool but this may not be necessary for future releases. 
+
+This installation is currently accomplished by a _install_battenberg rule but there is probably a better way (subworkflow?).  
+Obtaining and setting up reference directories for Battenberg is CURRENTLY UP TO THE USER. The forked repository (https://github.com/morinlab/battenberg) describes how this can be achieved. I decided to forego auto-installation of references due to some finnicky steps required (e.g. creating a file that includes the full path to each reference file on the local filesystem). I am open to suggestions for how to elegantly handle this annoyance.
```
Kdreval, line 15: we can download the reference files from different links in the reference_files workflow, but I think it will make more sense to leave it up to the user as is now. I hope hg38 native support is coming to battenberg, and then it will easier to download all references from the same source. Otherwise, I work on reference files support similar to the leftover chains I recently made

modules/battenberg/1.0/battenberg.smk
rdmorin: I've run this on all three genome builds but it could be that it defaults to a sex (silently) when this fails. We should confirm for sure.

modules/battenberg/1.0/config/default.yaml
rdmorin: Can you comment directly on which ones should be deleted?

modules/battenberg/CHANGELOG.md
rdmorin: Not sure what you mean

modules/battenberg/CHANGELOG.md
rdmorin: Feel free to look at how they tell the user to set up their references. There's some file manipulation that I'm sure could be accomplished but it will be finicky

battenberg_fresh.yaml
```diff
@@ -0,0 +1,219 @@
+name: battenberg-fresh
```
Kdreval: I don't see a reference to this environment

modules/battenberg/1.0/envs/battenberg-1.0.yaml
```diff
@@ -0,0 +1,224 @@
+name: /home/rmorin/rrDLBCL_trios/.snakemake/conda/2f3a0b06
```
Kdreval, line 1: I don't see a reference of this environment

battenberg_fresh.yaml
rdmorin: Where is this file? I can't tell what the path is and I don't see it in my local repo

modules/battenberg/1.0/envs/battenberg-1.0.yaml
rdmorin: I'm a little worried the other one isn't complete. Can you test that the one used in the module works in a fresh environment (i.e. not in GAMBL)? If so, I'll delete this one.

battenberg_fresh.yaml
Kdreval: It does not show a path, only a file name. This file shows as "Outdated" now. Maybe it is something with git, and this file has been deleted/mowed since then

rdmorin: I found it. I had to "pull" to the repo. It's now deleted

modules/battenberg/1.0/battenberg.smk
rdmorin: I've checked that this is working as expected for the genomes that are handled.

lkhilton:
But is it in the demo??? 👀

rdmorin:
> But is it in the demo??? 👀

I deserve that. My response is: It's not meant for exome data.

rdmorin:
I would really appreciate it if this could be reviewed for completeness or any outstanding changes that are needed.

Kdreval (approved):

PR #145, opened 2020-12-07
Lofreq patch

rdmorin:
# Pull Request Checklists
See Changelog.


modules/lofreq/1.0/lofreq.smk
```diff
@@ -129,10 +133,13 @@ rule _lofreq_filter_vcf:
         CFG["conda_envs"]["bcftools"]
     shell:
         op.as_one_line("""
-        bash {input.lofreq_filter} {input.vcf_all} | bgzip > {output.vcf_all_clean}
+        PATH={SCRIPT_PATH}:$PATH;
+        SCRIPT=$(which lofreq_filter.sh); 
+        echo $SCRIPT;
```
lkhilton: Can you make this more informative? E.g. `echo "Using the script at $SCRIPT to filter lofreq calls"`

modules/lofreq/1.0/lofreq.smk
```diff
@@ -73,6 +75,9 @@ rule _lofreq_run:
         **CFG["resources"]["lofreq"]
     shell:
         op.as_one_line("""
+        PATH={SCRIPT_PATH}:$PATH;
+        SCRIPT=$(which lofreq2_call_pparallel.py);
+        echo $SCRIPT;
```
lkhilton: Same as my comment below on line 138 - make a user-readable output.

modules/lofreq/1.0/lofreq.smk
```diff
@@ -73,6 +75,9 @@ rule _lofreq_run:
         **CFG["resources"]["lofreq"]
     shell:
         op.as_one_line("""
+        PATH={SCRIPT_PATH}:$PATH;
+        SCRIPT=$(which lofreq2_call_pparallel.py);
```
lkhilton: I wonder if using the path to the script instead of `which` here would help ensure it's not inadvertently using the Conda-installed version. Then you could include something like `if [[ $SCRIPT == "" ]]; then echo "Lofreq script not found in module directory. Please ensure you are using the latest version from Master.; fi`

modules/lofreq/1.0/src/lofreq_filter.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
lkhilton, line 1: What's the rationale for this change? I always use `#!/bin/bash` in my scripts, curious what the difference is.

lkhilton (changes requested):

modules/lofreq/1.0/lofreq.smk
rdmorin: I'm not sure I understand. How would I use the path to the script? Wouldn't this check only work if the user is already using the latest version?

modules/lofreq/1.0/lofreq.smk
rdmorin: Good call. Done

modules/lofreq/1.0/lofreq.smk
rdmorin: Done but not tested. Do you happen to know if quotes necessary?

lkhilton: I would use quotes to be safe, yes, to make sure each word in the echo phrase isn't parsed as a separate argument.

modules/lofreq/1.0/lofreq.smk
rdmorin: I understand now. That would indeed protect against any unexpected environment issues. In theory prepending to PATH should make this check unnecessary but from recent experience, it's better not to assume this will always work.

rdmorin:
I believe this is ready for re-review, @lkhilton

lkhilton (approved):

PR #147, opened 2020-12-09
Module/mixcr/1.0

lkhilton:
This is an update to mixcr-1.0 and will be migrated to version 1.1. Changes are mostly minor - please see change log. 

Also includes a minor fix to the bam2fastq module and reference files workflow.


modules/mixcr/1.0/config/default.yaml
```diff
@@ -16,12 +17,17 @@ lcr-modules:
...
             
         threads:
             mixcr_run: 4
 
-        mem_mb:
-            mixcr_run: 6000
+        resources:
+            mixcr_run: 
+                mem_mb: 6000
+                fastq: 1
```
Kdreval: I wonder if this should be the default of 2 for fastq1 and fastq2?

modules/mixcr/1.0/mixcr.smk
```diff
@@ -70,6 +70,8 @@ rule _mixcr_run:
     input:
         fastq_1 = rules._mixcr_input_fastq.output.fastq_1,
         fastq_2 = rules._mixcr_input_fastq.output.fastq_2,
+        fastq_1_real = CFG["inputs"]["sample_fastq_1"], # Prevent premature deletion of fastqs marked as temp
+        fastq_2_real = CFG["inputs"]["sample_fastq_2"],
```
Kdreval: sorry, I didn't get how this prevents premature deletion of fastq files. Is it because they are referenced in config? They don't seem to be referenced in the rules further down.

Kdreval (approved):
just have some minor questions

modules/mixcr/1.0/mixcr.smk
lkhilton: Snakemake deletes a temp file after all the rules that use that file have been completed. If the real fastq file is only referenced during the rule `_mixcr_input_fastq`, the would be deleted as soon as the symlink is created. To get around that I specify them as inputs to the actual `_mixcr_run` rule, even though they aren't used in the command. That way they won't get deleted until after the mixcr run is finished.

modules/mixcr/1.0/config/default.yaml
lkhilton: I think of it as the equivalent of the single bam file, e.g. how many samples I want to run at one time.

lkhilton:
Thanks for the quick review, @Kdreval!

modules/mixcr/1.0/config/default.yaml
Kdreval: I see, so when you run you specify fastq as n of samples * 2?

lkhilton: Well no, because this value really specifies how many instances of this rule will be run simultaneously, and since each run utilizes 2 fastqs, if you specify `--resources fastq=10` on the command line, it will use 20 fastqs at a time.

lkhilton:
Could you approve the version bump, @Kdreval?

Kdreval (approved):

lkhilton:
Sorry Kostia, I ran into a memory problem as soon as you approved this last night. I added a jvmheap memory parameter and increased the default memory request, do you mind approving one last time?

Kdreval (approved):
no problem!

Kdreval (approved):

PR #155, opened 2020-12-21
Control-freec patch with hard-masked genomes

Jwong684:
# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable

To be completed.

# Update patch:

Reference Workflow:
- Included hard_masked genomes (hg38, hg19, grch38, grch37) as new "genome_builds" in the config
- Included a new rule to download hard_masked genomes specifically and unzip them
- Used wildcard constraints to distinguish hard_masked genomes vs. standard genomes

Controlfreec:
- Added GEM mappability generation for hard-masked genomes (constraint with wildcards)
- i.e. will generate mappability files if the genome_build is hard_masked. Standard genomes just simply uses the mappability file downloaded directly from the control-freec website.


modules/controlfreec/1.2/etc/scripts/freec2circos.pl
```diff
@@ -0,0 +1,156 @@
+#!/usr/bin/env perl -w
```
lkhilton: This looks handy, but I couldn't find a rule in the Snakefile to generate circos plots?

modules/controlfreec/1.2/controlfreec.smk
```diff
@@ -0,0 +1,461 @@
...
+                            "hg38": " "}[w.genome_build],
+        outdir = CFG["dirs"]["inputs"] + "references/{genome_build}/freec/"
+    wildcard_constraints:
+        genome_build = ".+(?<!masked)"
+    shell:
+        "wget -O {output.tar} {params.url} "
+        "&& {params.command1} {output.tar} {params.command2} {params.outdir}{params.command3}"
+
+# mappability tracks for hard-masked genomes need to be generated using GEM
+rule _download_GEM:
```
lkhilton, line 105: It could be useful to make the location where GEM is installed user-configurable via the config. That way if they want to install software somewhere central that option is available.

lkhilton (changes requested):
Looking good! One minor request is that you store all the scripts in `modules/controlfreec/1.2/src` instead of `modules/controlfreec/etc/scripts` for consistency with all the other modules that come with packaged scripts. 

Is there/will there be a rule in the snakefile to generate circos plots?

modules/controlfreec/1.2/config/default.yaml
```diff
@@ -0,0 +1,101 @@
...
+            ploidy: 2 #will select the ploidy that explains the most CNAs (a range can be added and control-freec will assign ploidy based on best fit, ex. 2,3,4)
+            printNA: FALSE
+            telocentromeric: 50000 # size of pre-telomeric and pre-centromeric regions to exclude
+            uniqueMatch: TRUE # uses mappability profile to correct read counts
+            
+            #optional options: (uncomment these options in config_WGS.txt to implement them)
+            #if implemented, contamination will overrule contaminationAdjustment
+            #step and window will overrule coefficientOfVariation
+            contamination: 0 # a priori known value of tumor sample contamination by normals
+            sex: XX or XY # sex=XX will exclude chrY and sex=XY will not annotate one copy of X and Y as a loss
```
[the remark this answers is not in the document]
Jwong684: Unfortunately, the documentation for this is one of those grey-areas where the author doesn't really explain what she did here. The default value in the online manual had it as "XX or XY" and the default output so far would show some samples with X only and some with Y.

switch_on_column could be something interesting for later though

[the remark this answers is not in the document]
lkhilton: Weird, but ok!

modules/controlfreec/1.2/etc/scripts/freec2circos.pl
Jwong684: I haven't implemented this part yet. I can throw it in though if you want that as an option. It just translates the ratio file into a Circos track

modules/controlfreec/1.2/controlfreec.smk
Jwong684: Okay I can add a SOFTWARE_DIR variable in the config

modules/controlfreec/1.2/etc/scripts/freec2circos.pl
Jwong684: (It doesn't actually generate the circos plot though)

workflows/reference_files/2.4/reference_files_header.smk
```diff
@@ -124,11 +126,25 @@ rule download_genome_fasta:
...
 
+rule download_masked_genome_fasta:
+    output:
+        fasta = "downloads/genome_fasta/{genome_build}.fa"
+    wildcard_constraints:
+        genome_build = ".+_masked"
+    params:
+        dirOut = "downloads/genome_fasta/",
+        url = lambda w: config["genome_builds"][w.genome_build]["genome_fasta_url"]
+    conda: CONDA_ENVS["bedops"]
```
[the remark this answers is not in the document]
Jwong684: Oh shoot it was originally there to test out the curl function when I was having trouble with wget before. I will remove it.

rdmorin:
What's left to do on this, @Jwong684 ?

Jwong684:
Just need to add a configurable SOFTWARE_DIR for GEM and add a rule to generate a circos track

rdmorin:
@Kdreval and @lkhilton Any reason why this is still stuck in a PR? Are there more changes needed?

Kdreval:
I think it has been run successfully and works as expected. The only suggestion I had was to generate seg files as Controlfreec output, either by adding script here or by updating cnv2igv.

rdmorin:
What's the output format currently? Do you have an example?

Kdreval:
The current format is CNV file, for example:
```
chr	start	end	copy number	status	genotype	uncertainty	somatic/germline	precentageOfGermline	WilcoxonRankSumTestPvalue	KolmogorovSmirnovPvalue
1	89208684	89807400	2	neutral	AA	71.4154	somatic	0	0.00685763509471829	0.00457486617089953
10	129389180	129921372	2	neutral	AA	82.0433	somatic	0	7.22174241145704e-06	4.17080331172048e-06
11	107968452	110363316	2	neutral	AA	5.83435	somatic	0	0.857102561183487	0.192080849466446
13	45236320	47165516	4	gain	AAAB	0.220032	somatic	0	3.78744827066444e-19	0
```
It is easy to convert, but it would be very convenient if this is generated by the module directly and will not require any post-processing. The outputs are currently not IGV- or GISTIC-friendly

rdmorin:
Yeah. Seems like cnv2igv should be updated and should run automatically

Jwong684:
Do you mean you want cnv2igv to run as another rule in the module? Would there be issues later on if cnv2igv gets updated?

Aside: Control-freec module is now CRAM-compatible.

Kdreval:
It can be implemented as a separate rule, yes. There is an example in Sequenza or Battenberg modules, [for example](https://github.com/LCR-BCCRC/lcr-modules/blob/b701fffe69606b48d480b838c09e53b2bcb59aef/modules/battenberg/1.2/battenberg.smk#L214-L230). There would be no issue if the cnv2igv will be updated, because it is tracked by [lcr-scripts](https://github.com/LCR-BCCRC/lcr-scripts/blob/master/cnv2igv/1.4/cnv2igv.py) repo, and it is always possible to refer to specific version. The latest one is 1.4, and Controlfreec support can become version 1.5.

Jwong684:
hm I see. Would you want to add the control-freec support for cnv2igv? I can follow your Battenberg rule and integrate it into my module (using same inputs/params)

Kdreval:
Sure, I will work on the new version of cnv2igv with Controlfreec support

Kdreval:
@Jwong684 , I have made a modification to cnv2igv that supports Controlfreec outputs and made a PR: https://github.com/LCR-BCCRC/lcr-scripts/pull/24 There is usage statement example in the description, and it can be used similar to other modules. Can you please review it? If it looks good, I will then update the version to 1.5

Jwong684:
@Kdreval Okay. The rule is now up-dated. I gave it a test-run on a sample and it works fine. Also updated to the new lcr-module conventions.

Kdreval (approved):

Kdreval:
Thank you Jasper! It will really make it easier to view the outputs now and use in the downstream analysis. I think it is good to go now. I approved PR, but I could not merge it - it is waiting for Laura's approval. @lkhilton , when you have time, can you please approve it if there are no questions/updates left? Thanks!

demo/config.yaml
```diff
@@ -7,6 +7,7 @@ lcr-modules:
         scratch_directory: "scratch/"
         unmatched_normal_ids:
             capture--grch37: "TCRBOA7-N-WEX"
+            genome--hg38: "BLGSP-71-23-00361-14A-01E"
```
lkhilton: We don't need this line in the lcr-module version. I'll remove it but FYI you'll have to add it to your GAMBL config if it isn't already there.

PR #156, opened 2020-12-22
Battenberg with cram support

rdmorin:
Because of the restructuring, I haven't modified the files in place. I'll add comments to highlight the main changes.


modules/battenberg/1.1/battenberg.smk
```diff
@@ -0,0 +1,220 @@
...
+# Setup module and store module-specific configuration in `CFG`
+# `CFG` is a shortcut to `config["lcr-modules"]["battenberg"]`
+CFG = op.setup_module(
+    name = "battenberg",
+    version = "1.1",
+    subdirectories = ["inputs", "infer_sex","battenberg", "outputs"],
+)
+
+#set variable for prepending to PATH based on config
+SCRIPT_PATH = CFG['inputs']['src_dir']
```
rdmorin, line 45: added to simplify use of bundled scripts

modules/battenberg/1.1/battenberg.smk
```diff
@@ -0,0 +1,220 @@
...
+##### RULES #####
+
+# Symlinks the input files into the module results directory (under '00-inputs/')
+rule _battenberg_input_bam:
+    input:
+        bam = CFG["inputs"]["sample_bam"]
+    output:
+        bam = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{sample_id}.bam",
+        bai = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{sample_id}.bam.bai",
+        crai = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{sample_id}.bam.crai"
```
rdmorin, line 65: added for cram support

modules/battenberg/1.1/battenberg.smk
```diff
@@ -0,0 +1,220 @@
...
+    group: "setup_run"
+    run:
+        op.relative_symlink(input.bam, output.bam)
+        op.relative_symlink(input.bam + ".bai", output.bai)
+        op.relative_symlink(input.bam + ".bai", output.crai)
+
```
rdmorin, line 71: I am still hoping for some suggestions on how to better handle this. I am not particularly happy with the installation status being written to "config/envs".

modules/battenberg/1.1/battenberg.smk
```diff
@@ -0,0 +1,220 @@
...
+        complete = "config/envs/battenberg_dependencies_installed.success"
+    conda:
+        CFG["conda_envs"]["battenberg"]
+    shell:
+        """
+        R -q -e 'devtools::install_github("Crick-CancerGenomics/ascat/ASCAT")' && ##move some of this to config?
+        R -q -e 'devtools::install_github("morinlab/battenberg")' &&              ##move some of this to config?
+        touch {output.complete}"""
+
+# this process is very fast on bam files and painfully slow on cram files.
```
rdmorin, line 86: This is a new rule

modules/battenberg/1.1/battenberg.smk
```diff
@@ -0,0 +1,220 @@
...
+    shell:
+        op.as_one_line("""
+        PATH={SCRIPT_PATH}:$PATH; 
+        echo "running {rule} for {wildcards.normal_id} on $(hostname) at $(date)" > {log.stderr} ;
+        calc_sex_status.sh {input.normal_bam} {wildcards.normal_id} > {output.sex_result} 2>> {log.stderr}; 
+        sleep 20
+        """)
+
+
+# This rule runs the entire Battenberg pipeline. Eventually we may want to set this rule up to allow re-starting
```
rdmorin, line 108: Reworked to support cram by passing genome_fasta

modules/battenberg/1.1/battenberg.smk
```diff
@@ -0,0 +1,220 @@
...
+        script = CFG["inputs"]["battenberg_script"],
+        chr_prefixed = lambda w: _battenberg_CFG["options"]["chr_prefixed_reference"][w.genome_build],
+        out_dir = CFG["dirs"]["battenberg"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}"
+    conda:
+        CFG["conda_envs"]["battenberg"]
+    resources:
+        **CFG["resources"]["battenberg"]
+    threads:
+        CFG["threads"]["battenberg"]
+    group: "setup_run"
```
rdmorin: added group to ensure quick jobs run on the same node as slow jobs without the user specifying this

modules/battenberg/1.1/src/battenberg_wgs_hg38.R
```diff
@@ -0,0 +1,135 @@
...
+  make_option(c("--skip_preprocessing"), type="logical", default=FALSE, action="store_true", help="Provide when pre-processing has previously completed. This expects the files on disk", metavar="character"),
+  make_option(c("--skip_phasing"), type="logical", default=FALSE, action="store_true", help="Provide when phasing has previously completed. This expects the files on disk", metavar="character"),
+  make_option(c("--cpu"), type="numeric", default=8, help="The number of CPU cores to be used by the pipeline (Default: 8)", metavar="character"),
+  make_option(c("--bp"), type="character", default=NULL, help="Optional two column file (chromosome and position) specifying prior breakpoints to be used during segmentation", metavar="character"),
+  make_option(c("--reference"), type="character", default=NULL, help="Path to reference files", metavar="character"),
+  make_option(c("-f","--reference_fasta"), type="character", default=NULL, help="Path to indexed genome fasta file (needed for CRAM compatability)", metavar="character"),
```
rdmorin, line 20: this is the added option

modules/battenberg/1.1/src/calc_sex_status.sh
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
```
rdmorin, line 1: overhauled script to infer sex from the bam

modules/battenberg/1.0/battenberg.smk
```diff
@@ -46,6 +46,7 @@ rule _battenberg_input_bam:
     output:
         bam = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{sample_id}.bam",
         bai = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{sample_id}.bam.bai"
```
rdmorin: I'm just realizing I should revert this file to the one in Master

modules/battenberg/1.1/battenberg.smk
```diff
@@ -0,0 +1,220 @@
...
+rule _battenberg_input_bam:
+    input:
+        bam = CFG["inputs"]["sample_bam"]
+    output:
+        bam = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{sample_id}.bam",
+        bai = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{sample_id}.bam.bai",
+        crai = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{sample_id}.bam.crai"
+    group: "setup_run"
+    run:
+        op.absolute_symlink(input.bam, output.bam)
```
rdmorin, line 68: ABSOLUTE!

modules/battenberg/1.1/battenberg.smk
```diff
@@ -0,0 +1,220 @@
...
+        sub = rules._run_battenberg.output.sub
+    output:
+        seg = CFG["dirs"]["outputs"] + "seg/{seq_type}--{genome_build}/{tumour_id}--{normal_id}_subclones.igv.seg",
+        sub = CFG["dirs"]["outputs"] + "txt/{seq_type}--{genome_build}/{tumour_id}--{normal_id}_subclones.txt"
+    params: 
+        batt_dir = CFG["dirs"]["battenberg"] + "/{seq_type}--{genome_build}/{tumour_id}--{normal_id}",
+        png_dir = CFG["dirs"]["outputs"] + "png/{seq_type}--{genome_build}"
+    group: "post_process"
+    run:
+        plots = glob.glob(params.batt_dir + "/*.png")
```
rdmorin, line 207: This works as advertised. :delicioso:

modules/battenberg/1.0/src/R/battenberg_wgs_hg38.R
```diff
@@ -17,6 +17,7 @@ option_list = list(
   make_option(c("--cpu"), type="numeric", default=8, help="The number of CPU cores to be used by the pipeline (Default: 8)", metavar="character"),
   make_option(c("--bp"), type="character", default=NULL, help="Optional two column file (chromosome and position) specifying prior breakpoints to be used during segmentation", metavar="character"),
```
rdmorin: Also needs to be reverted

rdmorin:
I'm still doing some testing but I'm now very happy to report that cram support seems to be working along with some other new features

modules/battenberg/1.1/battenberg.smk
Kdreval: Do you want to add assert here, to ensure the correct oncopipe version will be used? The same goes for `in_module` at the output rule.

modules/battenberg/1.1/battenberg.smk
Kdreval: I think it can be written to `CFG["inputs"]+"dependencies/"`. Also, in sequenza we install packages at the beginning of R script using if statement, so it is installed only once. It works there efficiently, and doesn't seem to have issues since the last update. So maybe this can be moved to the top of `battenberg_wgs_hg38.R`?

modules/battenberg/1.1/battenberg.smk
Kdreval: So it doesn't need reference for cram files here?

modules/battenberg/1.1/src/calc_sex_status.sh
```diff
@@ -0,0 +1,48 @@
...
+samtools idxstats $(readlink -e ${BAM}) \
+    | awk -v sample="$SAMPLE" ' \
+        BEGIN { \
+            FS=OFS="\t" \
+        } \
+        $1 == "'${X_CHROM}'" { \
+            chrX = $3 \
+        } \
+        $1 == "'${Y_CHROM}'" { \
+            chrY = $3 \
```
Kdreval: I got a bit lost here. Shouldn't this be `$4` for Y chromosome? Or more like `$2` for X chromosome, since it the number of column in the output file?

modules/battenberg/1.1/battenberg.smk
rdmorin: Sure. Do you have an example I can copy?

modules/battenberg/1.1/src/calc_sex_status.sh
rdmorin: I've re-written this script. Should be easier to follow now.

modules/battenberg/1.1/battenberg.smk
Kdreval: I believe this should work:
```
from packaging import version

try:
    assert version.parse(oncopipe.__version__) >= version.parse('1.0.11')
except AssertionError:
    raise( AssertionError( "Oncopipe version should be equal to or higher than 1.0.11. Please update oncopipe in your environment" ) )
```
It will need packaging added to the environment and can be installed from pip.
I can also think of a way to check the version without the need for additional packages and environment modification.

modules/battenberg/1.1/src/calc_sex_status.sh
Kdreval: Thank you! It is!

rdmorin:
I pushed a few bug fixes. Were there other unresolved issues I need to address?

modules/battenberg/1.1/battenberg.smk
rdmorin: Shouldn't this work? I see that it doesn't but I don't understand why. 
```
>>> print(op.__version__)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'oncopipe' has no attribute '__version__'
```

rdmorin: @Lakshay-sethi you probably know the best way to deal with this, right?

Kdreval:
I think only reverting 1.0 version files back and (possibly) check for the oncopipe version

rdmorin:
How exactly do I revert the unintentionally modified files from version 1?

Kdreval:
I usually delete the folder with edits and then merge master (which recreates the 1.0 version).  Maybe there is a better approach with git commands, but this way worked for me.

rdmorin:
Since my last few commits this is in need of more/re-review, especially 
706137a

Kdreval:
> Since my last few commits this is in need of more/re-review, especially
> [706137a](https://github.com/LCR-BCCRC/lcr-modules/commit/706137af6a969ecb96d6ca665d764571af232c95)

I have tested this iteration of oncopipe control version and can confirm it: exits if packaging is not installed, exits if oncopipe is leas than the minimal version, runs successfully if oncopipe is updated

Kdreval (approved):

PR #157, opened 2020-12-29
Module/sage/1.0

Kdreval:
# Pull Request Checklists

### Required

- [X] I tested the module on real data for all supported `seq_type` values and `genome_build`s.

- [X] I ensured that all symbolic links are relative and self-contained (_i.e._ do not point outside of the repository). This includes environment files.

  ```bash
  ```

### If applicable

- [X] I added more granular output subdirectories.


modules/sage/1.0/sage.smk
```diff
@@ -0,0 +1,255 @@
...
+        -reference_bam {input.normal_bam}
+        -tumor {wildcards.tumour_id} 
+        -tumor_bam {input.tumour_bam}
+        -assembly {params.assembly}
+        -ref_genome {input.fasta}
+        -hotspots {input.hotspots}
+        -panel_bed {input.panel_bed}
+        -high_confidence_bed {input.high_conf_bed}
+        -out {output.vcf}
+        >>  {log.stdout} 2> {log.stderr}
```
rdmorin: It shouldn't matter here because stderr was only used once but I try to habitually use >> for all my redirects to log files.

modules/sage/1.0/sage.smk
```diff
@@ -0,0 +1,255 @@
...
+    shell:
+        op.as_one_line("""
+        echo "running {rule} for {wildcards.tumour_id}--{wildcards.normal_id} on $(hostname)" > {log.stdout}
+        &&
+        java -Xms1G -Xmx{params.jvmheap}m
+        -cp {params.sage} com.hartwig.hmftools.sage.SageApplication
+        -threads {threads}
+        {params.opts}
+        -chr {params.chromosomes}
+        -reference {wildcards.normal_id}
```
rdmorin, line 160: I was puzzled by this at first. Reference is the name for the normal!

modules/sage/1.0/sage.smk
```diff
@@ -0,0 +1,255 @@
...
+        CFG["conda_envs"]["bcftools"]
+    threads:
+        CFG["threads"]["filter"]
+    resources:
+        **CFG["resources"]["filter"]
+    shell:
+        op.as_one_line("""
+        bcftools view -f ".,PASS" {input.vcf} -Ov
+          | 
+        bcftools sort --max-mem {resources.mem_mb}M -Oz -o {output.vcf_passed}
```
rdmorin: To be safe, should we use 0.8*mem_mb here as you did in the earlier example?

modules/sage/1.0/sage.smk
```diff
@@ -0,0 +1,255 @@
...
+        stderr = CFG["logs"]["vcf"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/split_passed.stderr.log"
+    conda:
+        CFG["conda_envs"]["bcftools"]
+    threads:
+        CFG["threads"]["filter"]
+    resources:
+        **CFG["resources"]["filter"]
+    shell:
+        op.as_one_line("""
+        bcftools view -v indels {input.vcf_passed} -Oz -o {output.indels} > {log.stdout} 2> {log.stderr}
```
rdmorin: I had no idea you could use bcftools view this way. Nice!

modules/sage/1.0/sage.smk
```diff
@@ -0,0 +1,255 @@
...
+          &&
+        tabix -p vcf {output.indels} >> {log.stdout} 2>> {log.stderr}
+          &&
+        bcftools view -v snps {input.vcf_passed} -Oz -o {output.snvs} >> {log.stdout} 2>> {log.stderr}
+          &&
+        tabix -p vcf {output.snvs} >> {log.stdout} 2>> {log.stderr}
```
rdmorin, line 228: Notably, you can also use "bcftools index"

modules/sage/CHANGELOG.md
```diff
@@ -0,0 +1,16 @@
...
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [1.0] - 2020-12-29
+
+This release was authored by Kostiantyn Dreval.
+
+- Reference files are symlinked to `00-inputs` from references directory to ensure all required indexes and dictionaries are created if starting on fresh directory.
+- SAGE fails to write to .gz files directly, therefore outputs are marked as `temp()` to ensure they are deleted after bgzip.
+- `main_chromosomes.txt` from reference_files workflow is used to restrict SAGE only to standard chromosomes. Chromosomes listed in this file are returned as list to be included with SAGE `-chr` flag.
+- module outputs separate VCF files for indels, snvs, and combined (indels+snsvs+mnvs). This can be modified in later releases.
```
rdmorin, line 15: what's a mnv?

modules/sage/1.0/sage.smk
```diff
@@ -0,0 +1,255 @@
...
+        CFG["threads"]["filter"]
+    resources:
+        **CFG["resources"]["filter"]
+    shell:
+        op.as_one_line("""
+        bcftools view -v indels {input.vcf_passed} -Oz -o {output.indels} > {log.stdout} 2> {log.stderr}
+          &&
+        tabix -p vcf {output.indels} >> {log.stdout} 2>> {log.stderr}
+          &&
+        bcftools view -v snps {input.vcf_passed} -Oz -o {output.snvs} >> {log.stdout} 2>> {log.stderr}
```
rdmorin, line 226: Does the sum of the variants in the two outputs add up to the combined or do we lose some here?

modules/sage/1.0/config/default.yaml
```diff
@@ -0,0 +1,43 @@
...
+        
+        threads:
+            sage_run: 24
+            filter: 1
+
+
+        pairing_config:
+            genome:
+                run_paired_tumours: True
+                run_unpaired_tumours_with: "unmatched_normal"
```
rdmorin, line 38: I agree with this implementation but I am wondering if we should also support running without a normal if that option is available. This could also be added to a later version.

rdmorin (changes requested):
Very clean first version. A few minor suggestions and comments. Ideally we should make version 1.0 support crams if we can.

modules/sage/1.0/sage.smk
```diff
@@ -0,0 +1,255 @@
...
+        indels = str(rules._sage_split_vcf.output.indels),
+        snvs = str(rules._sage_split_vcf.output.snvs),
+    output:
+        combined = CFG["dirs"]["outputs"] + "combined/{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}.sage.combined.vcf.gz",
+        combined_tbi = CFG["dirs"]["outputs"] + "combined/{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}.sage.combined.vcf.gz.tbi",
+        indels = CFG["dirs"]["outputs"] + "indels/{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}.sage.indels.vcf.gz",
+        indels_tbi = CFG["dirs"]["outputs"] + "indels/{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}.sage.indels.vcf.gz.tbi",
+        snvs = CFG["dirs"]["outputs"] + "snvs/{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}.sage.snvs.vcf.gz",
+        snvs_tbi = CFG["dirs"]["outputs"] + "snvs/{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}.sage.snvs.vcf.gz.tbi"
+    run:
```
rdmorin, line 246: I urge you to consider changing this to use Laura's new "in_module" feature of relative_symlink as I did in the Battenberg module. Don't change it for the inputs but do it for all outputs and any intermediate steps that make symlinks.

modules/sage/1.0/sage.smk
```diff
@@ -0,0 +1,255 @@
...
+##### RULES #####
+
+
+# Symlinks the input files into the module results directory (under '00-inputs/')
+rule _sage_input_bam:
+    input:
+        bam = CFG["inputs"]["sample_bam"]
+    output:
+        bam = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{sample_id}.bam",
+        bai = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{sample_id}.bam.bai"
```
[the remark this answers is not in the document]
Kdreval: I tested it and it works as-is with cram files, the only downside is increased running time from ~2.5h to ~ overnight.

modules/sage/1.0/sage.smk
```diff
@@ -0,0 +1,255 @@
...
+    output:
+        vcf = temp(CFG["dirs"]["sage"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/{tumour_id}--{normal_id}--{pair_status}.vcf"),
+        vcf_gz = CFG["dirs"]["sage"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/{tumour_id}--{normal_id}--{pair_status}.vcf.gz"
+    log:
+        stdout = CFG["logs"]["sage"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/run_sage.stdout.log",
+        stderr = CFG["logs"]["sage"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/run_sage.stderr.log"
+    params:
+        opts = CFG["options"]["sage_run"],
+        chromosomes = get_chromosomes,
+        assembly = lambda w: "hg38" if "38" in str({w.genome_build}) else "hg19",
```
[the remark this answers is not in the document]
Kdreval: I am not sure why it needs this, but I found it is common with some other hmftools. It is a required argument, so I came up with this way of parsing it.

modules/sage/1.0/sage.smk
```diff
@@ -0,0 +1,255 @@
...
+    input: 
+        genome_fa = reference_files("genomes/{genome_build}/genome_fasta/genome.fa")
+    output: 
+        hotspots = CFG["dirs"]["inputs"] + "references/{genome_build}/sage/KnownHotspots.vcf.gz",
+        panel_bed = CFG["dirs"]["inputs"] + "references/{genome_build}/sage/ActionableCodingPanel.somatic.bed.gz",
+        high_conf_bed = CFG["dirs"]["inputs"] + "references/{genome_build}/sage/HighConfidence.bed.gz"
+    params:
+        url = "www.bcgsc.ca/downloads/morinlab/hmftools-references/sage",
+        build = lambda w: "hg38" if "38" in str({w.genome_build}) else "hg19"
+    shell:
```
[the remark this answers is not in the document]
Kdreval: It is not hard-coded to have hg in naming. We can supply a reference file with any name, and I opted for this way of naming the files to be consistent with the required argument `assembly`. I can create files with the actual `genome_build`, as in the wildcard, but I thought in this way I will need to create a reference file (or at least a symlink) for each genome build present in lcr-modules, and that might increase the number of files.

modules/sage/1.0/sage.smk
Kdreval: Yes it is confusing to me as well! Why not name it `normal`, if it is fact not reference, but the normal sample.

modules/sage/1.0/sage.smk
Kdreval: I think it is a safer way, I might miss something, or someone else can be modifying this snakefile later. So I changed in to `>>` throughout. I think it's a safe practice and I'll add it to other modules too!

modules/sage/1.0/sage.smk
Kdreval: good suggestion, I have added this to the rule

modules/sage/1.0/sage.smk
Kdreval: I have included this option, and as well included the check for oncopipe version.

modules/sage/CHANGELOG.md
Kdreval: That's how they call multi-nucleotide variants, or clusters of 2 or more nearby variants.

modules/sage/1.0/sage.smk
Kdreval: It does, but not always since I do not subset vcf file for mnvs. In other words, if there are no mnvs, the sum adds up to the total number of variants. If there are mnvs, then sum of the 2 outputs+number of mnvs adds up to the number in `combined` vcf.

Kdreval:
> Very clean first version. A few minor suggestions and comments. Ideally we should make version 1.0 support crams if we can.

Thank you! I have addressed the comments and made some minor modifications. I also can confirm it works with cram files I tested. The updated version is ready for review.

Kdreval:
After I converted the vcfs to maf, I see it does not populate t_depth, t_ref_coun, t_alt_count. I will work on it to make them compatible.

Kdreval:
> After I converted the vcfs to maf, I see it does not populate t_depth, t_ref_coun, t_alt_count. I will work on it to make them compatible.

It was more related to vcf2maf config than to SAGE itself. I fixed it and it is populating reads now. No changes are required to the SAGE module.

modules/sage/1.0/config/default.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+
+        conda_envs:
+            sage: "{MODSDIR}/envs/hmftools-sage-2.2.yaml"
+            bcftools: "{MODSDIR}/envs/bcftools-1.10.2.yaml"
+            
+        resources:
+            sage_run:
+                mem_mb: 60000
+                disk_mb: 20000 # Will ensure at least this much space is left
```
rdmorin: reserving/expecting 20G per sample seems a bit high, no?

modules/sage/1.0/sage.smk
rdmorin: I see. So we lose mnvs if we use the individual split files. That's not ideal but I expect these are pretty rare, right?

rdmorin (approved):

rdmorin:
Ideally this would be incorporated into the demo but that can wait since we really need this tool in production.

modules/sage/1.0/config/default.yaml
Kdreval: the largest file from the output is 3.5 Gb, but because at some point both compressed and decompressed files may exist (right before the _sage_run is completed), I gave it more space to accommodate both of these files. But I might decrease that further down if needed.

Kdreval:
> Ideally this would be incorporated into the demo but that can wait since we really need this tool in production.

I did not add it to the demo since the demo currently is being modified. I thought to include it in the demo with the demo development.

rdmorin:
Did you need this re-reviewed?

Kdreval:
I have included it in the demo and decreased the disk_mb requirement. I have also tested it outside of the GSC network and found that main_chromosomes need to be added to the rule. These are all modifications. If you approve it, it will be ready to go, so yes, please.

rdmorin (approved):

Issue #154, opened 2020-12-21
liftover outputs are not sorted


Kdreval:
UCSC liftOver does not output sorted bed files after conversion. Therefore, the outputs of `liftover` module are not sorted as well. This needs to be resolved as it creates problems for downstream applications.

Kdreval:
Addressed in version 1.2, PR #160

PR #160, opened 2021-01-11
Module/liftover/1.1

Kdreval:
This is an updated version of the module. It includes several modification, please see CHANGELOG. It is also addressing issue [#154](https://github.com/LCR-BCCRC/lcr-modules/issues/154).


Lakshay-sethi (commented):
### 1 . envs/bedtools/bedtools-2.29.yaml
We can add conda-forge in channels.

### 2 . modules/liftover/1.1/config/default.yaml
I think we can change the following comment "Path to seg files for lifting. The {genome_build} will be handled to include hg38 samples only. For example:" as we have allowed support for bidirectional conversion of genomic coordinates.

### 3. modules/liftover/1.1/envs/liftover-366.yaml
I think we should keep the python dependency to 3.7 and not increase it to 3.9 because many of our other code bases are depended on 3.7, also it will create a need for the user to update python to 3.9 and can introduce many issues for them as other packages may not be fully supported for version 3.9.

### 4. modules/liftover/1.1/liftover.smk
No comments
### 5. modules/liftover/1.1/liftover.smk.save 
No comments

### 6.  modules/liftover/1.1/src/chains/hg38ToHg19.over.chain
No comments

### 7. modules/liftover/CHANGELOG.md

Kdreval:
Thank you Lakshay! 
I have removed the bedtools environment file since I later opted for the shell `sort` and do not need it included in the module. I have also updated the comment in the config file to reflect the latest changes. The Python version should not cause issues, since we do not share an environment between modules/rules.

Lakshay-sethi (approved):

Lakshay-sethi (approved):

lkhilton (approved):
Very nice work! This is a big enhancement to the existing module. 

Is this exclusively a copy number tool? One thing to consider might be to change the functionality of this module to something similar to the `utils` module, where it's not loaded as its own module but can be automatically implemented within every copy number tool. That's for another day though.

PR #162, opened 2021-01-16
Modules/slms 3/1.0

lkhilton:
This module amalgamates the Manta/Strelka, Mutect2, SAGE, and LoFreq variant calling modules to generate a final VCF of variants called by three or more of these callers, as well as a union VCF file that enables tracking of which variants were called by which variant callers. 

This PR includes minor updates to the Mutect2 and Starfish modules to enable more seamless integration with this pipeline. I don't think any of these updates requires a version increase. 

# Pull Request Checklists

### Required

This has been tested on genomes, and Jasper will be trying it on exomes. 

  ```bash
  ```

### If applicable


modules/mutect2/2.0/config/default.yaml
```diff
@@ -6,12 +6,18 @@ lcr-modules:
...
+            # For capture data, this can be used to specify the capture space. 
+            # If you don't want to provide candidate positions, leave this field as-is. 
+            # Set additional interval parameters under options: mutect2_interval_rules. 
+            # See https://gatk.broadinstitute.org/hc/en-us/articles/360035531852-Intervals-and-interval-lists for how-to. 
+            candidate_positions: "" 
 
         scratch_subdirectories: [] # Recommended: mutect2
 
         options:
-            mutect2_run_matched_unmatched: ""
```
rdmorin, line 13: Why did these ones get removed/replaced with other options if the new options are still left blank?

modules/mutect2/2.0/mutect2.smk
```diff
@@ -71,6 +77,18 @@ rule _mutect2_get_sm:
         "samtools view -H {input.bam} | grep '^@RG' | "
         r"sed 's/.*SM:\([^\t]*\).*/\1/g'"" | uniq > {output.sm} 2> {log.stderr}"
```
rdmorin, line 98: Could you please add a bit of in-line documentation explaining what this function within a function is doing? It might help for future maintenance.

modules/mutect2/2.0/mutect2.smk
```diff
@@ -47,6 +47,12 @@ rule _mutect2_input_bam:
         op.relative_symlink(input.bam, output.bam)
```
rdmorin, line 66: Does this work on cram files as-is? If not, could you please add the second crai symlink with the ".crai" name so it can be automatically found?

modules/slms_3/1.0/config/default.yaml
```diff
@@ -0,0 +1,180 @@
+lcr-modules:
```
rdmorin, line 1: It's too bad the skeleton config is needed to allow these values to be updated. I guess the update-config doesn't work if a key isn't already present?

modules/slms_3/1.0/slms_3.smk
```diff
@@ -0,0 +1,591 @@
...
+
+current_version = pkg_resources.get_distribution("oncopipe").version
+if version.parse(current_version) < version.parse(min_oncopipe_version):
+    print('\x1b[0;31;40m' + f'ERROR: oncopipe version installed: {current_version}' + '\x1b[0m')
+    print('\x1b[0;31;40m' + f"ERROR: This module requires oncopipe version >= {min_oncopipe_version}. Please update oncopipe in your environment" + '\x1b[0m')
+    sys.exit("Instructions for updating to the current version of oncopipe are available at https://lcr-modules.readthedocs.io/en/latest/ (use option 2)")
+
+# End of dependency checking section 
+
+# Use modified setup_module function to skip the global `CFG` check. This module assigns the module config to a variable called `CFG_SLMS3` so that the submodules can still use `CFG`.
```
rdmorin: Did (or can) you document what you modified in this setup_module function? In the current form it's not clear to me what necessitates this duplication of code.

modules/slms_3/1.0/slms_3.smk
```diff
@@ -0,0 +1,591 @@
...
+# `CFG_SLMS3` is a shortcut to `config["lcr-modules"]["slms_3"]`
+# This is not assigned to `CFG` as in other modules because the sub-modules here will use `CFG`. 
+CFG_SLMS3 = setup_module(
+    name = "slms_3",
+    version = "1.0",
+    subdirectories = ["inputs", "strelka_gnomad", "lofreq_gnomad", "strelka_lofreq_union", "sage_gnomad", "mutect2_depth_filt", "union", "outputs"],
+)
+
+# Ensure each of the submodule versions meets requirements. 
+
```
rdmorin, line 48: Seems like a handy feature we might want to roll right into oncopipe eventually!

modules/slms_3/1.0/slms_3.smk
```diff
@@ -0,0 +1,591 @@
...
+rule _slms_3_input_strelka_vcf: 
+    input:
+        vcf = str(rules._strelka_output_filtered_vcf.output.vcf), 
+        tbi = str(rules._strelka_output_filtered_vcf.output.vcf_tbi)
+    output:
+        vcf = CFG_SLMS3["dirs"]["inputs"] + "vcf/{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}.strelka.combined.vcf.gz", 
+        tbi = CFG_SLMS3["dirs"]["inputs"] + "vcf/{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}.strelka.combined.vcf.gz.tbi"
+    run:
+        op.relative_symlink(input.vcf, output.vcf)
```
rdmorin: I've been advocating for all the input symlinking to change to absolute_symlink

rdmorin (dismissed):
Looks great. I caught a few very minor things but mostly cosmetic. A few possible improvements noted but not necessary.

modules/mutect2/2.0/config/default.yaml
lkhilton: This is actually really a bug fix. I changed the keyword in the Snakefile to `mutect2_run` on an earlier version but didn't fix the config when I did that.

modules/mutect2/2.0/mutect2.smk
lkhilton: Good question! I will actually test this on the Kridel CRAM files before I merge this.

modules/mutect2/2.0/mutect2.smk
lkhilton: Will do! I straight up stole this from the Manta module, lol.

modules/slms_3/1.0/config/default.yaml
lkhilton: The skeleton config is largely to overwrite values that have `__UPDATE__` strings in the module config, and also to highlight the config values that users might want to update.

lkhilton: Mostly because the Snakefile is set up to load the default configs from each module.

modules/slms_3/1.0/slms_3.smk
lkhilton: Now that I'm thinking again about the reason I did this it might not be necessary. Will test and either remove this or add more detail to the comment.

modules/slms_3/1.0/slms_3.smk
lkhilton: Does the fact that this is symlinking the output of another module change your thoughts about this?

lkhilton (commented):
Thanks for the review! Will implement your suggestions ASAP.

modules/slms_3/1.0/config/default.yaml
rdmorin: Ah. Makes sense.

modules/slms_3/1.0/slms_3.smk
rdmorin: If it was just to avoid clobbering the CFG from another module I think I had a way around this that worked in the Sequenza or Battenberg modules.

modules/slms_3/1.0/slms_3.smk
rdmorin: I suppose it depends on how deep these symlinks go. Would there be any reason not to use in_module=True here?

modules/mutect2/2.0/mutect2.smk
lkhilton: Done!

modules/mutect2/2.0/mutect2.smk
lkhilton: No .crai symlink is needed, but I added the reference fasta to one of the other Mutect rules and it runs perfectly on the Kridel CRAM files now.

modules/slms_3/1.0/slms_3.smk
lkhilton: I think this was giving me grief in an earlier prototype when I tried to load the sub-modules before setting up this variable. It did turn out to be unnecessary to duplicate this function so it's gone now!

modules/slms_3/1.0/slms_3.smk
lkhilton: I made these all `in_module=True` because the directories are all set by the module and will be adjacent in the file system.

modules/mutect2/2.0/mutect2.smk
rdmorin: Thanks. Do you know how to get github to allow the changes to show up? Most of the time these resolutions don't appear in the diff.

modules/mutect2/2.0/mutect2.smk
rdmorin: Awesome

modules/mutect2/2.0/mutect2.smk
rdmorin: It's probably just my weak Python but what is the second parameter ("input") for and where does it come from?

modules/slms_3/1.0/slms_3.smk
```diff
@@ -0,0 +1,451 @@
...
+assert float(CFG_SLMS3["module_versions"]["mutect2"]) >= 2.0, (
+    f"The current mutect2 module version is {CFG_SLMS3['module_versions']['mutect2']}. "
+    "SLMS-3 requires mutect2 module version 2.0 or higher. "
+)
+
+assert float(CFG_SLMS3["module_versions"]["strelka"]) >= 1.1, (
+    f"The current strelka module version is {CFG_SLMS3['module_versions']['strelka']}. "
+    "SLMS-3 requires strelka module version 1.1 or higher. "
+)
+
```
rdmorin, line 68: I thought I commented on this topic but now I can't find the original comment. What are your thoughts on defaulting to the latest version of all these modules so the user (and developers) won't need to keep updating the config any time one of these modules gets a version bump? If it could be done automatically I imagine it would reduce the complexity a lot and simplify the config.

modules/starfish/2.0/config/default.yaml
```diff
@@ -10,12 +10,22 @@ lcr-modules:
         inputs:
             
             vcf: 
-                # Available wildcards: {seq_type} {genome_build} {tumour_id} {normal_id} {pair_status}
-                # Each key:value pair is the name of the variant caller and the path to its VCF file
-                # Each key should be unique in this set. 
-                # Note: If you aren't using strelka as one of the inputs to this, you will have to either delete the
-                # following line or copy the entire config to your project and omit this line from your project config. 
-                strelka: "__UPDATE__"
+
```
rdmorin, line 13: I'm sure I had commented here too. Is this change to help ensure consistency in what order the tools get combined by starfish?

rdmorin (approved):
Do all the callers used by slms3 now work with crams?

lkhilton:
> Do all the callers used by slms3 now work with crams?

As far as I know, yes. Kostia has run SAGE-1.0 on all the Kridel CRAMs successfully, and you ran Manta, Strelka, and Lofreq.

modules/starfish/2.0/config/default.yaml
lkhilton: It is indeed. With Strelka as the first key here, it was always loaded first even if a second config was loaded afterwards that clobbered the UPDATE string with the VCFs in a different order.

modules/slms_3/1.0/slms_3.smk
lkhilton: My hesitation is that someone might have run an older version of one of the tools and wouldn't want to have to run it again, as I did for everything in icgc_dart which had been run through manta-2.2 instead of 2.3. This is a situation where I thin unintentional updates to the config could have far-reaching consequences for how many things get re-run.

modules/mutect2/2.0/mutect2.smk
lkhilton: Oh it took me a while to figure this out! `input` is the Snakemake object `input`. So when this function is called by either of the `_mutect2_run` rules, it passes the `wildcards` and `input` objects to this function so that it can parse the `input.candidate_positions` into the command line argument. 

Basically this exists because you can't have a blank input object. So I can't just have `input.candidate_positions` blank in the config and pass it to the shell command, I have to have a test for whether or not the `candidate_positions` file is specified in the config, and then have a dummy VCF in its place in the inputs for the rule, though that dummy VCF doesn't get added to a command-line arg.

lkhilton:
@rdmorin I've added this module to the demo, could you give one last review/approval?

rdmorin (approved):
Thanks for adding this to the demo. That should help demystify it a bit. Somehow I forgot to ask this time around!

modules/slms_3/1.0/slms_3.smk
rdmorin: That sounds like a very good reason to keep it as-is.

PR #163, opened 2021-01-17
Module/sage/1.0

Kdreval:
It was found that mapping quality of not 0 for unmapped reads causes SAGE to exit with an error. This [was addressed](https://github.com/hartwigmedical/hmftools/issues/65) by developers in version 2.6, where flag `-validation_stringency` was added as an optional argument for SAGE. I am updating SAGE version used by this module to the latest release, so validation stringency can be set through the module config file.
In addition, it was found that if the user runs a dry run of lcr-modules right after installation, the function I used to restrict for standard chromosomes causes an error since it doesn't know that the file will be generated by snakemake. In this version, I modified the function to avoid this bug. 
Lastly, the symlinks for input files are now absolute instead of being relative.

### DIFF
--- a/envs/hmftools/hmftools-sage-2.6.yaml
+++ b/envs/hmftools/hmftools-sage-2.6.yaml
@@ -30,7 +30,7 @@ dependencies:
   - gxx_impl_linux-64=7.3.0
   - gxx_linux-64=7.3.0
   - harfbuzz=2.4.0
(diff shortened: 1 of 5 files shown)


rdmorin (approved):
Looks good. Thanks for getting SAGE updated and for the quick bug fixes.

Kdreval:
Thank you!

PR #164, opened 2021-01-17
Sequenza with cram support and Blacklist filtering

Kdreval:
This update for the existing sequenza module includes support of cram files, filtering of variants that fall within blacklisted regions, and harmonizing the module with recent updates (oncopipe version check, absolute symlinks for inputs, `in_module` symlinking for outputs, and resource unpacking). I have also removed the default scratch directory from config.
I have edited v1.4 files to ensure easy view of the changes that I implemented. Once approved, I will bump this to version 1.5 and restore v1.4 back.

### DIFF
--- a/modules/sequenza/1.4/src/bash/filter_seqz.sh
+++ b/modules/sequenza/1.4/src/bash/filter_seqz.sh
@@ -18,10 +18,18 @@ set -euf -o pipefail
 zcat "${SEQZ_FILE}" \
 	| egrep -v "^chromosome" \
 	| awk 'BEGIN {FS="\t"} $9 == "het" {print $1 ":" $2}' \
-	| sort -S "${BUFFER_SIZE}" --parallel 10 \
+	| sort -S "${BUFFER_SIZE}" --parallel 8 \
 	| comm - "${DBSNP_POS_FILE}" -2 -3 \
 	| tr ":" "\t" \
 	> "${RARE_VARIANTS}"
+
+if [[ $SEQZ_BLACKLIST_BED_FILES != "" ]]; then
+	zcat "${SEQZ_FILE}" \
+		| awk 'BEGIN {FS=OFS="\t"} NR > 1 && $9 == "het" {print $1, $2-1, $2}' \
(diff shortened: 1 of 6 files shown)


lkhilton (dismissed):
Looks great. Does this really need a version bump though? I don't think I see anything that would break existing sequenza setups except for the way resources are unpacked.

Kdreval:
Thank you! The only new feature that is added here is filtering towards the encode blacklist. It will not break the existing outputs and pipeline. I am thinking to do version bump so we can distinguish whether it was filtered only towards dbsnp, or blacklist ad well.

rdmorin:
I agree with Laura that a version bump may not be warranted. If the rationale is to increase it so we can distinguish it in GAMBL then why not just change the config and update the parent directory to explicitly refer to the change in filtering?

Kdreval:
Thank you all for the comments! I agree that the update is minor and a version update is not required. I only changed the CHANGELOG to keep the record of added features, but associate it with v1.4. This dismissed the approval, so will need another one.

modules/sequenza/1.4/src/bash/filter_seqz.sh
```diff
@@ -18,11 +18,19 @@ set -euf -o pipefail
 zcat "${SEQZ_FILE}" \
 	| egrep -v "^chromosome" \
 	| awk 'BEGIN {FS="\t"} $9 == "het" {print $1 ":" $2}' \
-	| sort -S "${BUFFER_SIZE}" --parallel 10 \
+	| sort -S "${BUFFER_SIZE}" --parallel 8 \
```
rdmorin, line 21: In theory the 8 should be a parameter that is passed in from Snakemake (ideally)

rdmorin (approved):
Up to you if you want to change that script. I can live with this version since that issue has been there from the start.

rdmorin:
Should this be merged, @Kdreval ?

Kdreval:
Yes. I thought to reorganize it to remove checkpoint, but this can be done later for future versions

Issue #166, opened 2021-01-18
Rules should not explicitly refer to scripts as inputs 


rdmorin:
The setting of scripts used by specific rules as explicit inputs to that rule is problematic. When a second user is using a different copy of the lcr-modules repo the pipeline will then force re-running of the rules that depend on that input file. For example, if the following rule is run by user B (let's call her Laura... no, Dr. H) on a partially completed run of the pipeline (run by user A, or Dr. M), it will trigger a complete rerun of the pipeline by user A. I see no reason a script should ever be an input to a rule. These should be changed to params or, possibly better yet, just put into "src" and used by exporting that directory to the PATH at the start of the rule. 

```
rule _manta_augment_vcf:
    input:
        variants_dir = str(rules._manta_run.output.variants_dir),
        aug_vcf = CFG["inputs"]["augment_manta_vcf"]
```

PR #167, opened 2021-01-18
Modules/slms 3/1.0

lkhilton:
This PR fixes a problem with Mutect2 VCF sample name replacement that occurred when the sample ID pulled from the BAM header contains a space (shockingly not uncommon). It also incorporates the full `all` input from each sub-module to ensure all the final `99-outputs` directories are properly populated. 

I've also rolled in a patch to all the Manta modules (version 2.0 or higher) that moves the `augment_manta_vcf.py` script from an input to a parameter to prevent re-running if the lcr-scripts directory moves or changes (see #166).

### DIFF
--- a/modules/manta/2.0/config/default.yaml
+++ b/modules/manta/2.0/config/default.yaml
@@ -6,6 +6,8 @@ lcr-modules:
             # Available wildcards: {seq_type} {genome_build} {sample_id}
             sample_bam: null  # UPDATE
             sample_bai: null  # UPDATE
+
+        scripts:
             augment_manta_vcf: "{SCRIPTSDIR}/augment_manta_vcf/1.0/augment_manta_vcf.py"
 
         options:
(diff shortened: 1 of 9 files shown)


rdmorin:
Wow. How does someone even put a space into the sample ID field in the bam header?

rdmorin (changes requested):
I don't see updates to the default.yaml config file for versions 2.1 and 2.2. Can you confirm that you've applied this bug fix retroactively and to both the config and smk file for each?

lkhilton:
I can see it when I look at the files changed section...

rdmorin:
I see them now. I guess you used version 1.0 and 1.1 of the script consistently with what version was used in the Manta version?

lkhilton:
Right, I didn't change anything except how the script is called by the Snakemake rule and where it's specified in the config.

rdmorin:
Seeing as there's active development of Manta on another branch, could you also drop in the changes Lakshay has made to this module (just in version 2.3) so he doesn't get another conflict later? You should be able to paste them right from his PR. I'm specifically thinking of the oncopipe version check and updates to symlinking.

lkhilton:
I think this is going to result in a conflict for him anyway, but it will be trivial to fix. I can give him a heads up about it.

rdmorin:
It will auto-merge when it can. I just thought if you add in his changes the auto-merge might happen. It's hard to know

rdmorin (approved):

PR #171, opened 2021-03-02
Battenberg module updates

Lakshay-sethi:
# Pull Request Checklists

Changes In Battenberg module to allow for downloads of reference files automatically.

### Required

  ```bash
  ```

### If applicable

Tested In GSC
  - [x] Genome
  - [x] Capture


modules/battenberg/1.1/battenberg.smk
```diff
@@ -50,11 +50,48 @@ _battenberg_CFG = CFG
 
 # Define rules to be run locally when using a compute cluster
 localrules:
+    _battenberg_get_refrence
```
oncogenomics: Should this be `_battenburg_get_reference`?

modules/battenberg/1.1/battenberg.smk
```diff
@@ -50,11 +50,48 @@ _battenberg_CFG = CFG
...
     _battenberg_all
 
-
 ##### RULES #####
 
+# Downloads the refrence files into the module results directory (under '00-inputs/') from https://www.bcgsc.ca/downloads/morinlab/reference/ . 
+rule _battenberg_get_refrence:
```
oncogenomics: Typo, "refrence"

oncogenomics (commented):
Just a couple typos and then it looks good to me although you should have someone check it over for functionality as well.

modules/battenberg/1.1/battenberg.smk
Lakshay-sethi: I don't think so. I just copied the spelling of the word from from the file name

modules/battenberg/1.1/battenberg.smk
Lakshay-sethi: Taken care of

modules/battenberg/1.1/battenberg.smk
oncogenomics: If so it seems the filename has a typo too and should be changed. Unless "refrence" has a specific meaning here.

rdmorin: It's not the file name it's the rule name, right? Battenberg (not battenburg) is correct. "Refrence" is the typo but both the rule and the line indicated would need to change to be consistent or this will break.

modules/battenberg/1.1/battenberg.smk
```diff
@@ -50,11 +50,48 @@ _battenberg_CFG = CFG
...
 # Define rules to be run locally when using a compute cluster
 localrules:
+    _battenberg_get_reference
     _battenberg_all
 
-
 ##### RULES #####
 
+# Downloads the reference files into the module results directory (under '00-inputs/') from https://www.bcgsc.ca/downloads/morinlab/reference/ . 
+rule _battenberg_get_reference:
```
rdmorin: What will happen if a user provides genomes aligned to hs37d5? We need to handle all the genome builds covered by the reference workflow, not just the two we use in GAMBL.

modules/battenberg/1.1/battenberg.smk
```diff
@@ -145,8 +183,8 @@ rule _run_battenberg:
         sex=$(cut -f 4 {input.sex_result}| tail -n 1); 
         echo "setting sex as $sex";
         Rscript {params.script} -t {wildcards.tumour_id} 
-        -n {wildcards.normal_id} --tb {input.tumour_bam} --nb {input.normal_bam} -f {input.fasta}
-        -o {params.out_dir} --sex $sex --reference {params.reference_path} {params.chr_prefixed} --cpu {threads} >> {log.stdout} 2>> {log.stderr} &&  
+        -n {wildcards.normal_id} --tb {input.tumour_bam} --nb {input.normal_bam} -f {input.fasta} --ref {params.ref}
```
rdmorin: You seem to have dropped the chr_prefixed flag. How does the R script determine (or know) if the genome is chr prefixed?

modules/battenberg/1.1/config/default.yaml
```diff
@@ -10,18 +10,15 @@ lcr-modules:
 
         scratch_subdirectories: []
 
-        reference_path:
-            hg38: "__UPDATE__"
-            grch37: "__UPDATE__"
-
         options:
```
rdmorin: This is a feature of the genome builds. If those are auto-downloaded, then it doesn't need to (and shouldn't be) in the configs. You still need to handle it properly so the script works.

modules/battenberg/1.1/src/battenberg_wgs_hg38.R
```diff
@@ -75,18 +65,29 @@ MIN_BASE_QUAL = 20
...
+REPLICCORRECTPREFIX = paste0(REFERENCE_BASE, "/battenberg_wgs_replic_correction_1000g_v3/1000_genomes_replication_timing_chr_")
+G1000PREFIX = paste0(REFERENCE_BASE, "/battenberg_1000genomesloci2012_v3/1000genomesAlleles2012_chr")
+G1000PREFIX_AC = paste0(REFERENCE_BASE, "/battenberg_1000genomesloci2012_v3/1000genomesloci2012_chr")
+GCCORRECTPREFIX = paste0(REFERENCE_BASE, "/battenberg_wgs_gc_correction_1000g_v3/1000_genomes_GC_corr_chr_")
+IMPUTE_EXE = "impute2"  #install using conda
+
+# WGS specific static
+ALLELECOUNTER = "alleleCounter" #conda package that should have this: cancerit-allelecount
+PROBLEMLOCI = paste0(REFERENCE_BASE, "/probloci.txt.gz")
+
```
rdmorin: Can you move these lines back to their original location so it's easier to see what has been changed rather than moved?

modules/battenberg/1.1/config/default.yaml
```diff
@@ -10,18 +10,15 @@ lcr-modules:
...
             #update and add/remove these lines as needed for the reference genomes being used. 
             chr_prefixed_reference:
                 hg38: " --chr_prefixed_genome "
                 grch37: " "
 
         conda_envs:
-            battenberg: "{MODSDIR}/envs/battenberg-1.0.yaml"
+            battenberg: "{MODSDIR}/envs/battenberg-1.1.yaml"
```
rdmorin: Where is this file? I don't see it in the PR.

modules/battenberg/1.1/battenberg.smk
```diff
@@ -50,11 +50,48 @@ _battenberg_CFG = CFG
...
+        &&
+        wget -qO- {params.url}/battenberg_{params.build}_gc_correction.tar.gz  |
+        tar -xvz > {output.battenberg_gc_correction} -C {params.folder}
+        &&
+        wget -qO- {params.url}/battenberg_1000genomesloci_{params.build}.tar.gz | 
+        tar -xvz > {output.genomesloci} -C {params.folder}
+        &&
+        wget -O {output.impute_info} 'https://ora.ox.ac.uk/objects/uuid:2c1fec09-a504-49ab-9ce9-3f17bac531bc/download_file?file_format=plain&safe_filename=impute_info.txt&type_of_work=Dataset'
+        &&
+        python {params.PATH}/reference_correction.py {params.build}
```
rdmorin: Because this rule uses python it should use a conda environment with the appropriate python version.

modules/battenberg/1.1/src/reference_correction.py
```diff
@@ -0,0 +1,14 @@
+import os
```
rdmorin: This needs to be documented.

rdmorin (changes requested):

modules/battenberg/1.1/config/default.yaml
Lakshay-sethi: Already deleted and tested without it

modules/battenberg/1.1/src/battenberg_wgs_hg38.R
Lakshay-sethi: Done

modules/battenberg/1.1/config/default.yaml
Lakshay-sethi: The file was already present, I just linked it

modules/battenberg/1.1/battenberg.smk
Lakshay-sethi: Done, but please check and confirm (Not sure)

Kdreval (changes requested):

modules/battenberg/1.1/src/reference_correction.py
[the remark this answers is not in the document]
Lakshay-sethi: done

Lakshay-sethi: done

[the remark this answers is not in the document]
Kdreval: thank you for adding documentation. Still not clear how to use this script. What are the input arguments? What is the output? Can you also add example usage statement? You can find the example of custom script documentation in any lcr-script, for example https://github.com/LCR-BCCRC/lcr-scripts/blob/1827fcdad34996211e5725e17fc322b0940d9803/salmon2counts/1.0/salmon2counts.R#L1-L25

modules/battenberg/1.1/src/battenberg_wgs_hg38.R
```diff
@@ -16,17 +16,17 @@ option_list = list(
...
   make_option(c("--bp"), type="character", default=NULL, help="Optional two column file (chromosome and position) specifying prior breakpoints to be used during segmentation", metavar="character"),
-  make_option(c("--reference"), type="character", default=NULL, help="Path to reference files", metavar="character"),
   make_option(c("-f","--reference_fasta"), type="character", default=NULL, help="Path to indexed genome fasta file (needed for CRAM compatability)", metavar="character"),
   make_option(c("--chr_prefixed_genome"), type="logical", default=FALSE, action="store_true", help="Flag to specify if the genome has chr prefixes in chromosome names", metavar="character"),
-  make_option(c("--impute_log"), type="character", default="./", help="Full path for where to store impute logs. If blank, these will be written to the main output directory and cleared.")
+  make_option(c("--impute_log"), type="character", default="./", help="Full path for where to store impute logs. If blank, these will be written to the main output directory and cleared."),
+  make_option(c("--ref"), type="character", default=NULL, help="Path to reference file", metavar="character")
```
Kdreval: Is there any particular reason to rename the flag? The one you removed (`--reference`) seems to be same as the one you added (`--ref`), so why not to use the previous name?

Lakshay-sethi: done

modules/battenberg/1.1/src/reference_correction.py
```diff
@@ -0,0 +1,38 @@
+##### ATTRIBUTION #####
```
Kdreval: this file is not present in version 1.1 and needs to be removed

modules/battenberg/1.1/envs/wget-1.20.1.yaml
```diff
@@ -0,0 +1 @@
+../../../../envs/wget/wget-1.20.1.yaml
```
Kdreval: The version 1.1 should remain as is, with no changes

modules/battenberg/1.1/config/default.yaml
Kdreval: The version 1.1 should remain as is after you update it to 1.2, please make sure 1.1 corresponds to the one on master

modules/battenberg/1.1/src/reference_correction.py
Lakshay-sethi: done

modules/battenberg/1.2/battenberg.smk
```diff
@@ -0,0 +1,297 @@
...
+        op.absolute_symlink(input.bam, output.bam)
+        op.absolute_symlink(input.bam + ".bai", output.bai)
+        op.absolute_symlink(input.bam + ".bai", output.crai)
+
+# Installs the Battenberg R dependencies and associated software (impute2, alleleCounter)
+# Currently I think this rule has to be run twice for it to work properly because the conda environment is created here. 
+# I am open to suggestions for how to get around this.
+rule _install_battenberg:
+    output:
+        complete = "config/envs/battenberg_dependencies_installed.success"
```
Kdreval: can you change it to touch the file in the CFG["dirs"]["input"]instead of config/envs?

Lakshay-sethi: done

Kdreval: Thanks! Have you tried running it, or just modified the line? The `CFG["dirs"]["input"] + "/battenberg_dependencies_installed.success"` will create path with `//` in it, so it is best to avoid whenever possible. This file is also used as input in the subsequent rule, did you look at that?

Kdreval (approved):

modules/battenberg/1.2/src/reference_correction.py
```diff
@@ -1,3 +1,6 @@
+#!/usr/bin/python
```
rdmorin: Shouldn't this be 
#!/usr/bin/env python

rdmorin (changes requested):

modules/battenberg/1.2/src/reference_correction.py
Kdreval: Right, it should be `#!/usr/bin/env python`, I changed it in the last commit before approving here:
https://github.com/LCR-BCCRC/lcr-modules/pull/171/commits/381887835f60446f39d83747fb009201c9d53cba

rdmorin (approved):

PR #178, opened 2021-04-01
Module/ichorcna/1.0

Jwong684:
# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable

To be completed.


rdmorin (changes requested):
One big thing I'd like to raise is why the source code for ichorcna is seemingly bundled with this module. This should auto-install rather than containing all the source code.

modules/ichorcna/1.0/ichorcna.smk
```diff
@@ -0,0 +1,188 @@
...
+rule _ichorcna_input_bam:
+    input:
+        bam = CFG["inputs"]["sample_bam"],
+        bai = CFG["inputs"]["sample_bai"]
+    output:
+        bam = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{sample_id}.bam",
+        bai = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{sample_id}.bam.bai", # specific to readCounter
+        crai = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{sample_id}.bam.crai" 
+    run:
+        op.relative_symlink(input.bam, output.bam)
```
Kdreval: we started to use the `op.absolute_symlink` here instead of relative symlink. This will also need the check for oncopipe version, and you can find example in one of the recent modules (battenberg/1.1, pathseq/1.0 etc)

modules/ichorcna/1.0/ichorcna.smk
```diff
@@ -0,0 +1,188 @@
...
+        plot = str(rules._run_ichorcna.output.plot)
+    output:
+        corrDepth = CFG["dirs"]["outputs"] + "{seq_type}--{genome_build}/corrDepth/{binSize}/{tumour_id}--{normal_id}--{pair_status}.corrDepth.txt",
+        param = CFG["dirs"]["outputs"] + "{seq_type}--{genome_build}/param/{binSize}/{tumour_id}--{normal_id}--{pair_status}.param.txt",
+        cna = CFG["dirs"]["outputs"] + "{seq_type}--{genome_build}/binCNA/{binSize}/{tumour_id}--{normal_id}--{pair_status}.cna.seg",
+        segTxt = CFG["dirs"]["outputs"] + "{seq_type}--{genome_build}/seg_txt/{binSize}/{tumour_id}--{normal_id}--{pair_status}.seg.txt",
+        seg = CFG["dirs"]["outputs"] + "{seq_type}--{genome_build}/seg/{binSize}/{tumour_id}--{normal_id}--{pair_status}.seg",
+        plot = CFG["dirs"]["outputs"] + "{seq_type}--{genome_build}/plot/{binSize}/{tumour_id}--{normal_id}--{pair_status}_genomeWide.pdf"
+    run:
+        op.relative_symlink(input.corrDepth, output.corrDepth)
```
Kdreval: the latest version of oncopipe also supports argument `in_module=TRUE`, which creates "shallow" symlinks. You can add it here as well, and the recent modules have an example of this

modules/ichorcna/1.0/ichorcna.smk
```diff
@@ -0,0 +1,188 @@
...
+localrules:
+    _ichorcna_input_bam,
+    _ichorcna_output,
+    _ichorcna_all
+
+# ---------------------------------------------------------------------------- #
+##### RULES #####
+# ---------------------------------------------------------------------------- #
+
+CFG["runs"]["binSize"] = str(CFG["options"]["readcounter"]["binSize"])
```
Kdreval: what does this do? I understand that it basically tells rule all the binsize for each sample. If it is the same for all samples, you can in the rule all just use `str(CFG["options"]["readcounter"]["binSize"])` and then you can multiply itty the number of samples, for example `*len(CFG["runs"]["tumour_sample_id"]` so it expands the length for each sample. There is example in the rule all of  module liftover

demo/Snakefile
```diff
@@ -83,4 +84,5 @@ rule all:
         rules._strelka_all.input,
         rules._bwa_mem_all.input,
         rules._liftover_all.input,
-        rules._controlfreec_all.input
+        rules._controlfreec_all.input,
+        rules._ichorcna_all.input
```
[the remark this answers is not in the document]
Jwong684: Done

demo/config.yaml
```diff
@@ -77,5 +77,8 @@ lcr-modules:
         tool: "battenberg"
         inputs:
             sample_seg: "data/{tool}/hg38/{tumour_sample_id}--{normal_sample_id}_subclones.igv.seg"
-
-
+    
+    ichorcna:
+        inputs:
+            sample_bam: "data/{sample_id}.bam"
+            sample_bai: "data/{sample_id}.bam.bai"
```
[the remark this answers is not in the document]
Jwong684: Done

modules/ichorcna/1.0/config/default.yaml
```diff
@@ -0,0 +1,120 @@
...
+        options:
+            readcounter:
+                readCounterScript: "{MODSDIR}/src/readCounter"
+                chrs:
+                    hg19: "1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,X,Y"
+                    grch37: "1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,X,Y"
+                    hs37d5: "1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,X,Y"
+                    hg38: "chr1,chr2,chr3,chr4,chr5,chr6,chr7,chr8,chr9,chr10,chr11,chr12,chr13,chr14,chr15,chr16,chr17,chr18,chr19,chr20,chr21,chr22,chrX,chrY"
+                    grch38: "chr1,chr2,chr3,chr4,chr5,chr6,chr7,chr8,chr9,chr10,chr11,chr12,chr13,chr14,chr15,chr16,chr17,chr18,chr19,chr20,chr21,chr22,chrX,chrY"
+                qual: 20
```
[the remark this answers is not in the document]
Jwong684: Done

modules/ichorcna/1.0/config/default.yaml
```diff
@@ -0,0 +1,120 @@
...
+            run:
+                ichorCNA_libdir: "{MODSDIR}/src/"
+                ichorCNA_rscript:  "{MODSDIR}/src/runIchorCNA.R"
+                # use panel matching same bin size (optional)
+                ichorCNA_normalPanel: 
+                    "1000000": "{MODSDIR}/src/inst/extdata/HD_ULP_PoN_{genome_build}_1Mb_median_normAutosome_median.rds"
+                    "500000": "{MODSDIR}/src/inst/extdata/HD_ULP_PoN_{genome_build}_500kb_median_normAutosome_median.rds"
+                # must use gc wig file corresponding to same binSize (required)
+                ichorCNA_gcWig:
+                    "1000000": "{MODSDIR}/src/inst/extdata/gc_{genome_build}_1000kb.wig"
```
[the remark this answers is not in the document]
Jwong684: Yes, unfortunately ichorCNA's github repo is messy and inconsistent with their naming conventions. In my original version, I had to manually rename some of their reference files to fit this format. In the current version, there's one rule with a bunch of symlinks that renames the reference files so it would fit in downstream rules.

modules/ichorcna/1.0/config/default.yaml
```diff
@@ -0,0 +1,120 @@
...
+                ichorCNA_exons:  NULL
+                ichorCNA_centromere:  
+                    grch37: "{MODSDIR}/src/inst/extdata/GRCh37.p13_centromere_UCSC-gapTable.txt"
+                    hg19: "{MODSDIR}/src/inst/extdata/GRCh37.p13_centromere_UCSC-gapTable.txt"
+                    hs37d5: "{MODSDIR}/src/inst/extdata/GRCh37.p13_centromere_UCSC-gapTable.txt"
+                    grch38: "{MODSDIR}/src/inst/extdata/GRCh38.GCA_000001405.2_centromere_acen.txt"
+                    hg38: "{MODSDIR}/src/inst/extdata/GRCh38.GCA_000001405.2_centromere_acen.txt"
+                ichorCNA_minMapScore: 0.75
+                ichorCNA_chrs:  
+                    grch37: "c('1', '2', '3','4','5','6','7','8','9','10','11','12','13','14','15','16','17','18','19','20','21','22','X')"
```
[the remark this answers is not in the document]
Jwong684: Just input-syntax for two different programs (the first one is formatted for readCounter as part of hmmcopy_utils and this format is for ichorCNA (to be used in R))

modules/ichorcna/1.0/config/default.yaml
```diff
@@ -0,0 +1,120 @@
...
+                    hg19: "c('1', '2', '3','4','5','6','7','8','9','10','11','12','13','14','15','16','17','18','19','20','21','22','X')"
+                    hs37d5: "c('1', '2', '3','4','5','6','7','8','9','10','11','12','13','14','15','16','17','18','19','20','21','22','X')"
+                    grch38: "c('chr1', 'chr2', 'chr3','chr4','chr5','chr6','chr7','chr8','chr9','chr10','chr11','chr12','chr13','chr14','chr15','chr16','chr17','chr18','chr19','chr20','chr21','chr22','chrX')"
+                    hg38: "c('chr1', 'chr2', 'chr3','chr4','chr5','chr6','chr7','chr8','chr9','chr10','chr11','chr12','chr13','chr14','chr15','chr16','chr17','chr18','chr19','chr20','chr21','chr22','chrX')"
+                ichorCNA_fracReadsInChrYForMale: 0.002
```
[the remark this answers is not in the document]
Jwong684: Done

modules/ichorcna/1.0/ichorcna.smk
```diff
@@ -0,0 +1,188 @@
...
+
+import pandas as pd
+import numpy as np
+import oncopipe as op
+import glob
+import os
+
+### Directories ###
+
+resultsDir = "results/"
```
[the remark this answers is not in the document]
Jwong684: I left them in by accident when I was initially crafting the module. They're not used at all in the module. They've been removed now

modules/ichorcna/1.0/ichorcna.smk
```diff
@@ -0,0 +1,188 @@
...
+    _ichorcna_all
+
+# ---------------------------------------------------------------------------- #
+##### RULES #####
+# ---------------------------------------------------------------------------- #
+
+CFG["runs"]["binSize"] = str(CFG["options"]["readcounter"]["binSize"])
+
+# Symlinks the input files into the module results directory (under '00-inputs/')
+rule _ichorcna_input_bam:
```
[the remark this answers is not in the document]
Jwong684: No, unfortunately readCounter does not work on CRAMs. readCounter (part of bamtools) never updated for CRAMs, though it seemed like a popular idea (https://github.com/pezmaster31/bamtools/issues/149)

modules/ichorcna/1.0/ichorcna.smk
```diff
@@ -0,0 +1,188 @@
...
+            binSize=CFG["runs"]["binSize"])
+
+
+
+##### CLEANUP #####
+
+
+# Perform some clean-up tasks, including storing the module-specific
+# configuration on disk and deleting the `CFG` variable
+op.cleanup_module(CFG)
```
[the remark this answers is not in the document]
Jwong684: Done

modules/ichorcna/1.0/ichorcna.smk
```diff
@@ -0,0 +1,188 @@
...
+        # norm = CFG["dirs"]["readDepth"] + "{seq_type}--{genome_build}/{binSize}/{normal_id}.bin{binSize}.wig"
+    output:
+        corrDepth = CFG["dirs"]["seg"] + "{seq_type}--{genome_build}/{binSize}/{tumour_id}--{normal_id}--{pair_status}/{tumour_id}.correctedDepth.txt",
+        param = CFG["dirs"]["seg"] + "{seq_type}--{genome_build}/{binSize}/{tumour_id}--{normal_id}--{pair_status}/{tumour_id}.params.txt",
+        cna = CFG["dirs"]["seg"] + "{seq_type}--{genome_build}/{binSize}/{tumour_id}--{normal_id}--{pair_status}/{tumour_id}.cna.seg",
+        segTxt = CFG["dirs"]["seg"] + "{seq_type}--{genome_build}/{binSize}/{tumour_id}--{normal_id}--{pair_status}/{tumour_id}.seg.txt",
+        seg = CFG["dirs"]["seg"] + "{seq_type}--{genome_build}/{binSize}/{tumour_id}--{normal_id}--{pair_status}/{tumour_id}.seg",
+        plot = CFG["dirs"]["seg"] + "{seq_type}--{genome_build}/{binSize}/{tumour_id}--{normal_id}--{pair_status}/{tumour_id}/{tumour_id}_genomeWide.pdf",
+        #rdata = "results/ichorCNA/{sample_id}/{sample_id}.RData"
+    params:
```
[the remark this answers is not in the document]
Jwong684: Not sure I follow here

modules/ichorcna/1.0/ichorcna.smk
Jwong684: Changed bam inputs to op.absolute_symlink

modules/ichorcna/1.0/ichorcna.smk
Jwong684: Added for all output relative_symlinks

modules/ichorcna/1.0/ichorcna.smk
Jwong684: Your interpretation is correct - I modified it to your method from liftOver.

Jwong684:
Modified version has been tested - it runs to completion for a grch37, hg19, hg38 sample.

Since I had reference files I needed to link and modify from the github repo of ichorCNA, I decided to get snakemake to clone their repo into the CFG["dirs"]["inputs"] directory so I could access these files in downstream rules.

The naming convention for ichorCNA's reference files is messy (ex. HD_ULP_PoN_1Mb_median_normAutosome_**mapScoreFiltered**_median.rds for hg19, and HD_ULP_PoN_**hg38**_1Mb_median_normAutosome_median.rds for hg38). I originally just renamed these files when I had them as part of the module, and symlinked reference files for related genome_builds (ex. hs37d5, hg19, grch37 would all have symlinks to the same reference file). I incorporated this symlinking step inside the module.

Also added Chris's suggestion of tweaking the default parameters.

modules/ichorcna/1.0/src/runIchorCNA.R
```diff
@@ -0,0 +1,423 @@
+# file:   ichorCNA.R
```
Kdreval, line 1: This file is now downloaded directly, right?

modules/ichorcna/1.0/config/default.yaml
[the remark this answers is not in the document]
Kdreval: Can this be moved to snakelike, instead of being in config? There, you can use file listing chromosomes generated by reference files (main_chromosomes.txt) or use function to generate chromosome names (there is example in sage module)

modules/ichorcna/1.0/config/default.yaml
```diff
@@ -0,0 +1,119 @@
...
+                    hg19: "1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,X,Y"
+                    grch37: "1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,X,Y"
+                    hs37d5: "1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,X,Y"
+                    hg38: "chr1,chr2,chr3,chr4,chr5,chr6,chr7,chr8,chr9,chr10,chr11,chr12,chr13,chr14,chr15,chr16,chr17,chr18,chr19,chr20,chr21,chr22,chrX,chrY"
+                    grch38: "chr1,chr2,chr3,chr4,chr5,chr6,chr7,chr8,chr9,chr10,chr11,chr12,chr13,chr14,chr15,chr16,chr17,chr18,chr19,chr20,chr21,chr22,chrX,chrY"
+                qual: 20 # only includes reads with mapping quality greater than 20
+                binSize:  1000000 # set window size to compute coverage 
+                # available binSizes are: 1000000, 500000, 50000, 10000
+            run:
+                ichorCNA_libdir: ""
```
Kdreval, line 19: Are all these files in inst are being auto downloaded? They are probably then not needed in the config since you set the naming convention and the path to these. files in the snakelike rule

modules/ichorcna/1.0/src/runIchorCNA.R
Jwong684: No this one is the one that I had to modify to allow hs37d5 to be included. They hard-coded the available genomes

modules/ichorcna/1.0/config/default.yaml
Jwong684: For the chromosomes, I could use the function for the readCounter step since I need the command to look like: 
 "1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,X,Y"
but for the ichorCNA step, it needs to be in R-vector format:
"c('1', '2', '3','4','5','6','7','8','9','10','11','12','13','14','15','16','17','18','19','20','21','22','X')"
Wouldn't it be simpler to keep it like this in the config? Otherwise, I'd need to make a function to include c( ), and ' ' 

'ichorCNA_libdir:' parameter is supposed to be set to where the github repo resides for ichorCNA (it'll use this to search for inst/extdata/ underneath that directory). I just left it in the config since it might be useful if someone needs to modify the path one day

rdmorin: How is that vector being given to ichorCNA?

modules/ichorcna/1.0/config/default.yaml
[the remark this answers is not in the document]
rdmorin: Good suggestion. I agree it's better to embed anything that generates code (python, R etc) into the snakefile

modules/ichorcna/1.0/config/default.yaml
Jwong684: In the command it would be --chrs \"c('1', '2', '3','4','5','6','7','8','9','10','11','12','13','14','15','16','17','18','19','20','21','22','X')\"

rdmorin: You could construct that in Python probably and then just add that as a "param" in your rule.

Jwong684: Done

modules/ichorcna/1.0/ichorcna.smk
```diff
@@ -242,7 +256,7 @@ rule _run_ichorcna:
         scStates = CFG["options"]["run"]["ichorCNA_scStates"],
         maxCN = CFG["options"]["run"]["ichorCNA_maxCN"],
         includeHOMD = CFG["options"]["run"]["ichorCNA_includeHOMD"],
-        chrs = op.switch_on_wildcard("genome_build", CFG["options"]["run"]["ichorCNA_chrs"]),
+        chrs = get_chromosomes_R,
```
rdmorin, line 259: Does this work?

Jwong684: Yeah it prints out:  `--chrs "c('chr1','chr2','chr3','chr4','chr5','chr6','chr7','chr8','chr9','chr10','chr11','chr12','chr13','chr14','chr15','chr16','chr17','chr18','chr19','chr20','chr21','chr22','chrX','chrY')" ` in dry run

rdmorin (approved):

PR #181, opened 2021-05-01
Lofreq improvements

rdmorin:
# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable

To be completed.


modules/lofreq/1.0/src/lofreq2_call_pparallel.py
```diff
@@ -459,7 +459,7 @@ def main():
     if idx != -1:
         final_vcf_out = lofreq_call_args[idx+1]
         if os.path.exists(final_vcf_out):
-            LOG.fatal("Cowardly refusing to overwrite already existing"
+            LOG.fatal("On line 462 Cowardly refusing to overwrite already existing"
```
Kdreval: very nice!!

modules/sequenza/1.4/src/samtools.py
```diff
@@ -0,0 +1,112 @@
+import shlex
```
Kdreval: Where is this script used? I don't think it is referenced in the Lofreq.smk or module config

rdmorin: Oops. This shouldn't be here.

rdmorin: I'll remove for final PR once I move everything to the new version directory

modules/lofreq/1.0/lofreq.smk
```diff
@@ -101,22 +113,76 @@ rule _lofreq_run:
...
+        opts = CFG["options"]["lofreq"]
+    conda:
+        CFG["conda_envs"]["lofreq"]
+    threads:
+        CFG["threads"]["lofreq"]
+    resources:
+        **CFG["resources"]["lofreq"]
+    shell:
+        op.as_one_line("""
+        ln -f {input.vcf_relaxed} {output.out_dir};
```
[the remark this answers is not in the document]
rdmorin: On purpose, yes. It effectively makes a redundant copy of the file so other parallel processes can use it and it only gets completely deleted once the last job is done.

modules/lofreq/1.0/lofreq.smk
```diff
@@ -101,22 +113,76 @@ rule _lofreq_run:
...
-            &&
-            lofreq somatic {params.opts} --threads {threads} -t {input.tumour_bam} -n {input.normal_bam}
-            -f {input.fasta} -o $(dirname {output.vcf_snvs_filtered})/ -d {input.dbsnp} {params.regions} --bed {input.bed}
-            > {log.stdout} 2> {log.stderr}
-            &&
-            rm {params.rm_files};
+            touch {output.preprocessing_start}
+            && 
+            lofreq somatic --normal_only {params.opts} --threads {threads} -t {input.normal_bam} -n {input.normal_bam}
+            -f {input.fasta} -o {output.out_dir}/ -d {input.dbsnp} --bed {input.bed}
```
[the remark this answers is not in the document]
rdmorin: Doesn't really matter if it does. The directory won't actually have that second slash.

Kdreval (dismissed):

Kdreval (approved):

PR #184, opened 2021-05-07
Module/gatk rnaseq/1.0

Jwong684:
# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable

To be completed.


demo/config.yaml
```diff
@@ -171,3 +171,9 @@ lcr-modules:
         # include here any additional flags to modify default parameters
         options:
             sage_run: ""
+
+
+    gatk_rnaseq:
+        inputs:
+            sample_bam: "data/{sample_id}.bam"
+            sample_bai: "data/{sample_id}.bam.bai"
```
Kdreval: This needs new line added at the end

modules/gatk_rnaseq/1.0/config/default.yaml
```diff
@@ -0,0 +1,66 @@
...
+        scratch_subdirectories: []
+
+        options:
+            gatk_variant_calling:
+                min_conf_thres: 20.0
+            gatk_variant_filtration:
+                window: 35 # window size between SNPs in cluster
+                cluster_size: 3 # at least 3 SNPs in cluster
+                # hard filtering (filters OUT) based on metrics: 
+                    # FS (FisherStrand): Phred-scale probability that there is a strand bias from a Fisher's test. (default FS > 30.0)
```
Kdreval, line 29: Thanks for detailed documentation!

modules/gatk_rnaseq/1.0/config/default.yaml
```diff
@@ -0,0 +1,66 @@
...
+                # For example, to remove all variants with -log10(POPAF) > 4.0: 
+                #"-f '.,PASS' -i 'INFO/POPAF > 4'"
+
+
+        conda_envs:
+            gatk_rnaseq: "{MODSDIR}/envs/gatk_rnaseq.yaml"
+            bcftools: "{MODSDIR}/envs/bcftools-1.10.2.yaml"
+            
+        threads:
+            gatk_splitntrim: 12
```
Kdreval, line 46: It is possible to combine these keys and reuse them across rules. In other words, if several rules need the same number of threads, they can refer to the same key in config. Same can be applied to resources as well. I think this reduces the number of keys to specify/adjust if needed, and reduces complexity of the config. What do you think?

modules/gatk_rnaseq/1.0/envs/bcftools-1.10.2.yaml
```diff
@@ -0,0 +1,36 @@
+name: test-bcftools
```
Kdreval: This (and another environment file) should be rather in lcr-modules/envs/ folder, and symlinked in the module. This will make it easier to reuse same environments across modules, therefore reducing the need to build multiple environments. There is already an environment for bcftools with the same version, so you can just as well symlink `lcr-modules/envs/bcftools/bcftools-1.10.2.yaml` to this module

modules/gatk_rnaseq/1.0/envs/gatk_rnaseq.yaml
```diff
@@ -0,0 +1,352 @@
+name: bioinformatics
```
Kdreval: Is this environment just GATK or there is something else besides it that is required for this module to run? There is a GATK environment `lcr-modules/envs/gatk/gatk-4.1.8.1.yaml`, maybe it can be reused here? I see this one contains tools like, for example, bedtools, vcf2maf, tmux, and a lot of perl dependencies, but are they used in the module?

modules/gatk_rnaseq/1.0/gatk_rnaseq.smk
```diff
@@ -0,0 +1,319 @@
...
+    _gatk_rnaseq_output_vcf,
+    _gatk_rnaseq_all,
+
+
+##### RULES #####
+
+
+# Symlinks the input files into the module results directory (under '00-inputs/')
+rule _gatk_rnaseq_input_bam:
+    input:
```
Kdreval, line 58: Do you know if it works on cram files? There are some RNASeq crams that can be used for compatibility testing, for example DLBCL cell lines

modules/gatk_rnaseq/1.0/gatk_rnaseq.smk
```diff
@@ -0,0 +1,319 @@
...
+        bam = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{sample_id}.bam",
+        fasta = reference_files("genomes/{genome_build}/genome_fasta/genome.fa")
+    output:
+        bam = temp(CFG["dirs"]["gatk_splitntrim"] +  "bam/{seq_type}--{genome_build}--{pair_status}/{sample_id}.split_reassign_mq.bam")
+    log:
+        stdout = CFG["logs"]["gatk_splitntrim"] + "{seq_type}--{genome_build}--{pair_status}/{sample_id}.gatk_splitntrim.stdout.log",
+        stderr = CFG["logs"]["gatk_splitntrim"] + "{seq_type}--{genome_build}--{pair_status}/{sample_id}.gatk_splitntrim.stderr.log"
+    params:
+        mem_mb = lambda wildcards, resources: int(resources.mem_mb / 1000)
```
Kdreval: For tools running java it is sometimes useful to include heat buffer, otherwise they might jump to excessive memory usage and will get killed by slurm. I mean the same approach as you did for the rule that merges vcf files. Also, here you can give -Xmx memory in `m`, so this calculation may not be necessary. Same can be applied to the rules below as well

modules/gatk_rnaseq/1.0/gatk_rnaseq.smk
```diff
@@ -0,0 +1,319 @@
...
+        gnomad = reference_files("genomes/{genome_build}/variation/af-only-gnomad.{genome_build}.vcf.gz"),
+        mem_mb = lambda wildcards, resources: int(resources.mem_mb / 1000)
+    conda:
+        CFG["conda_envs"]["gatk_rnaseq"]
+    threads: CFG["threads"]["gatk_base_recalibration"]
+    resources:
+        **CFG["resources"]["gatk_base_recalibration"]
+    shell:
+        op.as_one_line("""
+        gatk --java-options "-Xmx{params.mem_mb}G" BaseRecalibrator
```
Kdreval: I wonder if you know whether any rule is using java garbage collector, which may by default use all available threads? If so, it may create a lot of competitive usage when launched for many samples simultaneously. One way to get around this is to provide `-XX:ConcGCThreads=1` option that basically limits java to only 1 thread for garbage collector. Please ignore this comment if there is no issue with garbage collector!

modules/gatk_rnaseq/1.0/gatk_rnaseq.smk
```diff
@@ -0,0 +1,319 @@
...
+    shell:
+        op.as_one_line("""
+        gatk --java-options "-Xmx{params.mem_mb}G" SplitNCigarReads -fixNDN TRUE -RF GoodCigarReadFilter
+        -R {input.fasta} -I {input.bam} -O {output.bam}
+        > {log.stdout} 2> {log.stderr}
+        """)
+
+rule _gatk_base_recalibration:
+    input:
+        bam = rules._gatk_splitntrim.output,
```
Kdreval: can you please wrap the inputs referring to the outputs of other rules with `str()`?

modules/gatk_rnaseq/1.0/gatk_rnaseq.smk
```diff
@@ -0,0 +1,319 @@
...
+        op.relative_symlink(input.chrs, output.chrs)
+        
+        
+rule _gatk_variant_calling:
+    input:
+        bam = rules._gatk_applybqsr.output.bam,
+        bai = rules._gatk_applybqsr.output.bai,
+        fasta = reference_files("genomes/{genome_build}/genome_fasta/genome.fa")
+    output:
+        vcf = temp(CFG["dirs"]["gatk_variant_calling"] + "{seq_type}--{genome_build}--{pair_status}/{sample_id}.{chrom}.vcf.gz"),
```
Kdreval: Does HaplotypeCaller and VariantFiltration compress vcf files by default, or do you need to do that as a separate step? I was having impression that majority of variant calling tools do not compress by default, but may be very wrong!

modules/gatk_rnaseq/1.0/gatk_rnaseq.smk
```diff
@@ -0,0 +1,319 @@
...
+    run:
+        op.absolute_symlink(input.bam, output.bam)
+        op.absolute_symlink(input.bai, output.bai)
+
+rule _gatk_splitntrim:
+    input:
+        bam = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{sample_id}.bam",
+        fasta = reference_files("genomes/{genome_build}/genome_fasta/genome.fa")
+    output:
+        bam = temp(CFG["dirs"]["gatk_splitntrim"] +  "bam/{seq_type}--{genome_build}--{pair_status}/{sample_id}.split_reassign_mq.bam")
```
Kdreval: What do you think about making it `seq_type}--{genome_build}/{sample_id}--{pair_status}`? This would make it consistent eith the directory structure of other modules

Kdreval (commented):
Thanks Jasper! I left some comments. I have another general question: there is `{pair_status}` wildcard, but no reference to normal samples in the module. RNASeq would be unpaired, so do we need this wildcard to be included?

demo/config.yaml
Jwong684: Done

modules/gatk_rnaseq/1.0/envs/bcftools-1.10.2.yaml
Jwong684: Done

modules/gatk_rnaseq/1.0/envs/gatk_rnaseq.yaml
Jwong684: Hm that is a good question. I pulled it from a working environment. I'll test to see if it'll work with the gatk module

modules/gatk_rnaseq/1.0/gatk_rnaseq.smk
Jwong684: Will test

modules/gatk_rnaseq/1.0/gatk_rnaseq.smk
Jwong684: Done

modules/gatk_rnaseq/1.0/gatk_rnaseq.smk
Jwong684: Yes, the output files are gzipped (tested with: `file results/gatk_rnaseq-1.0/06-gatk_variant_filtration/mrna--grch37--no_normal/Toledo.filtered.vcf.gz`): output:  gzip compressed data, extra field

modules/gatk_rnaseq/1.0/gatk_rnaseq.smk
Jwong684: I prefer making the final file name simple like: Toledo.ouput.filt.vcf.gz.

You're right about no-normal pair status. Pair status just comes out as "no_normal" in the output directories. I'm not sure if it'd be informative or not to keep it in, but I can remove it altogether too

modules/gatk_rnaseq/1.0/gatk_rnaseq.smk
Jwong684: Hm I'm not sure about any rules specifically. I do see GATK documentation about some functions sometimes using too many resources. I will add `-XX:ConcGCThreads=1` as a pre-emptive measure

modules/gatk_rnaseq/1.0/gatk_rnaseq.smk
Jwong684: Changed all -Xmx {params.mem_mb}G to -Xmx {params.mem_mb}m with same calculation (0.8 * assigned mem_mb)

modules/gatk_rnaseq/1.0/gatk_rnaseq.smk
Jwong684: It is CRAM compatible, as-is

modules/gatk_rnaseq/1.0/envs/gatk_rnaseq.yaml
Jwong684: Yes it worked with gatk module. I will just use the symlinked version for both gatk and bcftools

Jwong684:
Removed pair-status as a wildcard

modules/gatk_rnaseq/1.0/config/default.yaml
Kdreval: You are right, let's keep the more detailed and informative names

demo/config.yaml
Kdreval: It still shows no new line here, maybe I am looking at outdated file?

Jwong684: Sorry, I must have missed it. It's fixed

Kdreval (approved):

Kdreval:
@Jwong684 is this ready to go?

Jwong684:
This should be completed - I have some output tested on the BL/DLBCL cell lines in GAMBL from way before. I want to add a grouping resource so it doesn't get clogged up with temp bams in the middle of the pipeline

Kdreval:
Great, thanks! It shows some merge conflicts with the recent update - can you resolve them and we'll get this on master?

Kdreval:
@Jwong684  what should we do with this? Is this still in works? Should we close this PR? Any plans on continuing this?

Jwong684:
It is functional but not that great. The final vcf becomes way too big for this to be implemented on a large GAMBL-wide scale. There are just too many SNVs that are called with RNA-seq.

I don't have a particular interest in variants in RNAseq at the moment, but it might be useful as a verification tool. It does capture many of the same SNPs as you would see in WES in the same sample.

PR #185, opened 2021-05-12
bug fixes galore

rdmorin:
Bug fixes are detailed in the changelog. This version should work for Schmitz exomes but the handling of custom bed file needs to be set up to work properly without breaking the default main_chromosomes.bed used for genomes.

### DIFF
--- a/modules/lofreq/1.1/lofreq.smk
+++ b/modules/lofreq/1.1/lofreq.smk
@@ -46,13 +46,6 @@ SCRIPT_PATH = CFG['inputs']['src_dir']
 #obtain default bed and update config
 bed = str(reference_files("genomes/{genome_build}/genome_fasta/main_chromosomes.bed"))
 CFG['switches']['regions_bed']['_default'] = bed
-print(CFG['switches']['regions_bed']['_default'])
-
-sample_ids = list(CFG['samples']['sample_id'])
-unmatched_normal_ids = list(config["lcr-modules"]["_shared"]["unmatched_normal_ids"].values())
-
-all_other_ids = list(set(sample_ids) - set(unmatched_normal_ids))
-
 
 
 
--- a/modules/lofreq/1.1/lofreq.smk
+++ b/modules/lofreq/1.1/lofreq.smk
@@ -43,6 +43,10 @@ CFG = op.setup_module(
 SCRIPT_PATH = CFG['inputs']['src_dir']
 #this is used in place of the shell.prefix() because that was not working consistently. This is not ideal. 
 
+#obtain default bed and update config
+bed = str(reference_files("genomes/{genome_build}/genome_fasta/main_chromosomes.bed"))
+CFG['switches']['regions_bed']['_default'] = bed
+print(CFG['switches']['regions_bed']['_default'])
 
 sample_ids = list(CFG['samples']['sample_id'])
 unmatched_normal_ids = list(config["lcr-modules"]["_shared"]["unmatched_normal_ids"].values())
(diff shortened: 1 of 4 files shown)


modules/lofreq/1.1/lofreq.smk
```diff
@@ -122,25 +124,50 @@ rule _lofreq_preprocess_normal_unmatched:
...
-        vcf_snvs_stringent = rules._lofreq_preprocess_normal_unmatched.output.vcf_snvs_stringent,
-        normal_relaxed_log = rules._lofreq_preprocess_normal_unmatched.output.normal_relaxed_log,
-        preprocessing_complete = rules._lofreq_preprocess_normal_unmatched.output.preprocessing_complete,
+        preprocessing_complete = rules._lofreq_link_to_preprocessed.output.preprocessing_complete,
         tumour_bam = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{tumour_id}.bam",
         normal_bam = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{normal_id}.bam",
+        vcf_relaxed = rules._lofreq_link_to_preprocessed.output.vcf_relaxed,
         fasta = reference_files("genomes/{genome_build}/genome_fasta/genome.fa"),
         dbsnp = reference_files("genomes/{genome_build}/variation/dbsnp.common_all-151.vcf.gz"), #in our experience, this filter doesn't remove as many SNPs as one would expect
         bed = reference_files("genomes/{genome_build}/genome_fasta/main_chromosomes.bed")
```
lkhilton: I think we could make this configurable to use exome beds with a single line of code like this at the top of the module Snakefile:  
```
CFG['switches']['regions_bed']['_default'] = reference_files("genomes/{genome_build}/genome_fasta/main_chromosomes.bed") 
```
Then in the lofreq_run rule(s): 

```
rule _lofreq_run: 
    input: 
        ...
        bed = op.switch_on_wildcard("seq_type", CFG["switches"]["regions_bed"])
```

If no bed files are specified under additional keys in the `regions_bed` part of the config, it will use the default reference_workflow chromosome bed.

rdmorin: I had spotted a line of code that looked almost identical to this (and commented out) in one of the edited snakefiles Jasper was using. What would the corresponding section of the config look like for this to use other bed files?

lkhilton: It currently looks like this:
```
regions_bed:
                _default: ""
                # Set to "" if you have no capture data
                capture: "-l __UPDATE__"
```
But you can't supply a second bed file argument to lofreq (it ignores the second one).

lkhilton: In this update it would change only slightly to: 
```
regions_bed:
                _default: ""
                # Set to "" if you have no capture data
                capture: ""
```

rdmorin: Should we completely remove this part? If I understand correctly, it doesn't matter what the user puts there it would get immediately clobbered anyway, right? Or does it need to be there so that it can be replaced later? 
 _default: ""

lkhilton: Yes it could be left blank, but it makes the code to put the default bed file into the config more complicated because the key is missing. I think it would be reasonable to leave it in the default config with a comment telling the user what gets used there and that any custom BED files must be specified with a seq_type wildcard.

lkhilton (approved):

PR #187, opened 2021-08-19
Modified reference workflow and SLMS-3 to use capture space

ckrushton:
- Updated Strelka, MuTect2, LoFreq, and SAGE to automatically use a capture space (if specified and a "capture" sample) or a default capture space (if no "capture_space" is provided in the sample_table
- Updated reference workflow to automatically parse (from the config file) and prepare reference files for a given set of BED intervals. This will automatically generate a bgzipped and tabix-indexed version of the BED, as well as generate an interval list
- The reference workflow will also check and update contig names (aka add or remove chr prefixes) as necessary


rdmorin (changes requested):
For version bumps you should submit initial pull requests with the modified files included in the most recent version that was in Master so the changes can be seen relative to that version. Can you please do that?

PR #189, opened 2021-08-20
Module/sigprofiler/1.0

ppararaj:
# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable

Example:
```
```


modules/sigprofiler/1.0/sigprofiler.smk
```diff
@@ -0,0 +1,207 @@
...
+rule _sigprofiler_all:
+    input:
+        expand(
+            expand(
+                CFG["dirs"]["outputs"] + "cosmic_sigs/{seq_type}--{genome_build}/{file}.COSMIC_{{type}}.De_Novo_map_to_COSMIC_{{type}}.csv",
+            zip,
+            seq_type = CFG["mafs"]["seq_type"],
+            genome_build = CFG["mafs"]["projected_build"],
+            file = CFG["mafs"]["filename"]),
+         type = ['SBS96', 'ID83', 'DBS78'])
```
rdmorin: type should be set in the config or something, don't you think?

ppararaj: I've specified type as a config parameter. I've also added the ability to specify other available k-mers.

modules/sigprofiler/1.0/src/python/run_generator.py
```diff
@@ -0,0 +1,20 @@
+import argparse
```
rdmorin, line 1: Please document what this script actually does with a few lines of in-line comments.

modules/sigprofiler/1.0/src/python/run_extractor.py
```diff
@@ -0,0 +1,42 @@
+import argparse
```
rdmorin, line 1: Document what this script does with a few lines of in-line comments

modules/sigprofiler/1.0/sigprofiler.smk
```diff
@@ -0,0 +1,215 @@
...
+# Symlinks the input files into the module results directory (under '00-inputs/')
+rule _sigprofiler_input_maf:
+    input:
+        maf = CFG["inputs"]["maf"]
+    output:
+        maf = CFG["dirs"]["inputs"] + "matrices/{seq_type}--{genome_build}/{sample_set}/{sample_set}.maf"
+    run:
+        op.relative_symlink(input.maf, output.maf)
+
+
```
rdmorin: Explain what this rule is actually doing with a comment at the top

modules/sigprofiler/1.0/sigprofiler.smk
```diff
@@ -0,0 +1,215 @@
...
+    "DBS2976" : 20
+}
+
+##### RULES #####
+
+rule _install_sigprofiler_genome:
+    output:
+        complete = CFG["dirs"]["inputs"] + "sigprofiler_prereqs/{genome_build}.installed"
+    params:
+        ref = lambda w: {"grch37":"GRCh37", "hg19":"GRCh37",
```
rdmorin: This needs to handle hs37d5 and eventually other genomes. Try to think of a way to generalize this.

modules/sigprofiler/1.0/sigprofiler.smk
```diff
@@ -0,0 +1,215 @@
...
+        $(awk 'BEGIN {{OFS=FS=","}} $1 ~ "*" {{S=substr($1,1,length($1)-1); print S+2}}' {input.stat})
+        {params.nmf_repl} {params.norm} {params.nmf_init} {threads}
+        """)
+
+# Symlinks the final output files into the module results directory (under '99-outputs/')
+rule _sigprofiler_output_tsv:
+    input:
+        decomp = str(rules._sigprofiler_run_extract.output.decomp)
+    output:
+        decomp = CFG["dirs"]["outputs"] + "cosmic_sigs/{seq_type}--{genome_build}/{sample_set}.COSMIC_{type}.De_Novo_map_to_COSMIC_{type}.csv"
```
rdmorin: Is genome_build here still actually genome_build or is it the projection?

modules/sigprofiler/1.0/sigprofiler.smk
```diff
@@ -0,0 +1,215 @@
...
+        $(awk 'BEGIN {{OFS=FS=","}} $1 ~ "*" {{S=substr($1,1,length($1)-1); if (S-2<1) {{print 1}} else {{print S-2}}}}' {input.stat})
+        $(awk 'BEGIN {{OFS=FS=","}} $1 ~ "*" {{S=substr($1,1,length($1)-1); print S+2}}' {input.stat})
+        {params.nmf_repl} {params.norm} {params.nmf_init} {threads}
+        """)
+
+# Symlinks the final output files into the module results directory (under '99-outputs/')
+rule _sigprofiler_output_tsv:
+    input:
+        decomp = str(rules._sigprofiler_run_extract.output.decomp)
+    output:
```
rdmorin, line 222: Can you output a .contents file too, like the examples I showed you?

workflows/reference_files/2.4/reference_files.smk
```diff
@@ -381,3 +381,37 @@ rule get_mutect2_small_exac:
...
+    shell:
+        "pip install SigProfilerMatrixGenerator && touch {output.complete}"
+
+rule install_sigprofiler_genome:
+    input:
+        str(rules.install_sigprofiler_matrix_generator.output.complete)
+    output:
+        complete = "downloads/sigprofiler_prereqs/{genome_build}.installed"
+    params:
+        ref = lambda w: {"grch37":"GRCh37", "hg19":"GRCh37",
```
rdmorin: Same comment regarding the additional genomes not handled here (there are likely going to be more)

rdmorin (changes requested):

modules/sigprofiler/1.0/sigprofiler.smk
ppararaj: This refers to the projection build. See the expand call in rule `_sigprofiler_all`

modules/sigprofiler/1.0/src/python/run_generator.py
ppararaj: Added comments

modules/sigprofiler/1.0/src/python/run_extractor.py
ppararaj: Added comments and argparse help

modules/sigprofiler/1.0/sigprofiler.smk
ppararaj: Added rule comments

ppararaj:
- [x] Rule comments
- [x] Script comments
- [x] Additional genome support
- [x] contents file generation
- [x] parameterize sigprofiler options

modules/sigprofiler/1.0/src/python/run_extractor.py
```diff
@@ -24,17 +25,17 @@ def main():
...
-    parser.add_argument('ref')
-    parser.add_argument('context_type', default = '96,DINUC,ID')
-    parser.add_argument('exome', default = False, type = bool)
-    parser.add_argument('minimum_signatures', default = 1, type = int)
-    parser.add_argument('maximum_signatures', default = 25, type = int)
-    parser.add_argument('nmf_replicates', default = 100 , type = int)
-    parser.add_argument('matrix_normalization', default = 'gmm')
-    parser.add_argument('nmf_init', default = 'nndsvd_min')
-    parser.add_argument('cpu', default = -1, type = int)
+    parser.add_argument('input_data', help = 'Path to matrix that will undergo NMF.')
```
rdmorin, line 44: Have you set it up so these options are available to be tweaked by the user via the config?

modules/sigprofiler/1.0/src/python/run_extractor.py
```diff
@@ -2,9 +2,10 @@
 from SigProfilerExtractor import sigpro as sig
 import os
 
-os.environ["OPENBLAS_NUM_THREADS"] = "1"
+os.environ["OPENBLAS_NUM_THREADS"] = "1" # ulimit fix
```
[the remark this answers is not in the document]
ppararaj: Doesn't seem to affect the multithreading of NMF. However not setting this option causes errors on all machines (from what I remember).

modules/sigprofiler/1.0/config/default.yaml
```diff
@@ -30,10 +70,10 @@ lcr-modules:
 
         pairing_config:
             genome:
-                run_paired_tumours: True
-                run_unpaired_tumours_with: "unmatched_normal"
+                run_paired_tumours: False
```
rdmorin: Why was this changed? Won't this make it harder to find the input files?

modules/sigprofiler/1.0/sigprofiler.smk
```diff
@@ -89,22 +88,17 @@ max_sigs = {
     "DBS2976" : 20
 }
 
-##### RULES #####
```
rdmorin: We really shouldn't be hard-coding references into a module like this. Can't this be set in the config?

modules/sigprofiler/1.0/config/default.yaml
ppararaj: I was trying to figure out a way to get around an error but I can just include general.yaml config file. The error had to do with an unmatched normal not being provided.

lkhilton (approved):

lkhilton (approved):

PR #192, opened 2021-09-07
Added support for capture data to reference workflow + SLMS-3

ckrushton:
- Added functions and rules to the reference workflow that prepares capture spaces
    - Capture spaces are defined in the reference config
    - Intervals are padded, sorted, merged for overlaps, and the `chr` prefix is added/removed as necessary
    - A version of every capture space is generated for every genome build associated with a given genome build version (i.e. a capture space associated with GRCh37 will be built for hs37d5, hg19, GRCh37 etc)
- Added a function within Strelka, Manta, MuTect, LoFreq, and SAGE that obtains the appropriate capture space for a given sample
- Modified vcf2maf module to support additional reference genomes
- Bugfixes


modules/lofreq/1.1/lofreq.smk
```diff
@@ -97,7 +105,8 @@ rule _lofreq_preprocess_normal:
         stdout = CFG["logs"]["lofreq_normal"] + "{seq_type}--{genome_build}/{normal_id}/lofreq_pre.stdout.log",
         stderr = CFG["logs"]["lofreq_normal"] + "{seq_type}--{genome_build}/{normal_id}/lofreq_pre.stderr.log"
     params:
-        opts = CFG["options"]["lofreq"]
+        opts = CFG["options"]["lofreq"],
+        bed = lambda w: _lofreq_get_capspace(w.normal_id, w)
```
Kdreval: it looks like `normal_id` wildcard is passed here, but the function above 
```
def _lofreq_get_capspace(sample_id, wildcards):
```
treats it as `sample_id`. Does this matter? Are they always the same?

Kdreval: Another related question here: the `bed` is moved to params. It goes to the function `_lofreq_get_capspace`, which in turn refers to the file generated by the reference_files. If that file does not exist already, because it is not listed in the input it will not be generated by snakemake prior to run. Did you test it "de novo", when the user does not yet have the pre-existing reference files? The same applies to other snakefiles where input is moved to params and is evaluated with function utilizing reference_files

modules/lofreq/1.1/lofreq.smk
```diff
@@ -83,7 +92,6 @@ rule _lofreq_preprocess_normal:
         normal_bam = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{normal_id}.bam",
         fasta = reference_files("genomes/{genome_build}/genome_fasta/genome.fa"),
         dbsnp = reference_files("genomes/{genome_build}/variation/dbsnp.common_all-151.vcf.gz"), #in our experience, this filter doesn't remove as many SNPs as one would expect
```
Kdreval, line 107: trailing comma at the end of the line here, not sure if it causes (or will cause in later snakemake versions) any issues. Same also goes to other places where bed was moved to params

modules/manta/2.3/manta.smk
```diff
@@ -75,22 +75,22 @@ rule _manta_configure_paired:
...
     output:
         runwf = CFG["dirs"]["manta"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/runWorkflow.py"
     log:
         stdout = CFG["logs"]["manta"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/manta_configure.stdout.log",
         stderr = CFG["logs"]["manta"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/manta_configure.stderr.log"
     params:
         opts = op.switch_on_wildcard("seq_type", CFG["options"]["configure"]),
-        tumour_bam_arg_name = op.switch_on_wildcard("seq_type", CFG["switches"]["tumour_bam_arg_name"])
+        tumour_bam_arg_name = op.switch_on_wildcard("seq_type", CFG["switches"]["tumour_bam_arg_name"]),
+        bedz = lambda w:_manta_get_capspace(w)
```
Kdreval: Don't python need first to define a function before referencing it? It is good to know it works like that! I guess it is because we have rule `all` at the end, so the evaluation goes bottom-to-top?

modules/manta/2.3/manta.smk
```diff
@@ -75,22 +75,22 @@ rule _manta_configure_paired:
         normal_bam = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{normal_id}.bam",
         fasta = reference_files("genomes/{genome_build}/genome_fasta/genome.fa"),
         config = op.switch_on_wildcard("seq_type", CFG["switches"]["manta_config"]),
-        bedz = str(rules._manta_index_bed.output.bedz)
```
Kdreval, line 96: What does the rule `_manta_index_bed` do now when implemented with functions? Is it still necessary in this version?

modules/mutect2/2.0/mutect2.smk
```diff
@@ -102,12 +102,12 @@ rule _mutect2_get_sm:
 # The second replaces wildcards with those used in the rule. 
 def _mutect2_get_interval_cli_arg(
     vcf_in = config["lcr-modules"]["mutect2"]["inputs"]["candidate_positions"], 
-    interval_arg_in = config["lcr-modules"]["mutect2"]["options"]["mutect2_interval_rules"]
+    #interval_arg_in = config["lcr-modules"]["mutect2"]["options"]["mutect2_interval_rules"]
```
Kdreval: if it's not needed let's rather remove it and not leave commented

modules/mutect2/2.0/mutect2.smk
```diff
@@ -178,7 +179,8 @@ rule _mutect2_run_no_normal:
     params:
         mem_mb = lambda wildcards, resources: int(resources.mem_mb * 0.8),
         opts = CFG["options"]["mutect2_run"], 
-        interval_arg = _mutect2_get_interval_cli_arg 
+        interval_arg = _mutect2_get_interval_cli_arg(),
```
Kdreval: This reminds that more than one input function is allowed, thanks!

modules/vcf2maf/1.2/vcf2maf.smk
```diff
@@ -69,7 +69,7 @@ rule _vcf2maf_run:
         stderr = CFG["logs"]["vcf2maf"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/{base_name}_vcf2maf.stderr.log",
     params:
         opts = CFG["options"]["vcf2maf"],
-        build = lambda w: VERSION_MAP[w.genome_build],
+        build = lambda w: GENOME_VERSION_MAP[w.genome_build],
```
Kdreval: Can you please rename this to include module name? `VCF2MAF_GENOME_VERSION_MAP`

workflows/reference_files/2.4/config/default.yaml
```diff
@@ -27,6 +27,31 @@ genome_builds:
...
+        default: "true"
+        capture_bed_url: "https://www.bcgsc.ca/downloads/lcr-modules/genome_fastas/capture_bed/grch38_all_genes.canonical.sort.bed4"
+    exome-utr-grch37:
+        genome: "grch37"
+        provider: "ensembl"
+        default: "true"
+        capture_bed_url: "https://www.bcgsc.ca/downloads/lcr-modules/genome_fastas/capture_bed/grch37_all_genes.canonical.sort.bed4"
+
+capture_params:
+    padding_size: "200"
```
Kdreval, line 85: can this be changed through config?

workflows/reference_files/2.4/reference_files.smk
```diff
@@ -383,6 +383,160 @@ rule get_mutect2_small_exac:
...
+    sample_table = CFG["samples"]
+    sample = sample_table.loc[(sample_table['sample_id'] == sample_id) &
+            (sample_table['genome_build'] == genome_build) &
+            (sample_table['seq_type'] == seq_type)]
+    if len(sample) != 1:
+        raise AssertionError("Found %s matches when examining the sample table for pair \'%s\' \'%s\' \'%s\'" % (len(sample), sample_id, genome_build, seq_type))
+    panel = sample.iloc[0]['capture_space']
+
+    # If this panel is "none" (aka not specified) use the default for this reference genome
+    if panel == "none":
```
Kdreval: Do you think it will be helpful to allow for more flexibility here, allowing for N/As, NAs, None, etc.. ?

Kdreval (commented):
Thanks! Few notes I mentioned here, please see here. Also, I thought you mentioned earlier about updates to the SAGE's `def get_chromosomes(wildcards)` - it is not in this PR, so is it handled elsewhere?

modules/lofreq/1.1/lofreq.smk
```diff
@@ -47,15 +47,24 @@ SCRIPT_PATH = CFG['inputs']['src_dir']
...
     _lofreq_input_bam,
     _lofreq_output_vcf,
     _lofreq_all,
     _lofreq_link_to_preprocessed
 
+def _lofreq_get_capspace(sample_id, wildcards):
+
+    # If this is a genome sample, return a BED file listing all chromosomes
+    if wildcards.seq_type != "capture":
+        return CFG["switches"]["regions_bed"]
```
Kdreval: Another question regarding this function (and other similar in other modules): Usually you need to define `CFG` within a function because it gives NameError of `name 'CFG' is not defined` or similar. There is even [documentation](https://lcr-modules.readthedocs.io/en/latest/faq.html#why-am-i-running-into-a-nameerror-name-cfg-is-not-defined-exception) section on it. How does it work here?

modules/lofreq/1.1/lofreq.smk
ckrushton: I added sample_id as a variable to allow the same function to be used for ` _lofreq_run_tumour` and `_lofreq_preprocess_normal`, since the function can't know if the normal or tumour sample is being used

ckrushton: I'll set up a test and see if that works

modules/lofreq/1.1/lofreq.smk
ckrushton: In my testing, the trailing comma doesn't cause any issues (has to do with how iterables are parsed in python). For example:
```
>>> test_lst = [8, 1,5,]
>>> print(test_lst)
[8, 1, 5]
```

modules/manta/2.3/manta.smk
ckrushton: Python does need functions (and all variables) to be defined before they are called, but snakemake is weird. I think everything is compiled into python first before snakemake is actually called, so functions can be defined anywhere

modules/manta/2.3/manta.smk
ckrushton: I moved that function to the reference workflow

modules/mutect2/2.0/mutect2.smk
ckrushton: Will do, did this for testing purposes

workflows/reference_files/2.4/config/default.yaml
ckrushton: Yes, it should be able to be superseded

workflows/reference_files/2.4/reference_files.smk
ckrushton: Yes, I'll have to do a bit more testing. Let me look into this

modules/lofreq/1.1/lofreq.smk
ckrushton: I'm not entirely sure, but I think it's because everything in the `input` and `params` blocks are executed before `op.cleanup_module()` is called

modules/lofreq/1.1/lofreq.smk
rdmorin: Any update on this?

modules/lofreq/1.1/lofreq.smk
rdmorin: I leave trailing commas in my rules pretty often.

modules/lofreq/1.1/lofreq.smk
rdmorin: Yeah it's counter-intuitive but my read of the section in the docs is that functions in general will not have access to this but functions that are called by the input directive are an exception.
```
 If you use CFG in the rule directives that are evaluated when the module snakefile is parsed (e.g. input, output, log, params, etc.), it’s not an issue. However, if you use this variable in a function or run directive, i.e. code that is run after the op.cleanup_module() function is run, you will get the error above.
```

modules/lofreq/1.1/lofreq.smk
ckrushton: This approach doesn't seem to generate the BED files, which is sort of as expected. While I can't move this to `input` without causing a whole host of other problems, I think I might be able to modify the `_<tool>_all` rule to specify all the BED files which are needed and generate those prior to running the main workflow

rdmorin:
What's happening with this?

modules/lofreq/1.1/lofreq.smk
ckrushton: Note to self: Code LoFreq so it will use CFG["switches"]["regions_bed"] if it exists

modules/strelka/1.1/strelka.smk
```diff
@@ -108,7 +108,6 @@ rule _strelka_configure_paired: # Somatic
         tumour_bam = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{tumour_id}.bam",
         normal_bam = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{normal_id}.bam",
         fasta = reference_files("genomes/{genome_build}/genome_fasta/genome.fa"),
-        bedz = str(rules._strelka_index_bed.output.bedz),
```
ckrushton, line 131: Add this 01-folder back in when you move this file back as an input

modules/lofreq/1.1/lofreq.smk
Kdreval: I have tested new setup on Cedar and can confirm that when bed file is returned to inputs from params it works on "de novo" setup and all files required from `reference_files` age generated if not already present.

Kdreval:
All review comments and outstanding changes have been addressed. This version was tested both on GSC and Cedar. It is ready for re-review

rdmorin:
Does this still need more revisions?

Kdreval:
Yes, this still need some updates. I am tracking down bugs and making sure it works consistently. Haven't pushed some recent bug fixes, but will do it today. Can focus on finalizing it this weekend and making sure it is good to go.

workflows/reference_files/2.4/config/default.yaml
```diff
@@ -47,6 +47,36 @@ genome_builds:
         version: "grch38"
         provider: "ucsc"
         genome_fasta_url: "https://hgdownload.cse.ucsc.edu/goldenpath/hg38/bigZips/hg38.fa.masked.gz"
+    hg19-reddy:
```
lkhilton, line 50: Do these bespoke reference file configs need to be in master, or can they just be added in the context of GAMBL?

oncopipe/oncopipe/__init__.py
```diff
@@ -1784,3 +1785,52 @@ def cleanup_module(module_config):
...
+    # Convenient variable to access sample table
+    module_samples = module_config["samples"]
+
+    this_sample = module_samples.loc[(module_samples['sample_id'] == sample_id) &
+            (module_samples['genome_build'] == genome_build) &
+            (module_samples['seq_type'] == seq_type)]
+
+    if len(this_sample) != 1:
+        raise AssertionError("Found %s matches when examining the sample table for pair \'%s\' \'%s\' \'%s\'" % (len(sample), sample_id, genome_build, seq_type))
+    panel = this_sample.iloc[0]['capture_space']
```
lkhilton: Does this throw an error if there is no column called `capture_space` in the samples table?

demo/data/samples.tsv
```diff
@@ -1,6 +1,6 @@
-sample_id	seq_type	patient_id	tissue_status	genome_build	strand	read_length
-TCRBOA7-N-WEX	capture	TCRBOA7	normal	grch37	positive	100
-TCRBOA7-T-WEX	capture	TCRBOA7	tumour	grch37	positive	100
-TCRBOA7-T-RNA	mrna	TCRBOA7	tumour	grch37	positive	100
-TCRBOA7-N-WGS	genome	TCRBOA7	normal	grch37	positive	100
-TCRBOA7-T-WGS	genome	TCRBOA7	tumour	grch37	positive	100
+sample_id	seq_type	patient_id	tissue_status	genome_build	strand	read_length	capture_space
```
lkhilton: Can you clarify whether capture_space becomes a required column under these changes?

modules/mutect2/2.0/mutect2.smk
```diff
@@ -123,18 +123,34 @@ rule _mutect2_get_sm:
 # The first function loads the wildcard-containing file path and additional args from the config. 
 # The second replaces wildcards with those used in the rule. 
 def _mutect2_get_interval_cli_arg(
-    vcf_in = config["lcr-modules"]["mutect2"]["inputs"]["candidate_positions"],
```
lkhilton, line 126: I can't comment on code that's unchanged, but the Mutect2 has long been awaiting a permanent bug fix for the instance where the input candidate_positions file doesn't cover all of the chromosomes in `reference_files("genomes/{genome_build}/genome_fasta/main_chromosomes.txt")`. I think we should also cover the possibility that the cap_space bed file doesn't span all chromosomes (I have a capture panel for which this is true).  I have a few thoughts on how this could be handled but either way I think somehow it needs to precede the checkpoint.

modules/mutect2/2.0/mutect2.smk
```diff
@@ -212,7 +230,7 @@ rule _mutect2_run_no_normal:
         gatk Mutect2 --java-options "-Xmx{params.mem_mb}m" 
         {params.opts} -I {input.tumour_bam} -R {input.fasta} 
         -O {output.vcf} --germline-resource {input.gnomad} 
-        -L {wildcards.chrom} {params.interval_arg}
+        -L {wildcards.chrom} {params.interval_arg} -L {input.capture_arg}
         -pon {input.pon} --f1r2-tar-gz {output.f1r2}
         > {log.stdout} 2> {log.stderr}
         """)
```
lkhilton, line 250: Actually it's probably the functions below `def _mutect2_get_chr_vcfs(wildcards)` and `def _mutect2_get_chr_tbis(wildcards)` that need to be updated to subset the chromosome list while considering capspace and candidate positions

lkhilton (changes requested):
Some questions and some outstanding work on the Mutect2 module (sorry!)

oncopipe/oncopipe/__init__.py
Kdreval: Yes, it does throw an error, specifically related to the capture_space:
```
Building DAG of jobs...
InputFunctionException in line 139 of /home/kdreval/lcr-modules/modules/sage/1.0/sage.smk:
Error:
  KeyError: 'capture_space'
Wildcards:
  seq_type=capture
  genome_build=grch37
  tumour_id=TCRBOA7-T-WGS
  normal_id=TCRBOA7-N-WEX
  pair_status=unmatched
Traceback:
  File "/home/kdreval/lcr-modules/modules/sage/1.0/sage.smk", line 128, in _sage_get_capspace
  File "/home/kdreval/lcr-modules/oncopipe/oncopipe/__init__.py", line 1838, in get_capture_space
  File "/projects/rmorin_scratch/conda_environments/gambl-opv12/lib/python3.8/site-packages/pandas/core/series.py", line 942, in __getitem__
  File "/projects/rmorin_scratch/conda_environments/gambl-opv12/lib/python3.8/site-packages/pandas/core/series.py", line 1051, in _get_value
  File "/projects/rmorin_scratch/conda_environments/gambl-opv12/lib/python3.8/site-packages/pandas/core/indexes/base.py", line 3363, in get_loc
```

demo/data/samples.tsv
Kdreval: It is, yes. I will add it to the CHANGELOG to reflect this change

workflows/reference_files/2.4/config/default.yaml
Kdreval: I think having them here is fine, as they won't be generated unless there are samples with that specific reference.

modules/mutect2/2.0/mutect2.smk
Kdreval: I can work on these changes within this PR, and implement this bug fix. I haven't run mutect2 explicitly but can read the documentation on how to deal with this and come up with some suggestions. Will work on it

ckrushton: Is this the bug where MuTect2 errors out because basically there are no input regions specified (i.e. the overlap between all the files/regions is empty)

oncopipe/oncopipe/__init__.py
lkhilton: So does the GAMBL samples table now require this column? Is it possible to add a condition to check if that column is in the samples table, and if not return the default capspace?

Kdreval: I can make this an optional column, of course 👍

Kdreval: This has been modified to make a column not required. If the `capture_space` is not in sample table, the default capture space is returned

modules/mutect2/2.0/mutect2.smk
Kdreval: Yes, Chris - it is about this bug

ckrushton: I think me and Haya came up with a solution, but I'm not sure what happened to that

Kdreval: The checkpoint here has been modified to intersect 3 files of  a) main chromosomes, b) candidate positions, and c) capture space (which is set to main chromosomes in case of non-capture samples). The intersect of chrs in these 3 files will dictate mutect2 on which chromosomes to run ✅

modules/mutect2/2.0/mutect2.smk
Kdreval: I ended up keeping these functions as is - because they read in the file with chromosomes produced by checkpoint. Instead, I modified the checkpoint to address this. This is resolved now

Kdreval:
Happy to report this PR is ready for re-review. In addition to maintaining current functionality, I ensured this is now reproducible, works for both exomes and genomes, and regenerates all reference files if none are present upfront. In addition, the `capture_space` column is optional, mutect2 will run on chromosomes deemed mutated through candidate positions and will not error out if some chromosomes are missing. This version can take local `bed` files to use for specific custom capture without downloading them from the internet.

lkhilton (approved):
🏆

lkhilton (approved):

PR #195, opened 2021-11-03
Module/vcf2maf/1.3

Kdreval:
This modification bumps the vcf2maf version to 1.3 and uses modified perl script to ensure GnomAD compatibility by attaching the AF values.


modules/vcf2maf/1.2/vcf2maf.smk
```diff
@@ -79,19 +83,30 @@ rule _vcf2maf_run:
         **CFG["resources"]["vcf2maf"]
     shell:
         op.as_one_line("""
+        VCF2MAF_SCRIPT_PATH={VCF2MAF_SCRIPT_PATH};
```
rdmorin: I think there's some redundancy here. Why are you updating PATH but still explicitly specifying the path to the Perl script? You should be able to just say "vcf2maf.pl" but you should check to see how we handled this for LoFreq.

modules/vcf2maf/1.2/vcf2maf.smk
```diff
@@ -79,19 +83,30 @@ rule _vcf2maf_run:
...
-        --input-vcf {input.vcf} 
-        --output-maf {output.maf} 
-        --tumor-id {wildcards.tumour_id} --normal-id {wildcards.normal_id}
-        --ref-fasta {input.fasta}
-        --ncbi-build {params.build}
-        --vep-data {input.vep_cache}
-        --vep-path $vepPATH {params.opts}
-        --custom-enst {params.custom_enst}
-        > {log.stdout} 2> {log.stderr}
+        if [[ $(which vcf2maf.pl) =~ $VCF2MAF_SCRIPT ]]; then
```
rdmorin: Oh I see this is a check to ensure you're using the bundled script. Makes sense

workflows/reference_files/2.4/reference_files.smk
```diff
@@ -385,6 +385,20 @@ rule get_af_only_gnomad_vcf:
         tabix {output.vcf}
         """)
 
+rule normalize_af_only_gnomad_vcf:
+    input:
+        fasta = rules.get_genome_fasta_download.output.fasta,
+        vcf = str(rules.get_af_only_gnomad_vcf.output.vcf)
+    output:
+        vcf = "genomes/{genome_build}/variation/af-only-gnomad.normalized.{genome_build}.vcf.gz"
```
rdmorin: Can you make the tabix index an output also?

rdmorin (changes requested):
Looks good. Nothing major

workflows/reference_files/2.4/reference_files.smk
Kdreval: Sure, I have specified the tabix index as rule output here

Kdreval:
I made changes to implement the gnomAD filtering and handle bcftools normalisation of hg38 vcf with alternative contigs. This is now also tested in GAMBL settings with cell lines

Kdreval:
This module has been re-worked to implement gnomAD annotations through bcftools. It is tested both on hg38 and grch37 cell lines and ready for re-review

rdmorin (approved):

PR #196, opened 2021-11-03
Module (strelka, sage, lofreq, mutect) improvements

hayashaalan:
# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable

Example:
```
```

### DIFF
--- a/modules/lofreq/1.1/lofreq.smk
+++ b/modules/lofreq/1.1/lofreq.smk
@@ -47,7 +47,9 @@ SCRIPT_PATH = CFG['inputs']['src_dir']
 bed = str(reference_files("genomes/{genome_build}/genome_fasta/main_chromosomes.bed"))
 CFG['switches']['regions_bed']['_default'] = bed
 
-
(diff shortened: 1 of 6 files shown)


modules/lofreq/1.1/lofreq.smk
```diff
@@ -123,43 +125,43 @@ rule _lofreq_preprocess_normal:
...
-        ln -f {input.vcf_indels_stringent}.tbi $out_dir;
-        ln -f {input.vcf_snvs_stringent} $out_dir;
-        ln -f {input.vcf_snvs_stringent}.tbi $out_dir;
-        ln -f {input.normal_relaxed_log} $out_dir;
-        ln -f {input.preprocessing_complete} $out_dir;
-        """)
+    wildcard_constraints: 
+        pair_status = "unmatched"
+    run:
+        op.relative_symlink(input.vcf_relaxed, output.vcf_relaxed)
```
lkhilton: Could you please add the `in_module=True` flag to all of these relative_symlink commands?

lkhilton (approved):

PR #197, opened 2021-11-03
Modules/lymphgen/1.0

ckrushton:
Added LymphGen as a module

# Pull Request Checklists

### Required

- [DONE] I used the cookiecutter template and updated the placeholder rules.

- [DONE] The snakemake rules follow the [design guidelines](https://lcr-modules.readthedocs.io/en/latest/for_developers.html#module-rules). 

- [DONE] All references to the `rules` object (_e.g._ for input files) are wrapped with `str()`.

- [DONE] Every rule in the module is either listed under `localrules` or has the `threads` and `resources` directives.
    - They are all under localrules since they run so quickly
   
- [DONE] Input and output files are being symlinked into the `CFG["inputs"]` and `CFG["outputs"]` subdirectories, respectively.
     - Except for the optional SV annotation file

- [DONE] I updated the final target rule (`*_all`) to include every output rule.

- [DONE] I explained important module design decisions in `CHANGELOG.md`.

- [DONE ] I tested the module on real data for all supported `seq_type` values.

- [DONE ] I updated the `default.yaml` configuration file to provide default values for each rule in the module snakefile.

- [DONE ] I did not set any global wildcard constraints. Any/all wildcard constraints are set on a per-rule basis. 
    - Custom wildcards

- [DONE ] I ensured that all symbolic links are relative and self-contained (_i.e._ do not point outside of the repository).

- [DONE ] I replaced every value that should (or might need to) be updated in the default configuration file with `__UPDATE__`.

- [DONE ] I recursively searched for all comments containing `TODO` to ensure none were left. For example:

  ```bash
  ```

### If applicable

- [DONE ] I updated the list of available wildcards for the input files in the `default.yaml` configuration file.
  - No user-settable wildcards

### DIFF
--- a/modules/lymphgen/1.0/config/default.yaml
+++ b/modules/lymphgen/1.0/config/default.yaml
@@ -4,7 +4,6 @@ lcr-modules:
 
         inputs:
             # Available wildcards: None (specify your merged input MAF/SEG manually)
-            # Sorry, but this was the most robust way
             sample_maf: "__UPDATE__"
             sample_seg: "__UPDATE__"
             # Path to the directory where LGenIC should be installed
@@ -18,7 +17,7 @@ lcr-modules:
                 start: "start"
                 end: "end"
                 CN: "log.ratio"
-            add_svs:  # Use "" if N/A
+            add_svs:  # Use "" if N/A (you are not providing SVs:)
                 samplecol: "__UPDATE__"
                 bcl2col: "__UPDATE__"
                 # Note any field value not matching in either true/false is considered NA
@@ -29,13 +28,14 @@ lcr-modules:
                 bcl6truevalues: "__UPDATE__"  # Can use a list
                 bcl6falsevalues: "__UPDATE__"
             lymphgen_input:
-                seq_type: "genome"
+                seq_type: "genome"  # Availible options: exome/genome (they do the same thing) and targeted (for panels)
                 use_log_ratio: "__UPDATE__"
             lymphgen_run:
-                lymphgen_path: "/home/crushton/Software/LymphGen/Run_Lymphgen.R"
+                lymphgen_path: "/projects/rmorin/software/lcr-scripts/LymphGen/Run_Lymphgen.R"
 
         conda_envs:
-            samtools: "{MODSDIR}/envs/samtools-1.9.yaml"
+            optparse: "{MODSDIR}/envs/r-optparse-1.6.yaml"
+            sorted_containers: "{MODSDIR}/envs/sortedcontainers-2.4.0.yaml"
 
         # Is this needed???
         pairing_config:
--- a/modules/lymphgen/1.0/config/default.yaml
+++ b/modules/lymphgen/1.0/config/default.yaml
@@ -2,26 +2,42 @@ lcr-modules:
     
     lymphgen:
 
-        # TODO: Update the list of available wildcards, if applicable
         inputs:
-            # Available wildcards: {seq_type} {genome_build} {sample_id}
+            # Available wildcards: None (specify your merged input MAF/SEG manually)
+            # Sorry, but this was the most robust way
             sample_maf: "__UPDATE__"
-
-        scratch_subdirectories: []
+            sample_seg: "__UPDATE__"
+            # Path to the directory where LGenIC should be installed
+            lgenic_exec: "__UPDATE__"
 
         options:
-            step_1: ""
-            step_2: ""
+            outprefix: "my_cohort"
+            reformat_seg:  #  Default columns for Battenberg output
+                Tumor_Sample_Barcode: "ID"
+                chromosome: "chrom"
+                start: "start"
+                end: "end"
+                CN: "log.ratio"
+            add_svs:  # Use "" if N/A
+                samplecol: "__UPDATE__"
+                bcl2col: "__UPDATE__"
+                # Note any field value not matching in either true/false is considered NA
+                bcl2truevalues: "__UPDATE__" # Can use a list here
+                bcl2falsevalues: "__UPDATE__"
+                bcl6col: "__UPDATE__"
+                # Note any field value not matching in either true/false is considered NA
+                bcl6truevalues: "__UPDATE__"  # Can use a list
+                bcl6falsevalues: "__UPDATE__"
+            lymphgen_input:
+                seq_type: "genome"
+                use_log_ratio: "__UPDATE__"
+            lymphgen_run:
+                lymphgen_path: "/home/crushton/Software/LymphGen/Run_Lymphgen.R"
 
         conda_envs:
             samtools: "{MODSDIR}/envs/samtools-1.9.yaml"
-            
-        threads:
-            step_1: 4
 
-        mem_mb:
-            step_1: 2000
-
+        # Is this needed???
         pairing_config:
             genome:
                 run_paired_tumours: True
--- a/modules/lymphgen/1.0/config/default.yaml
+++ b/modules/lymphgen/1.0/config/default.yaml
@@ -1,0 +1,33 @@
+lcr-modules:
+    
+    lymphgen:
+
+        # TODO: Update the list of available wildcards, if applicable
+        inputs:
+            # Available wildcards: {seq_type} {genome_build} {sample_id}
+            sample_maf: "__UPDATE__"
+
+        scratch_subdirectories: []
+
+        options:
+            step_1: ""
+            step_2: ""
+
+        conda_envs:
+            samtools: "{MODSDIR}/envs/samtools-1.9.yaml"
+            
+        threads:
+            step_1: 4
+
+        mem_mb:
+            step_1: 2000
+
+        pairing_config:
+            genome:
+                run_paired_tumours: True
+                run_unpaired_tumours_with: "unmatched_normal"
+                run_paired_tumours_as_unpaired: False
+            capture:
+                run_paired_tumours: True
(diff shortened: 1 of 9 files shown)


modules/lymphgen/1.0/config/default.yaml
```diff
@@ -0,0 +1,49 @@
+lcr-modules:
+    
+    lymphgen:
+
+        inputs:
+            # Available wildcards: None (specify your merged input MAF/SEG manually)
+            # Sorry, but this was the most robust way
```
lkhilton: No need to apologize, the SigProfiler module is similar.

modules/lymphgen/1.0/lymphgen.smk
```diff
@@ -0,0 +1,385 @@
...
+        cnv_arm = CFG["dirs"]["lymphgen_input"] + "{outprefix}.{cnvs_wc}_cnv_arm.tsv"
+    log:
+        stdout = CFG["logs"]["lymphgen_input"] + "{outprefix}.{cnvs_wc}.LGenIC.stdout.log",
+        stderr = CFG["logs"]["lymphgen_input"] + "{outprefix}.{cnvs_wc}.LGenIC.stderr.log"
+    params:
+        seq_type = CFG["options"]["lymphgen_input"]["seq_type"],
+        outprefix = "{outprefix}.{cnvs_wc}",
+        logratio = "--log2" if CFG["options"]["lymphgen_input"]["use_log_ratio"].lower() == "true" else ""
+    conda:
+        "envs/sortedcontainers-2.4.0.yaml"
```
lkhilton: This should be specified through the config so that it can be updated if there are needed changes to the env. 
```
CFG['conda_envs']['sorted_containers']
```

modules/lymphgen/1.0/lymphgen.smk
```diff
@@ -0,0 +1,385 @@
...
+        cnv_arm = str(rules._lymphgen_input_cnv.output.cnv_arm)
+    output:
+        result = CFG["dirs"]["lymphgen_run"] + "{outprefix}.results.{cnvs_wc}.{sv_wc}.tsv"
+    log:
+        stderr = CFG["logs"]["lymphgen_run"] + "{outprefix}.lymphgen.{cnvs_wc}.{sv_wc}.stderr.log",
+        stdout = CFG["logs"]["lymphgen_run"] + "{outprefix}.lymphgen.{cnvs_wc}.{sv_wc}.stdout.log"
+    params:
+        lymphgen_path = CFG["options"]["lymphgen_run"]["lymphgen_path"]
+    conda:
+        "envs/r-optparse-1.6.yaml"
```
lkhilton: `CFG['conda_envs']['optparse']`

modules/lymphgen/1.0/config/default.yaml
```diff
@@ -0,0 +1,49 @@
...
+                bcl6truevalues: "__UPDATE__"  # Can use a list
+                bcl6falsevalues: "__UPDATE__"
+            lymphgen_input:
+                seq_type: "genome"
+                use_log_ratio: "__UPDATE__"
+            lymphgen_run:
+                lymphgen_path: "/home/crushton/Software/LymphGen/Run_Lymphgen.R"
+
+        conda_envs:
+            samtools: "{MODSDIR}/envs/samtools-1.9.yaml"
```
lkhilton: Add the conda envs here. You can also delete this samtools one since you don't use it in the module, as well as the symlink that gets added by default through the cookiecutter.

modules/lymphgen/1.0/lymphgen.smk
```diff
@@ -0,0 +1,385 @@
...
+
+outprefix = CFG["options"]["outprefix"]
+
+# Are we running LymphGen with CNV data?
+if "sample_seg" in CFG["inputs"] and CFG["inputs"]["sample_seg"] != "" and CFG["inputs"]["sample_seg"] != "None":
+    cnvs_included = True
+else:
+    cnvs_included = False
+
+# DOWNLOAD CHRIS'S VERY AWESOME LYMPHGEN CONVERSION SCRIPT. ALL CAPS
```
lkhilton, line 55: I approve of this descriptive and necessary comment.

modules/lymphgen/1.0/config/default.yaml
```diff
@@ -0,0 +1,49 @@
...
+                bcl2falsevalues: "__UPDATE__"
+                bcl6col: "__UPDATE__"
+                # Note any field value not matching in either true/false is considered NA
+                bcl6truevalues: "__UPDATE__"  # Can use a list
+                bcl6falsevalues: "__UPDATE__"
+            lymphgen_input:
+                seq_type: "genome"
+                use_log_ratio: "__UPDATE__"
+            lymphgen_run:
+                lymphgen_path: "/home/crushton/Software/LymphGen/Run_Lymphgen.R"
```
lkhilton: Probably a good idea to move this somewhere central e.g. `/projects/rmorin/software`

modules/lymphgen/1.0/lymphgen.smk
```diff
@@ -0,0 +1,385 @@
...
+        gene_list = str(rules._lymphgen_input_no_cnv.output.gene_list)
+    output:
+        result = CFG["dirs"]["lymphgen_run"] + "{outprefix}.results.{cnvs_wc}.{sv_wc}.tsv"
+    log:
+        stderr = CFG["logs"]["lymphgen_run"] + "{outprefix}.lymphgen.{cnvs_wc}.{sv_wc}.stderr.log",
+        stdout = CFG["logs"]["lymphgen_run"] + "{outprefix}.lymphgen.{cnvs_wc}.{sv_wc}.stdout.log"
+    params:
+        lymphgen_path = CFG["options"]["lymphgen_run"]["lymphgen_path"]
+    conda:
+        "envs/r-optparse-1.6.yaml"
```
lkhilton: Here too!

modules/lymphgen/1.0/lymphgen.smk
```diff
@@ -0,0 +1,385 @@
...
+
+rule _lymphgen_all:
+    input:
+        expand(
+            [
+                str(rules._lymphgen_output_txt.output.txt),
+            ],
+            outprefix = outprefix,
+            cnvs_wc = cnvs_wc,
+            sv_wc = sv_wc
```
lkhilton, line 373: I suppose if you have both SVs and CNVs this makes four different output files because of the combination of wildcards? Good thinking.

modules/lymphgen/1.0/schemas/base-1.0.yaml
```diff
@@ -0,0 +1 @@
+../../../../schemas/base/base-1.0.yaml
```
lkhilton: Since this doesn't use the samples table, it's probably a good idea to create a new schema for this module that doesn't make _any_ requirements of the samples table and remove this one. Otherwise you can get annoying errors about missing stuff.

modules/lymphgen/1.0/envs/samtools-1.9.yaml
```diff
@@ -0,0 +1 @@
+../../../../envs/samtools/samtools-1.9.yaml
```
lkhilton: Delete this symlink

modules/lymphgen/1.0/lymphgen.smk
```diff
@@ -0,0 +1,385 @@
...
+
+##### SETUP #####
+
+
+# Import package with useful functions for developing analysis modules
+import oncopipe as op
+
+# Setup module and store module-specific configuration in `CFG`
+# `CFG` is a shortcut to `config["lcr-modules"]["lymphgen"]`
+CFG = op.setup_module(
```
lkhilton, line 25: How does this behave without a samples table? Or do you always use one to prevent this from erroring out?

lkhilton (changes requested):
Some minor changes requested. Can't wait to have this in production!

modules/lymphgen/1.0/config/default.yaml
ckrushton: Moved to `/projects/rmorin/software/lcr-scripts/LymphGen/`, and the default path in the config has been updated

modules/lymphgen/1.0/lymphgen.smk
ckrushton: By default this generates all possible variants of the provided inputs. i.e. if you just provide SNVs, it will only generate one output (no_cnvs, no_svs), whereas if you provide SNVs and CNVs it will run LymphGen with both variants (with_cnvs, no_svs + no_cnvs, no_svs)

modules/lymphgen/1.0/config/default.yaml
```diff
@@ -29,13 +28,14 @@ lcr-modules:
...
             lymphgen_run:
-                lymphgen_path: "/home/crushton/Software/LymphGen/Run_Lymphgen.R"
+                lymphgen_path: "/projects/rmorin/software/lcr-scripts/LymphGen/Run_Lymphgen.R"
 
         conda_envs:
-            samtools: "{MODSDIR}/envs/samtools-1.9.yaml"
+            optparse: "{MODSDIR}/envs/r-optparse-1.6.yaml"
+            sorted_containers: "{MODSDIR}/envs/sortedcontainers-2.4.0.yaml"
 
         # Is this needed???
```
lkhilton: You can definitely delete the pairing config

lkhilton (approved):

lkhilton (approved):

PR #203, opened 2022-01-05
Modules/svar master/1.0

lkhilton:
# Pull Request Checklists

## Checklist for New Module

A module to: 
1. Intersect GRIDSS and Manta
2. Annotate the resulting BEDPE from a bed file of annotations
3. Liftover the BEDPE file

### Required

  ```bash
  ```

### If applicable


modules/liftover/2.0/src/convert_for_liftover.py
```diff
@@ -22,57 +22,90 @@ def main():
     # read required second argument from shell command as path to resulting .bed file
```
lkhilton, line 22: I should comment here that Kostia reviewed this on my last PR that was already merged into master and I accidentally overwrote it with the old version while completing the version bump process.

modules/svar_master/1.0/svar_master.smk
```diff
@@ -0,0 +1,251 @@
...
+##### Intersect Manta and GRIDSS #####
+
+rule _svar_master_input_gridss:
+    input:
+        vcf = str(rules._gridss_output_somatic_vcf.output.filtered), 
+        tbi = str(rules._gridss_output_somatic_vcf.output.filtered_tbi)
+    output:
+        vcf = CFG_SV["dirs"]["inputs"] + "sv_vcfs/{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}.gridss.somatic_filtered.vcf.gz", 
+        tbi = CFG_SV["dirs"]["inputs"] + "sv_vcfs/{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}.gridss.somatic_filtered.vcf.gz.tbi"
+    group:
```
Kdreval, line 105: Sorry, just a bit confused here. Seems that there are 2 rules combined in this group, and both are just symlinking steps. Is it meant to be bundled with some other rule that I missed, or maybe it is meant to bundle with part of gridss/hmftools pipelines?

modules/svar_master/CHANGELOG.md
```diff
@@ -0,0 +1,14 @@
...
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [1.0] - 2021-12-29
+
+This release was authored by Laura Hilton.
+
+- This module integrates Manta, GRIDSS, and the HMFtools PURPLE-LINX tools into a single pipeline. It combines the Manta and GRIDSS SV calls into a single bedpe file and annotates the breakpoints by the nearest gene/locus based on a user-provided bed file. 
+- Since this module includes the GRIDSS and Manta Snakefiles but also depends on the outputs of SLMS-3, SLMS-3 MUST be run before this module. It is also highly recommended to run GRIDSS on the majority of samples prior to launching the complete pipeline. This can be accomplished by specifying `_gridss_all` as the target when launching Snakemake.
```
Kdreval, line 13: Can this maybe be duplicated as commented instruction in config? I think not all (many) users might check the CHANGELOG to read this :smile:

modules/svar_master/1.0/src/combine_annotated_bedpe.R
```diff
@@ -0,0 +1,111 @@
...
+        "ANNOTATION_B", 
+        "DIST_TO_ANNOTATION_B"
+      )) %>% 
+      # Remove small intrachromosomsal events
+      filter(!(CHROM_A == CHROM_B & abs(START_A - START_B) < 5000)) %>% 
+      distinct(CHROM_A, CHROM_B, START_A, START_B, END_A, END_B, STRAND_A, STRAND_B, .keep_all = TRUE)
+  } 
+write_tsv(combined_bedpe, snakemake@output[["bedpe"]])
+
+})
```
Kdreval, line 119: this is missing a newline character

modules/svar_master/1.0/config/default.yaml
```diff
@@ -0,0 +1,107 @@
...
+                    grch37: "__UPDATE__"
+                    # Bed file of e.g. genes, regions, superenhancers, etc. to be used
+                    # Use genome_build as the key and the path to the bed file as the value here  
+                    # Must be sorted by the 'sort -k1,1 -k2,2n' method
+
+
+        options:
+            intersect: 
+                combine_svs: "{MODSDIR}/src/combine_sv_vcfs.R"
+                bedpe_cols: "{MODSDIR}/etc/bedpe_combined_cols.tsv"
```
Kdreval, line 50: Should we add an in-line comment here to mention what this file is in case if user needs to modify the column headers, or maybe have something other that they may want/need to have?

modules/svar_master/1.0/src/combine_annotated_bedpe.R
```diff
@@ -0,0 +1,111 @@
...
+                           "CHROM_ANN_B", 
+                           "START_ANN_B", 
+                           "END_ANN_B", 
+                           "ANNOTATION_B", 
+                           "DIST_TO_ANNOTATION_B"
+                         ))
+  
+  
+  
+  if(nrow(combined_a) > 0){
```
Kdreval, line 86: should there be a warning/error message if there are 0 rows in the input bed? Just in case the input is empty, I imagine the 0 rows would give a not straightforward error message. But this can also be kept in mind for future updates and next versions, I think it does not necessarily apply for the samples we've seen so far :smile:

modules/svar_master/1.0/src/combine_annotated_bedpe.R
```diff
@@ -0,0 +1,111 @@
+suppressWarnings({
```
Kdreval, line 3: Does it need a shebang line here?

Kdreval (approved):
Thanks! Just a few questions but looks very good!!

modules/svar_master/1.0/svar_master.smk
lkhilton: Ack! That's a new feature of the cookie cutter that I completely missed. Will fix.

modules/svar_master/CHANGELOG.md
lkhilton: Good idea

modules/svar_master/1.0/config/default.yaml
lkhilton: I think modifying this file would break the other parts of the pipeline, unfortunately.

modules/svar_master/1.0/src/combine_annotated_bedpe.R
lkhilton: Can add

modules/svar_master/1.0/src/combine_annotated_bedpe.R
lkhilton: Yep, this if statement was added because I did have 0-row files when I developed this script for the capture data. This allows the pipeline to run to completion even if the bed file is empty.

modules/svar_master/1.0/svar_master.smk
Kdreval: oh right, this is from the cookiecutter to show users how to group rules, I also forgot about this!

Kdreval (approved):

PR #207, opened 2022-01-14
Module/stringtie/1.0

kmcoyle:
# Pull Request Checklists

### Required

- [x ] I used the cookiecutter template and updated the placeholder rules.

- [x ] The snakemake rules follow the [design guidelines](https://lcr-modules.readthedocs.io/en/latest/for_developers.html#module-rules). 

  - [x ] All references to the `rules` object (_e.g._ for input files) are wrapped with `str()`.

- [ x] Every rule in the module is either listed under `localrules` or has the `threads` and `resources` directives.

- [ x] Input and output files are being symlinked into the `CFG["inputs"]` and `CFG["outputs"]` subdirectories, respectively.

- [ x] I updated the final target rule (`*_all`) to include every output rule.

- [ x] I explained important module design decisions in `CHANGELOG.md`.

- [ x] I tested the module on real data for all supported `seq_type` values.

- [ x] I updated the `default.yaml` configuration file to provide default values for each rule in the module snakefile.

- [ x] I did not set any global wildcard constraints. Any/all wildcard constraints are set on a per-rule basis. 

- [ x] I ensured that all symbolic links are relative and self-contained (_i.e._ do not point outside of the repository).

- [ x] I replaced every value that should (or might need to) be updated in the default configuration file with `__UPDATE__`.

- [ x] I recursively searched for all comments containing `TODO` to ensure none were left. For example:

  ```bash
  ```

### If applicable

Example:
```
```


modules/stringtie/1.0/envs/stringtie.yaml
```diff
@@ -0,0 +1,43 @@
+name: stringtie
```
lkhilton: Can you please put this file in `lcr-modules/envs` and then create a symlink from the `modules/stringtie/1.0/envs` directory pointing at this one?

modules/stringtie/1.0/envs/samtools-1.9.yaml
```diff
@@ -0,0 +1 @@
+../../../../envs/samtools/samtools-1.9.yaml
```
lkhilton: You can delete this symlink from your module since it's not used.

modules/stringtie/1.0/stringtie.smk
```diff
@@ -0,0 +1,124 @@
...
+    _stringtie_all,
+
+
+##### RULES #####
+
+
+# Symlinks the input files into the module results directory (under '00-inputs/')
+rule _stringtie_input_bam:
+    input:
+        bam = CFG["inputs"]["sample_bam"],
```
lkhilton, line 56: I recommend including the `.bai` file as a second input here. The STAR module creates the index in a separate rule, and if one tried to chain these modules together this rule could try to run before the index is created and error out.

modules/stringtie/1.0/config/default.yaml
```diff
@@ -0,0 +1,27 @@
+lcr-modules:
+
+    stringtie_run:
+
+        inputs:
+            # Available wildcards: {seq_type} {genome_build} {sample_id}
+            sample_bam: "__UPDATE__"
```
lkhilton, line 8: Does this module _require_ STAR bam files? If yes, perhaps consider renaming to `star_bam` or include a comment here reminding users that only STAR is acceptable.

modules/stringtie/CHANGELOG.md
```diff
@@ -0,0 +1,14 @@
...
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [1.0] - 2021-11-30
+
+This release was authored by Krysta Coyle.
+
+- awk script necessary to introduce XS tags into STAR-aligned BAMs. Without XS tags, stringtie generates single-exon transcripts.
+- GTF mandatory, highly recommended for well-annotated genomes.
```
lkhilton: I'm not sure what "GTF mandatory, highly recommended" means... Is this related to the output format? Is the output format modifiable?

lkhilton (changes requested):
I have a few nit-picky requests. Overall looks excellent and we can probably get it approved and merged today!

modules/stringtie/1.0/stringtie.smk
```diff
@@ -0,0 +1,124 @@
...
+    output:
+        bam = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{sample_id}.bam",
+        bai = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{sample_id}.bam.bai"
+    group:
```
Kdreval, line 61: Thank you Krysta, very good module!!! I would chime into the review suggesting to drop the grouping here since it is not used in the module.

modules/stringtie/1.0/src/tagXSstrandedData.awk
```diff
@@ -0,0 +1,41 @@
+## Authored by Alex Dobin, available at https://github.com/alexdobin/STAR/tree/master/extras/scripts
```
Kdreval, line 1: If this script is not modified, should we consider downloading it within the module, since this is not explicitly authored by us? If you are concerned about possible future script updates that would break the functionality, it is possible tor refer to a specific commit (instead of master) to ensure exact same version is always used

modules/stringtie/1.0/config/default.yaml
kmcoyle: Updated with comment re: STAR bams.

modules/stringtie/1.0/stringtie.smk
kmcoyle: Thanks for pointing that out! Done & added to the demo files.

modules/stringtie/1.0/envs/stringtie.yaml
kmcoyle: Done!

modules/stringtie/1.0/envs/samtools-1.9.yaml
kmcoyle: Thanks for catching it!

modules/stringtie/CHANGELOG.md
kmcoyle: I see how this was confusing. Updated to specify 'reference GTF'.

modules/stringtie/1.0/src/tagXSstrandedData.awk
lkhilton: I thought about this as well but doing it this way is probably the most robust and future-proof and it's still attributed to the repository/author.

modules/stringtie/1.0/stringtie.smk
lkhilton: I actually would recommend keeping the grouping but adding it to the _stringtie_run rule - it's meant to group the fast-running input job with the longer-running job to prevent overwhelming the head node or gphost with more jobs than allowed per user.

Kdreval: yes, sure, thanks Laura!

modules/stringtie/1.0/src/tagXSstrandedData.awk
Kdreval: this makes sense, let's keep it as-is 😄

modules/stringtie/1.0/stringtie.smk
kmcoyle: Moved the grouping to the _stringtie_run rule

lkhilton (approved):

PR #208, opened 2022-01-14
Lcr modules/oncopipe/op fix

mattssca:
# Pull Request
Added patch for handling un-matched normal IDs described in issue [#204](https://github.com/LCR-BCCRC/lcr-modules/issues/204)

Issue #204, opened 2022-01-05
Oncopipe mis-handling unmatched normal IDs


lkhilton:
If the config contains unmatched normal IDs with a seq_type that isn't in the samples table, oncopipe consistently throws an error: 

```
AssertionError in line 40 of /projects/rmorin/projects/gambl-repos/gambl-mutect2-lhilton/src/lcr-modules/modules/svar_master/1.0/svar_master.smk:
There are 0 capture samples matching the normal ID BLGSP-71-06-00286-99A-01D (instead of just one).
  File "/projects/rmorin/projects/gambl-repos/gambl-mutect2-lhilton/src/snakemake/run_svar_master.smk", line 43, in <module>
  File "/projects/rmorin/projects/gambl-repos/gambl-mutect2-lhilton/src/lcr-modules/modules/svar_master/1.0/svar_master.smk", line 40, in <module>
  File "/home/lhilton/miniconda3/lib/python3.7/site-packages/oncopipe/__init__.py", line 1665, in setup_module
  File "/home/lhilton/miniconda3/lib/python3.7/site-packages/oncopipe/__init__.py", line 1243, in generate_runs
```

The `op.generate_runs` function checks for all seq_type unmatched_normal_ids in the config regardless of whether the samples table has been subset to exclude a particular seq_type: 
```
unmatched_normals = dict()
            for key, normal_id in unmatched_normal_ids.items():
                _, genome_build = key.split("--", 1)
                if (
                    not key.startswith(f"{seq_type}--")
                    or genome_build not in sample_genome_builds
                ):
                    continue
                normal_row = samples[
                    (samples.sample_id == normal_id) & (samples.seq_type == seq_type)
                ]
                num_matches = len(normal_row)
                assert num_matches == 1, (
                    f"There are {num_matches} {seq_type} samples matching "
                    f"the normal ID {normal_id} (instead of just one)."
                )
```

One way to patch this would be to generate a list of all `{seq_type}--{genome_build}` combinations that are actually IN the loaded samples table, and ensure there's a key and value for each combination in the config.

mattssca:
The reported bug has been addressed in Pull Request [#208](https://github.com/LCR-BCCRC/lcr-modules/pull/208)

PR #209, opened 2022-01-15
bug fix for unmatched samples

Kdreval:
In the bug fix merged with PR #208 related to issue #204 we tested the update only in demo workflow setup. It worked with no issues, however once tried in GAMBL returned a different error:
```
AssertionError in line 43 of /projects/rmorin/projects/gambl-repos/gambl-kdreval/src/lcr-modules/modules/slms_3/1.0/slms_3.smk:
`run_unpaired_tumours_with` was set to 'unmatched_normal' whereas `unmatched_normal` and `unmatched_normals` were both None. For {seq_type!r}, provide an unmatched normal sample ID. See README for format.
 File "/projects/rmorin/projects/gambl-repos/gambl-kdreval/src/snakemake/run_slms_3_icgc_cram.smk", line 40, in <module>
 File "/projects/rmorin/projects/gambl-repos/gambl-kdreval/src/lcr-modules/modules/slms_3/1.0/slms_3.smk", line 43, in <module>
 File "/home/kdreval/lcr-modules/oncopipe/oncopipe/__init__.py", line 1668, in setup_module
 File "/home/kdreval/lcr-modules/oncopipe/oncopipe/__init__.py", line 1273, in generate_runs
 File "/home/kdreval/lcr-modules/oncopipe/oncopipe/__init__.py", line 1176, in walk_through_dict
 File "/home/kdreval/lcr-modules/oncopipe/oncopipe/__init__.py", line 1176, in walk_through_dict
 File "/home/kdreval/lcr-modules/oncopipe/oncopipe/__init__.py", line 1178, in walk_through_dict
 File "/home/kdreval/lcr-modules/oncopipe/oncopipe/__init__.py", line 1069, in generate_runs_for_patient_wrapper
 File "/home/kdreval/lcr-modules/oncopipe/oncopipe/__init__.py", line 999, in generate_runs_for_patient
```
It turned out that this worked because all samples in demo are matched, and there are no unmatched samples. 
I have tracked this error down and it needed to move the seq_type check to a different place. Since the part of code that evaluated unmatched normal ids
```
            "run_unpaired_tumours_with" in args_dict
            and args_dict["run_unpaired_tumours_with"] == "unmatched_normal"
            and unmatched_normal_ids is not None
            and seq_type in sample_seq_type["seq_type"]
```
is not evaluated because this check returns FALSE, this line never gets run:
`unmatched_normals[key] = Sample(*normal_row.squeeze())`
and is not included in the dictionary generated to store unmatched normals for a given seq_type--genome_build. It later causes the function generate_runs_for_patient, which uses generate_runs within it, to complain of no normal given :mag:
In this PR, the check for seq_type is moved to wrap the `assert` statement and is only evaluated if the seq_type is present in the sample table after subsetting. This has been tested both in demo (adding mock unmatched sample) and in GAMBL set up.

### DIFF
--- a/oncopipe/oncopipe/__init__.py
+++ b/oncopipe/oncopipe/__init__.py
@@ -1253,8 +1253,16 @@ def generate_runs(
                     f"config[‘unmatched_normal_ids’] is not excluded from the samples table"
                     )
                     quit()
-                elif num_matches >= 1:
+                elif num_matches == 1:
                     unmatched_normals[key] = Sample(*normal_row.squeeze())
+                elif num_matches > 1:
+                    print(
+                    f"There are {num_matches} {seq_type} samples matching "
+                    f"the normal ID {normal_id}. This means there are {num_matches} normal samples for {key} in "
+                    f"the samples table and it is not desired. Please ensure all sample_id, seq_type, "
+                    f"and genome_build combinations are unique."
+                    )
+                    quit()
             args_dict["unmatched_normals"] = unmatched_normals
         elif (
             "run_unpaired_tumours_with" in args_dict
--- a/oncopipe/oncopipe/__init__.py
+++ b/oncopipe/oncopipe/__init__.py
@@ -1228,7 +1228,7 @@ def generate_runs(
             "run_unpaired_tumours_with" in args_dict
             and args_dict["run_unpaired_tumours_with"] == "unmatched_normal"
             and unmatched_normal_ids is not None
-            and seq_type in sample_seq_type["seq_type"]
+
         ):
             unmatched_normals = dict()
             for key, normal_id in unmatched_normal_ids.items():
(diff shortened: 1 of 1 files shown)


rdmorin:
If it hasn't already been done can someone create an issue that will address this deficiency in the demo data? i.e. adding an unmatched sample

oncopipe/oncopipe/__init__.py
```diff
@@ -1242,11 +1243,18 @@ def generate_runs(
...
+                    print(
                     f"There are {num_matches} {seq_type} samples matching "
-                    f"the normal ID {normal_id} (instead of just one)."
-                )
-                unmatched_normals[key] = Sample(*normal_row.squeeze())
+                    f"the normal ID {normal_id}. Make sure the default unmatched normal for {key} specified in "
+                    f"config[‘unmatched_normal_ids’] is not excluded from the samples table"
+                    )
+                    quit()
+                elif num_matches >= 1:
```
lkhilton: Are you 100% sure there's no way this will break in changing it from the current `==1` condition? I would be more comfortable if this was `elif num_matches ==  1` and there was an `else` statement for if num_matches is not 0 or 1.

Kdreval: I'll add that `>1` scenario here, sure. Do we want it to exit with an error or give a warning? I think the error is preferable, since this would mean there are several samples with the same sample_id and seq_type--genome build mapping, and also error will be consistent with existing functionality.

mattssca: @lkhilton suggestion has my vote as well, just to be safe. An error message seems to be the more appropriate alternative.

lkhilton: I agree an error is better! Thanks.

rdmorin: Better to never assume we can't get a certain condition and always handle it. Python exceptions when not caught are often useless.

Kdreval: Thank you all for the suggestions! The check for `>1` is now added and will result in the error message like this:
```
There are 2 capture samples matching the normal ID TCRBOA7-N-WEX. This means there are 2
normal samples for capture--grch37 in the samples table and it is not desired. Please
ensure all sample_id, seq_type, and genome_build combinations are unique.
```
This update was tested to behave normally in GAMBL 🙌

lkhilton (approved):

PR #212, opened 2022-01-22
Jawong module/cnvkit/1.0

Jwong684:
# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable

Example:
```
```


modules/cnvkit/1.0/cnvkit.smk
```diff
@@ -0,0 +1,318 @@
...
+    _cnvkit_all,
+
+
+##### RULES #####
+if CFG["options"]["hard_masked"] == True:
+    CFG["runs"]["masked"] = "_masked"
+else:
+    CFG["runs"]["masked"] = ""
+    
+wildcard_constraints:
```
Kdreval: One of the PR checklist items is specifically not to set the global wildcards - I think Laura found they were causing issues with reference_files at some points
```
 I did not set any global wildcard constraints. Any/all wildcard constraints are set on a per-rule basis.
```
Can you please move these to be within the rule?

modules/cnvkit/1.0/cnvkit.smk
```diff
@@ -0,0 +1,318 @@
...
+    genome_build = ".+(?<!masked)"
+    
+    
+# cnvkit reference files
+# gene annotation files
+rule _get_refFlat:
+    output:
+        refFlat = CFG["dirs"]["inputs"] + "ref/refFlat.txt"
+    params:
+        url = lambda w: {
```
Kdreval: Can this be implemented in a more generalized way? You can build url from several strings, something similar to this: https://github.com/LCR-BCCRC/lcr-modules/blob/82439d6721d0b1587cd16200d19a7bfc1f4a27e8/modules/rainstorm/1.0/rainstorm.smk#L63. 
or this example:
https://github.com/LCR-BCCRC/lcr-modules/blob/82439d6721d0b1587cd16200d19a7bfc1f4a27e8/modules/sage/1.0/sage.smk#L101
Here is another example, too:
https://github.com/LCR-BCCRC/lcr-modules/blob/82439d6721d0b1587cd16200d19a7bfc1f4a27e8/modules/svar_master/1.0/svar_master.smk#L70
This will help to avoid "hard-coding" genome builds, as there are now lots supported by lcr-modules directly and even more of GAMBL-specific are coming.

modules/cnvkit/1.0/cnvkit.smk
```diff
@@ -0,0 +1,318 @@
...
+        crai = CFG["dirs"]["inputs"] + "{seq_type}--{genome_build}/{capture_space}/{sample_id}.bam.crai"
+    run:
+        op.relative_symlink(input.bam, output.bam)
+        op.relative_symlink(input.bai, output.bai)
+        op.relative_symlink(input.bai, output.crai)
+
+
+def get_normals(wildcards):
+    CFG = config["lcr-modules"]["cnvkit"]
+    capture_space = CFG["runs"][CFG["runs"]["normal_capture_space"].isin([wildcards.capture_space])]
```
Kdreval, line 101: What is with capture_space column here? Does this mean that the capture space column is required in the sample table? The general idea in the capture support for other modules was not to use the name of the space as a wildcard - because it can expand to wary greatly as more samples are added, creating inconsistent naming patterns for downstream tools. Is the goal here to get list of sample ids corresponding to a specific subset of samples?

modules/cnvkit/1.0/cnvkit.smk
```diff
@@ -0,0 +1,318 @@
...
+rule _run_cnvkit_batch:
+    input:
+        samples = expand(CFG["dirs"]["inputs"] + "{seq_type}--{genome_build}/{capture_space}/{tumour_id}.bam", 
+                        zip,
+                        tumour_id = CFG["runs"]["tumour_sample_id"],
+                        capture_space = CFG["runs"]["tumour_capture_space"],
+                        genome_build = CFG["runs"]["tumour_genome_build"],
+                        seq_type = CFG["runs"]["tumour_seq_type"]),
+        controls = get_normals,
+        refFlat = CFG["dirs"]["inputs"] + "ref/refFlat.txt",
```
Kdreval: The first rule seems to download this in the gzipped format - any potential problems with this? Maybe, refer here to the `rules._get_refFlat.outout.refFlat`?

modules/cnvkit/1.0/cnvkit.smk
```diff
@@ -0,0 +1,318 @@
...
+        opts = CFG["options"]["batch"]
+    conda:
+        CFG["conda_envs"]["cnvkit"]
+    threads:
+        CFG["threads"]["batch"]
+    resources:
+        **CFG["resources"]["batch"]
+    log: 
+        stdout = CFG["logs"]["cnvkit"] + "{seq_type}--{genome_build}{masked}/{capture_space}_cnr.log"
+    shell:
```
Kdreval, line 149: can this please be converted to op.as_one_line, so it is easier to follow the arguments? Thanks!

modules/cnvkit/1.0/cnvkit.smk
```diff
@@ -0,0 +1,318 @@
...
+    output:
+        bed = CFG["dirs"]["SNPs"] + "{genome_build}/dbsnp.common_all-151.bed"
+    resources: **CFG["resources"]["SNPs"]
+    shell:
+        op.as_one_line(""" gunzip -c {input.vcf} | awk {{'printf ("%s\\t%s\\t%s\\n", $1,$2-1,$2)'}} | zgrep -v -h "^#" > {output.bed} """)
+
+#### set-up mpileups for BAF calling ####
+def _cnvkit_get_chr_mpileups(wildcards):
+    CFG = config["lcr-modules"]["cnvkit"]
+    chrs = reference_files("genomes/" + wildcards.genome_build + "/genome_fasta/main_chromosomes_withY.txt")
```
Kdreval, line 293: I see it uses chromosomes with Y here - does the sex of the patient matter for the cnvkit? If so, how does it determine it and what happens when it is not available? Should we consider dropping Y chr here and stick with X similar to other CNV modules?

modules/cnvkit/1.0/cnvkit.smk
```diff
@@ -0,0 +1,318 @@
...
+        CFG["dirs"]["SNPs"] + "{{seq_type}}--{{genome_build}}/{{capture_space}}/{{sample_id}}.{chrom}.vcf.gz.tbi", 
+        chrom = chrs
+    )
+    return(mpileups)
+
+# vcf needs DP, GT, AD - bcftools -mv calls multiallelic variants (will annotate GT)
+# without it, GT will not be annotated
+rule _cnvkit_mpileup_per_chrom:
+    input:
+        bam = CFG["dirs"]["inputs"] + "{seq_type}--{genome_build}/{capture_space}/{sample_id}.bam",
```
Kdreval, line 317: here and in other rules, can the inputs be used as `str(rules...output...)` snakemake objects instead of referring to them directly?

modules/cnvkit/1.0/cnvkit.smk
```diff
@@ -0,0 +1,318 @@
...
+    input:
+        expand(
+            [
+                CFG["dirs"]["BAF"] + "{{seq_type}}--{{genome_build}}{{masked}}/{capture_space}/{tumour_id}.call.cns"
+            ],
+            zip,  # Run expand() with zip(), not product()
+            tumour_id=CFG["runs"]["tumour_sample_id"],
+            capture_space=CFG["runs"]["tumour_capture_space"])
+    output:
+        CFG["dirs"]["outputs"] + "seg/{seq_type}--{genome_build}{masked}/cnvkit.seg"
```
Kdreval: so it generates 1 seg/sample set (not individual sample)?

modules/cnvkit/1.0/cnvkit.smk
```diff
@@ -0,0 +1,318 @@
...
+                str(rules._cnvkit_output.output.scatter),
+                str(rules._cnvkit_to_seg.output)
+            ],
+            zip,  # Run expand() with zip(), not product()
+            seq_type=CFG["runs"]["tumour_seq_type"],
+            genome_build=CFG["runs"]["tumour_genome_build"],
+            tumour_id=CFG["runs"]["tumour_sample_id"],
+            normal_id=CFG["runs"]["normal_sample_id"],
+            pair_status=CFG["runs"]["pair_status"],
+            capture_space=CFG["runs"]["tumour_capture_space"],
```
Kdreval, line 659: Can we avoid using `capture_space` as a wildcard?

modules/cnvkit/1.0/config/default.yaml
```diff
@@ -0,0 +1,64 @@
...
+
+        # TODO: Update the list of available wildcards, if applicable
+        inputs:
+            # Available wildcards: {seq_type} {genome_build} {sample_id}
+            sample_bam: "__UPDATE__"
+            sample_bai: "__UPDATE__"
+
+        scratch_subdirectories: []
+
+        options:
```
Kdreval, line 14: Can you please add in-line documentation for these options? What do they mean/how to use them?

modules/cnvkit/1.0/config/default.yaml
```diff
@@ -0,0 +1,64 @@
...
+        pairing_config:
+            genome:
+                run_paired_tumours: True
+                run_unpaired_tumours_with: "unmatched_normal"
+                run_paired_tumours_as_unpaired: False
+            capture:
+                run_paired_tumours: True
+                run_unpaired_tumours_with: "unmatched_normal"
+                run_paired_tumours_as_unpaired: False
+            exome:
```
Kdreval: I don't think this a supported seq_type in the lcr-modules. The variant callers include the exome and capture within only `capture` seq_type, can we do it here as well for consistency?

modules/cnvkit/1.0/envs/cnvkit-0.9.7.yaml
```diff
@@ -0,0 +1,120 @@
+name: null
```
Kdreval: Can you please put this `.yaml` in the `lcr-modules/envs` and place here a symlink instead, just like for bcftools?

Kdreval (changes requested):
Thank you so much Jasper! Amazing to see it here, will greatly improve the existing functionality! I left some comments and thoughts

modules/cnvkit/1.0/cnvkit.smk
Jwong684: Yes, it is to gather all normals that share the same capture space. CNVkit will run in batches based on a panel of normals you provide and a series of samples that would use the normals. So to separate analyses based on capture space, I have it as a wildcard. If capture_space is not a good column to use in the samples table, is there a better alternative to find samples that share the same capture kit?

modules/cnvkit/1.0/cnvkit.smk
Jwong684: Done

modules/cnvkit/1.0/cnvkit.smk
Jwong684: Done

modules/cnvkit/1.0/cnvkit.smk
Kdreval: What does it do differently for the different capture space? Say, if there is a target panel of 100 genes - what will happen with the cnvkit results? I am thinking if we can apply the genome_build here instead. You can also take a look at how the lofreq is now implemented: it uses a switch on wildcard and a default bed/genome guild. If no bed file is provided, it will use the default bed - and if there is a bed file given, lofreq will use that. Maybe, we can implement something similar here? That function returns the path to bed file, but can give an idea how to get the list of samples
https://github.com/LCR-BCCRC/lcr-modules/blob/82439d6721d0b1587cd16200d19a7bfc1f4a27e8/modules/lofreq/1.1/lofreq.smk#L61-L81

modules/cnvkit/1.0/cnvkit.smk
Jwong684: good catch - I added to the _get_refFlat rule to uncompress the gzip file

Also added a configuration in there to automatically remove  the chr prefix if genome build is grch37 or grch38

modules/cnvkit/1.0/cnvkit.smk
Jwong684: Done

modules/cnvkit/1.0/cnvkit.smk
Jwong684: This is a good catch. CNVkit can apparently determine sex just based on the coverage on the X and Y in the sample/normal. There is a flag -y to assume a male reference and (without -y , the default) would be for a female reference. The only difference with the flag is just the adjustments to the chrX and chrY copy numbers - without -y flag, a male would have log.ratio -1 for chrX and 0 for chr Y, while a female would have 0 for chrX and -Inf for chrY.

I think since cnvkit has that as an output already, we may as well keep it in. If we need to keep it consistent with other CNV modules, we can just simply remove all chrY calls in the final seg file.

modules/cnvkit/1.0/cnvkit.smk
Jwong684: Done

modules/cnvkit/1.0/cnvkit.smk
Jwong684: yes cnvkit has an option to just neatly compile all the samples into one seg

Kdreval: Is it possible to have individual seg-files? Imagine adding these to MySQL database, or having some downstream tool that needs a path to a per-sample file, this may be hard to deal with

modules/cnvkit/1.0/config/default.yaml
Jwong684: Done

modules/cnvkit/1.0/config/default.yaml
Jwong684: Okay I removed "exome" from the module

modules/cnvkit/1.0/envs/cnvkit-0.9.7.yaml
Jwong684: Even for the module-specific yaml file?

Kdreval: Yep, for consistency

modules/cnvkit/1.0/cnvkit.smk
Jwong684: Interestingly, this seems to shoot an error back when snakemake tries to figure out tumour_ids from sample_ids in the wildcards. Reverting back

modules/cnvkit/CHANGELOG.md
```diff
@@ -0,0 +1,32 @@
...
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [1.0] - 2022-01-22
+
+This release was authored by Jasper Wong.
+
+Cnvkit uses both on-target and off-target bait regions to determine CNV.
+
+Everything is bundled under one function "batch", which builds a normal reference from a panel of normals and calls CN for each tumour.
```
Kdreval, line 36: Thanks Jasper! Can you please update CHANGELOG to reflect the recent overhaul and I'll merge it to master 🙏

Jwong684: Okay done

Kdreval (approved):
Thanks Jasper!!

PR #213, opened 2022-01-26
Updated handling of VCF reference files

ckrushton:
VCF files will now automatically update the VCF header using the same contigs in the destination reference genome build
VCF entries (i.e. variants) will now be restricted to canonical contigs

### DIFF
--- a/workflows/reference_files/2.4/reference_files.smk
+++ b/workflows/reference_files/2.4/reference_files.smk
@@ -232,15 +232,18 @@ rule calc_gc_content:
 
 rule get_dbsnp_download: 
     input:
-        vcf = get_download_file(rules.download_dbsnp_vcf.output.vcf)
+        vcf = get_download_file(rules.download_dbsnp_vcf.output.vcf),
+        fai = str(rules.index_genome_fasta.output.fai),
+        bed = str(rules.get_main_chromosomes_withY_download.output.bed)
     output:
-        vcf = "genomes/{genome_build}/variation/dbsnp.common_all-{dbsnp_build}.vcf.gz"
-    conda: CONDA_ENVS["samtools"]
+        vcf = "genomes/{genome_build}/variation/dbsnp.common_all-{dbsnp_build}.vcf.gz",
+        tmpfile = temp("genomes/{genome_build}/variation/dbsnp.common_all-{dbsnp_build}.vcf.tmp")
+    conda: CONDA_ENVS["bcftools"]
     shell:
         op.as_one_line("""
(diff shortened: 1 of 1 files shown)


workflows/reference_files/2.4/reference_files.smk
```diff
@@ -232,15 +232,18 @@ rule calc_gc_content:
...
+        tmpfile = temp("genomes/{genome_build}/variation/dbsnp.common_all-{dbsnp_build}.vcf.tmp")
+    conda: CONDA_ENVS["bcftools"]
     shell:
         op.as_one_line("""
-        bgzip -c {input.vcf} > {output.vcf}
-            &&
-        tabix {output.vcf}
+        zgrep -v '##contig' {input.vcf} > {output.tmpfile} &&
+        bcftools reheader --fai {input.fai} {output.tmpfile} | bcftools view -T {input.bed} -O z -o {output.vcf} &&
+        bcftools index -t {output.vcf}
```
Kdreval, line 246: A quick question: was this tested in the SLMS-3? I remember some tools complaining about index not being produced by tabix - but maybe it is the other way around and it was about not being generated by bcftools. In any case, I wonder if at least demo capture sample was run through this

Kdreval (commented):
Thanks Chris! Just a question about whether it was tested, at least on the demo sample?

workflows/reference_files/2.4/reference_files.smk
Kdreval: Right, it was about bcftools indices not being compatible with downstream tools, so that is why tabix is used in the sage module:
https://github.com/LCR-BCCRC/lcr-modules/blob/82439d6721d0b1587cd16200d19a7bfc1f4a27e8/modules/sage/1.0/sage.smk#L234-L243

It should be also available in the same environment, I think

ckrushton: I'll test this, but `bcftools index -t` claims to generate a tabix index. I'll run the demo workflow quickly just to confirm

Kdreval: Thanks! Sorry bothering you with it, I just remember having troubles with this. It's fine to leave as is as long as everything works

ckrushton: Everything seems to work with MuTect2. The VCF files were parsed correctly with no complaints

Kdreval: Thanks!

Kdreval (approved):

PR #217, opened 2022-02-24
Updated CrossMap and chr-prefix handling; improved support for non-standard genome builds

ckrushton:
-All input VCFs are now annotated and converted to the desired genome builds and variants, adding or removing chr-prefixes as necessary. Note that output files which originated from CrossMap are placed in a different output directory. Variants and genome build "groups" are determined based on the reference config
- Updated handling of custom ENST ID files, to only require one per genome build type (ex. GRCh38 and GRCh37)
- Updated handling of unique genome builds (ex. hg19-reddy)

# Pull Request Checklists

Example:
```
```

### DIFF
--- a/modules/vcf2maf/1.3/vcf2maf.smk
+++ b/modules/vcf2maf/1.3/vcf2maf.smk
@@ -40,7 +40,7 @@ VCF2MAF_SCRIPT_PATH = CFG['inputs']['src_dir']
 # 1) Which iteration of the genome build something is (i.e. GRch37 or GRCh38)
 # 2) If a genome build is chr-prefixed or not
 # To determine this, lets load the reference config and re-parse it
-VCF2MAF_REFERENCE_CONFIG = CFG["options"]["reference_config"]  # TEMP, remove later
+VCF2MAF_REFERENCE_CONFIG = CFG["options"]["reference_config"]  # Reference config path
 configfile: VCF2MAF_REFERENCE_CONFIG
 # Store all the attributes we will need
 VCF2MAF_GENOME_VERSION = {}  # Will be a simple {"GRCh38-SFU": "grch38"} etc.
@@ -165,11 +165,12 @@ def get_original_genome(wildcards):
     # Determine the original (i.e. input) reference genome for this sample
     # Since this module projects to various output genome builds, we need to parse the sample table for the starting build
     # To determine what we need to do
-    runs_table = config['lcr-modules']["vcf2maf"]["runs"]
-    sample_entry = runs_table.loc[(runs_table["tumour_sample_id"] == wildcards.tumour_id) & (runs_table["normal_sample_id"] == wildcards.normal_id) & (runs_table["tumour_seq_type"] == wildcards.seq_type)]
+    print(config['lcr-modules']["vcf2maf"]["samples"].columns)
+    sample_table = config['lcr-modules']["vcf2maf"]["samples"]
+    sample_entry = sample_table.loc[(sample_table["sample_id"] == wildcards.tumour_id) & (sample_table["seq_type"] == wildcards.seq_type)]
     if len(sample_entry) == 0:
-        raise AttributeError("Unable to locate a a sample with tumour_id:{wildcards.tumour_id}, normal_id:{wildcards.normal_id}, seq_type:{wildcards.seq_type} in the \'runs\' table")
-    original_genome_build = sample_entry.iloc[0]["tumour_genome_build"]
+        raise AttributeError("Unable to locate a a sample with tumour_id:{wildcards.tumour_id}, normal_id:{wildcards.normal_id}, seq_type:{wildcards.seq_type} in the \'sample\' table")
+    original_genome_build = sample_entry.iloc[0]["genome_build"]
     return original_genome_build
 
 def get_chain(genome_build):
@@ -220,7 +221,7 @@ def get_normalize_input(wildcards, genome_build_only = False):
     original_genome_version = VCF2MAF_GENOME_VERSION[original_genome_build]
     new_genome_version = VCF2MAF_GENOME_VERSION[new_genome_build]
 
-    # If using this function as an input function for snakemake
+    # If using this as an input function for snakemake
     if not genome_build_only:
         # Do we need to run CrossMap on this? Check the genome version
         if original_genome_version != new_genome_version:
--- a/modules/vcf2maf/1.3/vcf2maf.smk
+++ b/modules/vcf2maf/1.3/vcf2maf.smk
@@ -28,8 +28,7 @@ CFG = op.setup_module(
 localrules:
     _vcf2maf_input_vcf,
     _vcf2maf_gnomad_filter_maf,
-    _vcf2maf_output_maf_native,
-    _vcf2maf_output_maf_projection,
+    _vcf2maf_output_maf,
     _vcf2maf_crossmap,
     _vcf2maf_all
 
@@ -147,7 +146,7 @@ rule _vcf2maf_gnomad_filter_maf:
     input:
         maf = str(rules._vcf2maf_run.output.maf)
     output:
-        maf = CFG["dirs"]["vcf2maf"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/{base_name}.gnomad_filtered.maf",
+        maf = temp(CFG["dirs"]["vcf2maf"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/{base_name}.gnomad_filtered.maf"),
         dropped_maf = CFG["dirs"]["vcf2maf"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/{base_name}.gnomad_filtered.dropped.maf.gz"
     params:
         opts = CFG["options"]["gnomAD_cutoff"],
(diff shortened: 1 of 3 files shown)


modules/vcf2maf/CHANGELOG.md
```diff
@@ -5,6 +5,15 @@ All notable changes to the `vcf2maf` module will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.3] - 2022-02-23
+
+This release was kind of authored by Christopher Rushton. Someone else originally authored this release
+
+- Now supports automatic CrossMap/chr prefix changing for target genome builds for all input VCF files. For instance, an input VCF called against hg38-nci will now be used to generate a MAF file against both hg38 (no changes, chr-prefixed), GRCh38 (remove chr prefix), hg19 (CrossMap and chr-prefixed) and GRCh37 (CrossMap and chr-prefix). The reference workflow config is used to determine the "same" and "different" genome versions and chr prefixes. Also handles cases where CrossMap adds extra chr-prefixes to already prefixed output files, and updates the NCBI_Build column of the output MAF file.
+- Improved handling of non-standard genome versions
+- Updated custom ENST ID files to only require one per genome build version ("grch38" and "grch37")
```
Kdreval, line 14: you meant hg38 here, right?

modules/vcf2maf/1.3/vcf2maf.smk
```diff
@@ -15,31 +15,55 @@ import sys, os
...
-}
-
 #set variable for prepending to PATH based on config
 VCF2MAF_SCRIPT_PATH = CFG['inputs']['src_dir']
 
+# This is going to be annoying, but to normalize prefixes, we need to determine
+# 1) Which iteration of the genome build something is (i.e. GRch37 or GRCh38)
+# 2) If a genome build is chr-prefixed or not
+# To determine this, lets load the reference config and re-parse it
+VCF2MAF_REFERENCE_CONFIG = CFG["options"]["reference_config"]  # TEMP, remove later
```
Kdreval: what do you mean by TEMP here? Where is it getting removed?

modules/vcf2maf/1.3/vcf2maf.smk
```diff
@@ -83,8 +107,8 @@ rule _vcf2maf_run:
         stderr = CFG["logs"]["vcf2maf"] + "{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}/{base_name}_vcf2maf.stderr.log",
     params:
         opts = CFG["options"]["vcf2maf"],
-        build = lambda w: VCF2MAF_GENOME_VERSION_MAP[w.genome_build],
-        custom_enst = op.switch_on_wildcard("genome_build", CFG["switches"]["custom_enst"])
+        build = lambda w: VCF2MAF_VERSION_MAP[w.genome_build],
+        custom_enst = lambda w: config['lcr-modules']["vcf2maf"]["switches"]["custom_enst"][VCF2MAF_GENOME_VERSION[w.genome_build]]
```
Kdreval, line 110: I am not sure I follow how this is tracking with module config and switches. There is no key for example for `grch38` in config. Are these updates tested/all updates are pushed?

modules/vcf2maf/1.3/vcf2maf.smk
```diff
@@ -137,66 +161,162 @@ rule _vcf2maf_gnomad_filter_maf:
...
+    sample_entry = runs_table.loc[(runs_table["tumour_sample_id"] == wildcards.tumour_id) & (runs_table["normal_sample_id"] == wildcards.normal_id) & (runs_table["tumour_seq_type"] == wildcards.seq_type)]
+    if len(sample_entry) == 0:
+        raise AttributeError("Unable to locate a a sample with tumour_id:{wildcards.tumour_id}, normal_id:{wildcards.normal_id}, seq_type:{wildcards.seq_type} in the \'runs\' table")
+    original_genome_build = sample_entry.iloc[0]["tumour_genome_build"]
+    return original_genome_build
+
+def get_chain(genome_build):
+    # NOTE: This only currently supports hg38 and hg19. If you are using other genome builds, this will need to be handled
+    if VCF2MAF_GENOME_VERSION[genome_build] == "grch38":
+        return reference_files("genomes/" + genome_build + "/chains/grch38/hg38ToHg19.over.chain")
```
Kdreval, line 178: does this work if no reference files are present and need to be generated _de novo_?

modules/vcf2maf/1.3/vcf2maf.smk
```diff
@@ -137,66 +161,162 @@ rule _vcf2maf_gnomad_filter_maf:
...
         """)
 
-def get_chain(wildcards):
-    if "38" in str({wildcards.genome_build}):
-        return reference_files("genomes/{genome_build}/chains/grch38/hg38ToHg19.over.chain")
+def get_original_genome(wildcards):
+    # Determine the original (i.e. input) reference genome for this sample
+    # Since this module projects to various output genome builds, we need to parse the sample table for the starting build
+    # To determine what we need to do
+    runs_table = config['lcr-modules']["vcf2maf"]["runs"]
```
Kdreval: I just wonder if the sample table can be used here instead? Probably not so important, but using a sample table will allow to not use the normal id match and therefore decrease the number of checks. Pandas are very slow for large tables and it may take a long time for all of this to be computed if we try this even on the GAMBL-wide scale.

modules/vcf2maf/1.3/vcf2maf.smk
```diff
@@ -137,66 +161,162 @@ rule _vcf2maf_gnomad_filter_maf:
...
-            pair_status = CFG["runs"]["pair_status"],
-            base_name = [CFG["vcf_base_name"]] * len(CFG["runs"]["tumour_sample_id"]))
+        expand(
+            expand(str(rules._vcf2maf_output_dispatch.output.dispatch), zip,
+                seq_type = CFG["runs"]["tumour_seq_type"],
+                tumour_id = CFG["runs"]["tumour_sample_id"],
+                normal_id = CFG["runs"]["normal_sample_id"],
+                pair_status = CFG["runs"]["pair_status"],
+                base_name = [CFG["vcf_base_name"]] * len(CFG["runs"]["tumour_sample_id"]),
+                allow_missing = True),
```
Kdreval, line 295: what is missing here?

modules/vcf2maf/1.3/vcf2maf.smk
```diff
@@ -137,66 +161,162 @@ rule _vcf2maf_gnomad_filter_maf:
...
+            target_build = CFG["options"]["target_builds"])
```
Kdreval, line 296: Can you please show an example of what does the output path for each genome build looks like now?

Kdreval (commented):
Thanks @ckrushton ! Few minor questions

modules/vcf2maf/CHANGELOG.md
ckrushton: I was using "grch38" and "grch37" to be consistent with the "build" field in the reference config, but I can change it to "hg38" and "hg19"

modules/vcf2maf/1.3/vcf2maf.smk
ckrushton: Sorry I forgot to delete this comment (originally this was a hard-coded file path for testing purposes)

modules/vcf2maf/1.3/vcf2maf.smk
ckrushton: I changed the vcf2maf config in gambl, but not in the module itself. Will update

modules/vcf2maf/1.3/vcf2maf.smk
ckrushton: I haven't explicitly tested this, but it should work. At the end of the day the subworkflow is just getting a string for a desired file; this shouldn't matter if this results from wildcard expansion or string concatenation

modules/vcf2maf/1.3/vcf2maf.smk
ckrushton: I can take a look, or do some performance optimization

modules/vcf2maf/1.3/vcf2maf.smk
ckrushton: This is missing the wildcard `target_build`, so it remains unexpanded (used by the outer expand function)

modules/vcf2maf/1.3/vcf2maf.smk
ckrushton: 99 outputs will generate the following subfolders:
`<seq_type>--<genome_build>
<seq_type>--<genome_build>--projection`
For instance, if a VCF was originally referenced against hg38, the hg38 and grch38 MAFs will be placed in `<seq_type>--<genome_build>`, while the hg19 and grch37 MAFs will be placed in `<seq_type>--<genome_build>--projection`

rdmorin: We need them all in the SAME directory per projection regardless of their parental genome_build

ckrushton: Done and implemented as of current commit

Kdreval (approved):

rdmorin:
Are we 100% ready to merge?

Kdreval:
I think so - unless there are any other edits/updates that were not pushed. @ckrushton can you please make sure everything is pushed, so we don't repeat the capture support PR

ckrushton:
Everything is pushed on my end

PR #220, opened 2022-03-09
fastq accommodation: bam2fastq and tiny star fix

hayashaalan:
This PR includes:
1) Changes to constrain rule bam2fastq_run to only samples with bam files. The goal is to disregard samples that only have fastq files. 
2) Minor change to the star and star_grouped snakefiles to return the variable "overhang" as an integer rather than a float (149 vs 149.0). Returning a float triggers the creation of already existing star indices.

# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable

Example:
```
```

### DIFF
--- a/modules/bam2fastq/1.2/bam2fastq.smk
+++ b/modules/bam2fastq/1.2/bam2fastq.smk
@@ -43,7 +43,7 @@ CFG = op.setup_module(
     version = "1.2",
     subdirectories = ["inputs", "fastq", "outputs"],
 )
-
+print(CFG["samples"]["sample_id"].tolist())
 # Define rules to be run locally when using a compute cluster
 localrules:
     _bam2fastq_input_bam,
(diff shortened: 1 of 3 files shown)


Kdreval (approved):
Thanks @hayashaalan , looks good to me!

modules/bam2fastq/1.2/bam2fastq.smk
```diff
@@ -93,6 +93,8 @@ if CFG["temp_outputs"]:
         log:
             stdout = CFG["logs"]["fastq"] + "{seq_type}/{sample_id}/bam2fastq.stdout.log",
             stderr = CFG["logs"]["fastq"] + "{seq_type}/{sample_id}/bam2fastq.stderr.log"
+        wildcard_constraints:
+            sample_id = "|".join(mrna_bams)
```
lkhilton: Where is mrna_bams defined in the module? This should probably instead be `"|".join(sample_id=CFG["samples"]["sample_id"].tolist())`

lkhilton (changes requested):

modules/bam2fastq/1.2/bam2fastq.smk
hayashaalan: I had it specified in the header, but I should've put it in the module. I can replace it with what you wrote above ^.

lkhilton (approved):

PR #222, opened 2022-03-18
Augment ssm

lkhilton:
# Pull Request Checklists

These updates incorporate blacklisting and augmenting into the vcf2maf module and minimizes data duplication through the creation of symlinks where file contents are identical.


modules/vcf2maf/1.3/config/default.yaml
```diff
@@ -7,7 +7,15 @@ lcr-modules:
             # Available wildcards: {out_dir}, {seq_type}, {genome_build}, {tumour_id}, {normal_id}, {pair_status}, {vcf_name}, {base_name}
             vep_cache: "__UPDATE__" #example "ref/ensembl_vep_cache/"
             sample_vcf_gz: "__UPDATE__" #full path to your compressed vcfs from favourite variant caller (with wildcards. You must encapsulate everything unique to the variant caller naming scheme in {base_name})
+            sample_bam: "__UPDATE__"
```
rdmorin, line 10: Should these lines have some documentation beside them like the others do?

modules/vcf2maf/1.3/config/default.yaml
```diff
@@ -31,19 +45,31 @@ lcr-modules:
...
+            deblacklist: 1
+            augment: 12
         
         resources:
             vcf2maf: 
                 mem_mb: 12000
                 vcf: 1
+            augment: 
+                mem_mb: 12000
+                augment: 1
```
rdmorin, line 63: What does the augment: 1 do?

modules/vcf2maf/1.3/src/deblacklist_ssm.R
```diff
@@ -0,0 +1,42 @@
...
+project_base = snakemake@params[["project_base"]]
+seq_type_param = snakemake@params[["seq_type_blacklist"]]
+drop_threshold = snakemake@params[["threshold"]]
+
+maf = snakemake@input[["maf"]]
+
+seq_type = if (seq_type_param == "both") c("genome", "capture") else snakemake@wildcards[["seq_type"]]
+
+fread_maf(maf) %>% 
+    mutate(Chromosome = as.character(Chromosome)) %>%
```
rdmorin, line 31: I had thought fread_maf already gives you a character chromosome. Were there exceptions?

modules/vcf2maf/1.3/src/deblacklist_ssm.R
```diff
@@ -0,0 +1,42 @@
...
+
+log <- file(snakemake@log[[1]], open="wt")
+sink(stdout(), type = "message")
+sink(log)
+
+library(GAMBLR)
+library(data.table)
+library(tidyverse)
+library(readr)
+library(glue)
```
rdmorin, line 13: Should I add these missing dependencies to the GAMBLR function instead?

modules/vcf2maf/1.3/vcf2maf.smk
```diff
@@ -166,6 +185,140 @@ rule _vcf2maf_gnomad_filter_maf:
         touch {output.dropped_maf}
         """)
 
+# Obtain the path to the GAMBLR conda environment
+md5hash = hashlib.md5()
```
rdmorin, line 190: Sneaky!

modules/vcf2maf/1.3/vcf2maf.smk
```diff
@@ -166,6 +185,140 @@ rule _vcf2maf_gnomad_filter_maf:
...
+            .groupby(["tumour_patient_id", "tumour_seq_type", "tumour_genome_build"])\
+            .filter(lambda x: len(x) > 1)\
+            .reset_index()
+    else: 
+        MULTI_SAMPLES = CFG["runs"]\
+            .groupby(["tumour_patient_id", "tumour_genome_build"])\
+            .filter(lambda x: len(x) > 1)\
+            .reset_index()
+else: 
+    MULTI_SAMPLES = op.discard_samples(CFG["runs"], tumour_sample_id = CFG["runs"]["tumour_sample_id"].tolist())
```
rdmorin, line 282: Is this just discarding everything?

rdmorin (commented):
Very nice enhancement!

modules/vcf2maf/1.3/vcf2maf.smk
```diff
@@ -166,6 +185,140 @@ rule _vcf2maf_gnomad_filter_maf:
...
+    conda_prefix = os.path.abspath(".snakemake/conda")
+
+md5hash.update(conda_prefix.encode())
+f = open("config/envs/GAMBLR.yaml", 'rb')
+md5hash.update(f.read())
+f.close()
+h = md5hash.hexdigest()
+GAMBLR = glob.glob(conda_prefix + "/" + h[:8] + "*")[0]
+
+rule _vcf2maf_install_GAMBLR:
```
Kdreval, line 203: I think this is missing from the local rules directive above, just want to confirm if this is intended

modules/vcf2maf/1.3/config/default.yaml
lkhilton: Probably not necessary, they're consistent across modules and the allowed wildcards are described above.

modules/vcf2maf/1.3/config/default.yaml
lkhilton: It's just a resource name, it lets you control how many of each step you want run simultaneously on the cluster.

lkhilton: It's just a resource name, it lets you control how many of each step you want run simultaneously on the cluster.

modules/vcf2maf/1.3/src/deblacklist_ssm.R
lkhilton: I think this is how Haya had it in her version of this script in the GAMBL repo. I'm not sure if she ran into issues.

modules/vcf2maf/1.3/src/deblacklist_ssm.R
lkhilton: Sure!

modules/vcf2maf/1.3/vcf2maf.smk
lkhilton: 😜

modules/vcf2maf/1.3/vcf2maf.smk
lkhilton: Yes, it was just an easy way to create an empty data frame with all the same column headers as the samples table.

modules/vcf2maf/1.3/config/default.yaml
lkhilton: I will add a note about this to the top of the Snakefile, which is something we've been doing in other modules lately.

modules/vcf2maf/1.3/vcf2maf.smk
```diff
@@ -222,88 +383,169 @@ rule _vcf2maf_crossmap:
...
+                str(rules._vcf2maf_output_original.output.maf), 
+                zip, 
+                seq_type = CFG["runs"]["tumour_seq_type"], 
+                genome_build = CFG["runs"]["tumour_genome_build"], 
+                tumour_id = CFG["runs"]["tumour_sample_id"], 
+                normal_id = CFG["runs"]["normal_sample_id"], 
+                pair_status = CFG["runs"]["pair_status"], 
+                allow_missing = True
+            ),
+            maf = "maf",
```
Kdreval, line 499: here and below `maf = "maf"` seems to be one element. I wonder how this works, especially where there is expand on at least two items: `filter = CFG["options"]["filter"]`  is set to `filter: ["raw", "deblacklisted"]` so I would think the maf  needs to be something like `maf = ["maf"]*length(CFG["options"]["filter"])` to generate 2 files, in this case `raw/maf` and `deblacklisted/maf`. Otherwise if the length is 1, I expect it to produce only the first instance after expansion, so only `raw/maf`. Asking because this might help me to wrap my head around the CNV merging 😄

modules/vcf2maf/1.3/vcf2maf.smk
lkhilton: Good catch, should definitely be a localrule

modules/vcf2maf/1.3/vcf2maf.smk
rdmorin: I think this is because it's set up to be run twice with the config specifically set to do this for either the raw or deblacklisting phase (separated on purpose for blacklist creation)

rdmorin (approved):

modules/vcf2maf/1.3/src/deblacklist_ssm.R
hayashaalan: I did run into some issues with some samples failing because the chromosome column was integer. I can't remember why this was happening.

rdmorin: Were you using fread_maf or perhaps this was before? If you're casting it to an integer here then the damage is already done, I would think, i.e. the non-integer rows have NA.

hayashaalan: I'm pretty sure I was using fread_maf. It was strange because this wasn't happening with all the samples, just a few of them. I see your point though.

Kdreval (approved):

Issue #221, opened 2022-03-09
Harmonize all CNV caller modules and automatically convert seg files to projections


Kdreval:
We have so far 4 different CNV callers in lcr-modules: battenberg, sequenza, controlfreec, and cnvkit. They all are very different in the naming convention of the outputs and neither of them does the automatic conversion of generated seg files to the alternative projection.

The liftover module should be automatically imported by each CNV module and perform conversion to alternative genome build and filling of the segments on a per-sample level to make outputs compatible with downstream tools (GISTIC, pyclone, phylowgs etc). The filling of the SEG files should occur in the liftover module to avoid code duplication between modules, and be implemented for both native and lifted SEG files.

The naming convention of the output files should resemble the convention of vcf2maf outputs as close as possible to allow for convenient globbing:
```
CFG["dirs"]["outputs"] + "seg/{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}.{tool}.seg"
```
and
```
CFG["dirs"]["outputs"] + "seg/{seq_type}--projection/{tumour_id}--{normal_id}--{pair_status}.{tool}.{projection}.seg"
```

This can start from generating `grch37` and `hg38` projections. Support for other versions (`hg19` and `grch38`) can be implemented in future CNV merging module or in the later versions of the CNV callers.

Kdreval:
Unfortunately, snakemake does not allow to import of the same snakefile as a sub-module within several module snakefiles as part of a project. It results in conflicting rules/outputs, with whichever module gets loaded latest gets the preference. This update will need a separate script implemented for liftOver, similar to how crossmap is organized within vcf2maf.

Kdreval:
This has been resolved in the PR #224

PR #224, opened 2022-03-26
Update CNV callers

Kdreval:
This PR is in response to the issue #221.
This update brings consistent logic of CNV modules with that of vcf2maf in terms of the structure of the output folders and automatic conversion of the seg files to the different genome build coordinate system.
In addition, this update will also automatically fill battenberg subclones files for their compatibility with downstream tools.


modules/battenberg/1.1/battenberg.smk
```diff
@@ -189,18 +220,167 @@ rule _battenberg_cleanup:
...
+        {wildcards.tumour_id}
+        SEG
+        2>> {log.stderr};
+        """)
+    
+
+# Symlinks the final output files into the module results directory (under '99-outputs/')
+rule _battenberg_normalize_projection:
+    input:
+        grch37_filled = str(rules._battenberg_fill_segments.output.grch37_filled),
```
lkhilton: Is it possible to generalize this to different projection builds? E.g. prefixed hg19?

lkhilton: Also for samples that start as grch37 or hg38, this step will effectively duplicate the output file from `_battenberg_fill_segments`. Does it make sense to mark some earlier outputs as temp files?

modules/cnvkit/1.0/cnvkit.smk
```diff
@@ -557,26 +558,191 @@ rule _cnvkit_to_seg:
...
     output:
-        call_cns = CFG["dirs"]["outputs"] + "wildcard_capture/BAF_cns/{seq_type}--{genome_build}/{capture_space}/{tumour_id}.call.cns",
-        scatter = CFG["dirs"]["outputs"] + "wildcard_capture/scatter/{seq_type}--{genome_build}/{capture_space}/{tumour_id}_scatter.png",
-        diagram = CFG["dirs"]["outputs"] + "wildcard_capture/diagram/{seq_type}--{genome_build}/{capture_space}/{tumour_id}_diagram.pdf",
-        breaks = CFG["dirs"]["outputs"] + "wildcard_capture/breaks/{seq_type}--{genome_build}/{capture_space}/{tumour_id}_genebreaks.txt",
-        gene_seg = CFG["dirs"]["outputs"] + "wildcard_capture/geneMetrics/{seq_type}--{genome_build}/{capture_space}/{tumour_id}_geneSeg.txt",
-        geneList = CFG["dirs"]["outputs"] + "wildcard_capture/geneList/{seq_type}--{genome_build}/{capture_space}/{tumour_id}_genebreaks.txt",
-        sex = CFG["dirs"]["outputs"] + "wildcard_capture/sex/{seq_type}--{genome_build}/{capture_space}/{tumour_id}_genebreaks.txt",
-        seg = CFG["dirs"]["outputs"] + "wildcard_capture/seg/{seq_type}--{genome_build}/{capture_space}/{tumour_id}.seg"
+        call_cns = CFG["dirs"]["outputs"] + CFG["output"]["cns"]["call"],
```
lkhilton: What's the rationale for moving the output file name specification to the config? This isn't something we do in any other modules I can think of.

lkhilton (commented):
This is a huge improvement.  I wonder if the output genome builds could be more generalized to have more options than grch37 and hg38 similar to the vcf2maf module?

modules/battenberg/1.1/battenberg.smk
Kdreval: I'll try to address the expanding to support more projections, and that will also address the duplication here. This was a way  to handle the hg19 to be converted to non-prefixed ones, but I agree it is not efficient for the ones that already have correct prefixing

modules/cnvkit/1.0/cnvkit.smk
Kdreval: Right, we don't do it in other modules yet - but I thought it might be a good starting point to start switching. This gives any downstream tool, like GAMBLR for example, or a merging snakefile, a way to know upfront the path to the output files, and the file extension of every output. This is easy to pull back into snakefile, but I wanted to see what others think whether this can simplify developing the downstream workflows/functions

lkhilton: Oh gotcha! I forgot that we had discussed that. Makes sense.

rdmorin: Does Python code work when embedded in the config like that?

Kdreval: Yes :heavy_check_mark:  I have run each of these modified modules on the demo sample and can confirm it works and executes as expected, generating files at the expected locations and does not give python/snakemake errors.

rdmorin: It won't work in R though. Can we change it to store the full path instead of the path and Python code or is that not possible?

Kdreval: I think it should be possible! I'll modify and test it out. Will push with the PR comments. Thanks!

modules/battenberg/1.1/battenberg.smk
Kdreval: This has been resolved and the PR is ready for re-review 😊

Kdreval:
This PR is ready for re-review. I have also grouped the post-processing rules to efficiently use cluster resources. To determine the chr prefixing status of the projection, I used a slightly different approach than in vcf2maf: I am reading the `main_chromosomes.txt` file of the desired projection, and then determined the prefixing status based on the chromosome names. This avoids using the reference_files config directly (so no need to add yet another config key to specify the path to it) and also avoids creating a separate dictionary for this purpose.
The files before projection are normalized are marked as `temp()`, so they are deleted afterward and this avoids duplication of outputs.

lkhilton (approved):
Brilliant! Thank you for tackling this.

PR #225, opened 2022-04-05
CNVkit: Added wildcard constraints, explicit lists

lkhilton:
When trying to run this module in a fresh project space I ran into all kinds of wildcard expansion problems. This patch addresses that.

### DIFF
--- a/modules/battenberg/1.1/battenberg.smk
+++ b/modules/battenberg/1.1/battenberg.smk
@@ -157,7 +157,7 @@ rule _run_battenberg:
 rule _battenberg_to_igv_seg:
     input:
         sub = rules._run_battenberg.output.sub,
-        cnv2igv = CFG["inputs"]["cnv2igv"]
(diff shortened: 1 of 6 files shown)


Kdreval (approved):
Thanks Laura! I wonder what can be causing this, so we can be catching/addressing these more systematically. Maybe a different snakemake or python version, or something like absolute/relative path (so more moving parts for snakemake to figure out) 🤔 Maybe we should start a thread on Slack to collect these, and then we can notice a trend associated with these errors.

lkhilton:
Yeah I went through this with vcf2maf as well. I'm using the gambl-opv12 environment, so same Snakemake as you, I think?

Kdreval:
right, yes  - same snakemake/python/pandas versions. Please feel free to merge it - or will be happy to approve again if anything else comes up

Kdreval (approved):

PR #228, opened 2022-04-12
Module/cutesv/1.0

nthomas11:
# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable

Example:
```
```


modules/cutesv/1.0/cutesv.smk
```diff
@@ -0,0 +1,144 @@
...
+    conda :
+        CFG["conda_envs"]["cutesv"] 
+    log:
+        stderr = CFG["logs"]["cutesv"] + "{seq_type}--{genome_build}/{sample_id}/cutesv.stderr.log"             
+    shell:
+         op.as_one_line("""
+            mkdir {output.dir} &&
+            cuteSV {input.bam} {input.fasta} {output.vcf} {output.dir}
+            --threads {threads} --max_cluster_bias_INS 100 --diff_ratio_merging_INS 0.3 
+            --max_cluster_bias_DEL 100 --diff_ratio_merging_DEL 0.3""")
```
lkhilton: Would users ever want/need to adjust these values, and if yes can you add them to the config and parse them under params?

lkhilton (changes requested):
Looks good - see some comments below. Make sure your log files are set up to capture the stderr for all your rules!

modules/cutesv/1.0/cutesv.smk
```diff
@@ -0,0 +1,144 @@
...
+
+# Symlinks the final output files into the module results directory (under '99-outputs/')
+rule _cutesv_output:
+    input:
+        cutesv = str(rules._cutesv.output.vcf)
+    output:
+        cutesv = CFG["dirs"]["outputs"] + "cutesv/{seq_type}--{genome_build}/{sample_id}.vcf"
+
+    run:
+        op.relative_symlink(input.cutesv, output.cutesv, in_module= True)
```
lkhilton: Same here as for Sniffles - can you add a bedpe-generating step?

nthomas11: Added rule for vcftobedpe

modules/cutesv/1.0/cutesv.smk
nthomas11: Updated rule to place values in params so user can specify if they want to change the values

modules/cutesv/1.0/config/default.yaml
```diff
@@ -6,17 +6,26 @@ lcr-modules:
             sample_bam: "__UPDATE__"
             sample_bai: "__UPDATE__"
         
-        scratch_subdirectories: []
+        scratch_subdirectories: ["cutesv_working"]
```
lkhilton: Could you please leave this blank for the default? It can cause some issues on re-runs of the same module.

lkhilton (changes requested):
Looks great, thank you for adding the bedpe step. Once you revert the default `scratch_subdirectories` value I will approve!

modules/cutesv/1.0/config/default.yaml
nthomas11: Fixed!

lkhilton (approved):

PR #229, opened 2022-04-12
Module/nanopolish/1.0

nthomas11:
# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable

Example:
```
```


modules/nanopolish/1.0/nanopolish.smk
```diff
@@ -0,0 +1,266 @@
...
+        gzi = CFG["inputs"]["sample_gzi"],
+        readdb = CFG["inputs"]["sample_readdb"]
+    output:
+        bam = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{sample_id}.bam",
+        bai = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{sample_id}.bam.bai",
+        fastq = CFG["dirs"]["inputs"] + "fastq/{seq_type}--{genome_build}/{sample_id}.fastq",
+        index = CFG["dirs"]["inputs"] + "fastq/{seq_type}--{genome_build}/{sample_id}.fastq.index",
+        fai = CFG["dirs"]["inputs"] + "fastq/{seq_type}--{genome_build}/{sample_id}.fastq.index.fai",
+        gzi = CFG["dirs"]["inputs"] + "fastq/{seq_type}--{genome_build}/{sample_id}.fastq.index.gzi",
+        readdb = CFG["dirs"]["inputs"] + "fastq/{seq_type}--{genome_build}/{sample_id}.fastq.index.readdb",
```
lkhilton, line 90: We don't have these .index.fai/gzi/readdb files symlinked into GAMBL. Are they generated by default for each fastq file by the GSC?

lkhilton: I see that you generated them within the data directories for the BLGSP promethION data. Could you add a rule or rules to generate these files in the `00-inputs` directory alongside the fastq files after you've symlinked them in?

modules/nanopolish/1.0/nanopolish.smk
```diff
@@ -0,0 +1,266 @@
...
+# Installs latest nanopolish release from github if the nanopolish folder is not present yet
+rule _install_nanopolish:
+    params:
+        nanopolish = CFG["inputs"]["nanopolish_exec"]
+    output: 
+        complete = CFG["inputs"]["nanopolish_exec"] + "/nanopolish_dependencies_installed.success"
+    shell:
+        '''
+
+        if [ ! -f {params.nanopolish}/nanopolish ]; then
```
lkhilton: You shouldn't need the `if` statement since Snakemake will only run a rule if the output file is missing or older than the output of a downstream rule. You can also wrap the output in touch: 
```
complete = touch(CFG["inputs"]["nanopolish_exec"] + "/nanopolish_dependencies_installed.success")
```
Then Snakemake will take care of touching the output file if the rule runs successfully.

modules/nanopolish/1.0/nanopolish.smk
```diff
@@ -0,0 +1,266 @@
...
+        calls = CFG["dirs"]["merged_nanopolish_calls"] + "{seq_type}--{genome_build}/{sample_id}.calls.tsv.gz"
+    resources: 
+        mem_mb = CFG["mem_mb"]["merged_nanopolish_calls"] 
+    log:
+        stderr = CFG["logs"]["merged_nanopolish_calls"] + "{seq_type}--{genome_build}/{sample_id}/merged_nanopolish_calls.stderr.log"               
+    shell:
+        op.as_one_line("""
+        zcat {input.calls} | sort -V -k1,1 -k2,2 |
+        cat <(grep "chromosome" | head -n1) <(grep -v "chromosome") | 
+        gzip >  {output.calls}
```
lkhilton: Can you add a redirect to write the stderr of this command to your log file?

lkhilton: This applies to all shell bash commands

lkhilton (changes requested):
Some requests in comments!

modules/nanopolish/1.0/nanopolish.smk
nthomas11: These files are actually generated by the GSC and are in the same directory as the fastqs. I don't know if it makes sense to just symlink them into gambl under the fastq directory?

lkhilton: Got it! I can ask Haya to update the GAMBL script that symlinks the PromethION data to include these files.

modules/nanopolish/1.0/nanopolish.smk
nthomas11: Removed if statement and changed output complete file as suggested

modules/nanopolish/1.0/nanopolish.smk
nthomas11: Redirected stderr to log file and removed logs from merge rules since there wouldn't be any stderr generated from bash cat commands

modules/nanopolish/1.0/nanopolish.smk
```diff
@@ -102,20 +102,17 @@ rule _install_nanopolish:
...
+        op.as_one_line('''
+        mkdir -p {params.nanopolish};
+        git clone git@github.com:jts/nanopolish.git {params.nanopolish};
+        cd {params.nanopolish};
+        make;
+        cd -;
         
-        touch  {output.complete};
-        '''
+        {output.complete};
```
lkhilton: Won't this error out? I think it'll try to run `nanopolish_dependencies_installed.success` as a command, right? I think you can just delete this line altogether and end at `make`.

lkhilton (changes requested):
Looks good - will approve once the shell command for installing Nanopolish is fixed. Thank you!

modules/nanopolish/1.0/nanopolish.smk
nthomas11: Good catch! Removed {output.complete} line to fix this issue!

lkhilton (approved):
Thanks!

PR #231, opened 2022-04-14
cleanup: squish sequenza config options under one key

Kdreval:
When doing the module updates, I missed the existing key `options` in sequenza config. This caused conflict when setting up the module in GAMBL format. This PR addresses this issue by combining the options under one key. Confirmed it works as expected by running this on all GAMBL genomes.


lkhilton (approved):

Kdreval:
Thanks Laura!

PR #232, opened 2022-04-21
CNV merging for master seg flatfiles

Kdreval:
This PR is not only an important step in GAMBL and GAMBLR functionality but also a step to facilitate the lymphgen-running snakefile for the demo.

# Pull Request Checklists

### Required

  ```bash
  ```

Example:
```
```


modules/cnv_master/1.0/config/default.yaml
```diff
@@ -0,0 +1,45 @@
+lcr-modules:
+
+    cnv_master:
+        # comma-separated list of modules used to call CNV
+        # this order will also determine the preference which module's output will be used when multiple outputs are available
+        # example: ["battenberg", "sequenza", "cnvkit", "controlfreec"]
+        names: ["__UPDATE__", "__UPDATE__", ..., "__UPDATE__"]
```
lkhilton: Why not have these as key:value pairs? e.g. 

```
seg: 
    battenberg: 'results/battenberg-1.2/99-outputs/seg/{seq_type}--projection/{tumour_id}--{normal_id}--{pair_status}.{tool}.{projection}.seg'
    sequenza: 'results/sequenza-1.4/99-outputs/seg/{seq_type}--projection/{tumour_id}--{normal_id}--{pair_status}.{tool}.{projection}.seg'

```
Then you can access the names using `CFG["seg"].keys()`

modules/cnv_master/1.0/src/R/merge_segs.R
```diff
@@ -0,0 +1,54 @@
...
+
+suppressWarnings(suppressPackageStartupMessages({
+
+  message("Loading packages...")
+
+  required_packages <- c(
+    "readr", "dplyr", "tidyr", "purrr"
+  )
+
+  installed_packages <- rownames(installed.packages())
```
lkhilton: I think installing packages in a Snakemake-run script causes problems especially if more than one instance of this is running simultaneously. Doesn't the conda env take care of installing the packages?

modules/cnv_master/1.0/config/default.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+        threads:
+            cnv_master: 4
+
+        resources:
+            cnv_master:
+                mem_mb: 12000
+                disk_mb: 1000 # Will ensure at least this much space is left
+
+        pairing_config:
```
lkhilton, line 32: Will this cause compatibility issues with Battenberg since it's not run with unmatched normals? Or is that handled by the input function that checks for availability of the output file?

lkhilton (commented):
Some questions and comments in review. 

One of the reasons I made my SV merging script in GAMBL and not LCR-modules is because unix_group is a GAMBL quirk. I don't think it's handled here - do you have a plan to handle this in GAMBL?

modules/cnv_master/1.0/config/default.yaml
Kdreval: Makes sense, will do!

modules/cnv_master/1.0/config/default.yaml
Kdreval: Right, it is handled by the input function. This pairing config here will only ensure that the output of any tool is present for any sample - so all tumors in the sample table are included in the merge.

modules/cnv_master/1.0/src/R/merge_segs.R
Kdreval: You are right, these packages are already there from conda env. I will just load them here and skip the installation part

Kdreval (commented):
Thanks Laura! My plan to handle unix groups in GAMBL is to have unix_group-specific outputs that would be merged then to `all_the_things`. I am thinking of this way because it will be then straightforward to set GAMBLR config for permission-defined access to flat files with merges. This can be converted to a GAMBL-specific snakefile if you think this will be a better approach. I thought of having it as a module because we would need a merged seg file as input to lymphgen module in upcoming lymphgen-centric project snakefile as part of demo

modules/cnv_master/1.0/src/R/merge_segs.R
lkhilton: > Thanks Laura! My plan to handle unix groups in GAMBL is to have unix_group-specific outputs that would be merged then to `all_the_things`. I am thinking of this way because it will be then straightforward to set GAMBLR config for permission-defined access to flat files with merges. This can be converted to a GAMBL-specific snakefile if you think this will be a better approach. I thought of having it as a module because we would need a merged seg file as input to lymphgen module in upcoming lymphgen-centric project snakefile as part of demo

This makes sense, although the alternative we've discussed is having LymphGen run per-sample rather than on the merge. That kind of update would still have to leverage the input functions you've used here to get the best CNV output though, and doesn't negate the need for individual unix_group merges in GAMBL. 

From a practicality standpoint and to avoid having to wait for merges before re-running LymphGen I favour the per-sample approach, especially now that the merges we're generating will be a perfect concatenation of the input maf files.

lkhilton: I don't know why GitHub put my quote reply in this conversation...

Kdreval: I see, right - we can run lymphgen per sample. For capture/exome data, the SVs from FISH would be supplied in a separate file and I guess it would not really affect the module if FISH data is per-sample or all in one file, so we can still run it individually

lkhilton: Awesome. This module is still useful though!

Kdreval: This has been addressed 👍

modules/cnv_master/1.0/config/default.yaml
Kdreval: I have switched to the named dictionary here 👍

modules/cnv_master/1.0/src/R/merge_segs.R
Kdreval: Another thing I also just remembered is that initially, I thought of this module to prepare inputs for GISTIC and SV-hotspot. The SV-hotspot did not prove to be that useful, but I am still planning on the GISTIC module and this merging would be needed there

lkhilton: Excellent. Another good reason to keep this.

lkhilton (approved):

PR #236, opened 2022-05-11
IgCaller module

ppararaj:
# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable


ppararaj:
- Tested on capture and genome BAMs and CRAMs using both paired and unpaired samples

lkhilton:
Looks good, Prasath. I think you could add a rule to auto-install IgCaller though. The MiXCR module has an example of this: 
https://github.com/LCR-BCCRC/lcr-modules/blob/dfa9cc27daa36c242906f9acec900503031c3b55/modules/mixcr/1.1/mixcr.smk#L69-L86

ppararaj:
Added a rule to install IgCaller in a specified directory. Successfully tested the pipeline on all relevant samples.

lkhilton (approved):
Thanks for adding that, Prasath!

PR #239, opened 2022-06-06
IgBLAST reference files rules

mannycruz:
# Pull Request Checklists

Added rules to:
- download required files for IgBLAST run
- download IMGT reference sequences for IG and TCR receptors
- convert IMGT sequences into format that can be passed to "makeblastdb" (function provided by igblast)
- make IgBLAST db from IMGT ref sequences
Also:
-  added igblast.yaml file in envs/ and updated the reference files config default.yaml to include the igblast environment
- removed URL code fetching lines from reference_files_header.smk that were causing snakemake to error out with a URLError

### DIFF
--- a/workflows/reference_files/2.4/config/default.yaml
+++ b/workflows/reference_files/2.4/config/default.yaml
@@ -144,6 +144,9 @@ tools:
     bcftools:
         conda_env: "envs/bcftools-1.10.2.yaml"
         version: "1.10.2"
+    igblast:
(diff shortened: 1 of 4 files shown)


lkhilton:
Manuela, how close are your updates to the MiXCR module to being ready for review? I would prefer to review this code along with the module updates.

mannycruz:
They're ready! I can push that as well

lkhilton:
This is redundant with #240 so the changes will be reviewed there.

PR #240, opened 2022-06-07
MiXCR 1.2 module updates

mannycruz:
# Pull Request Checklists
### Required

--> I forgot to do this! I'll do this all together after I get some feedback
--> I forgot to do this for the last rule _mixcr_chains

I have some edits I need to make still, the ones listed above ^ plus I have to fix the op.setup_module statement because it is still listed as 'version = "1.1"' I think.
I accidentally lost my staged changes and had to recover them by resetting back to a previous commit period in time so I had some trouble getting this ready

### DIFF
--- a/modules/mixcr/1.2/config/default.yaml
+++ b/modules/mixcr/1.2/config/default.yaml
@@ -2,6 +2,9 @@ lcr-modules:
     
     mixcr:
 
+        dirs:
+            _parent: "__UPDATE__"
+
         inputs:
             # Available wildcards: {seq_type} {sample_id}
             sample_fastq_1: "__UPDATE__" 
--- a/modules/mixcr/1.2/config/default.yaml
+++ b/modules/mixcr/1.2/config/default.yaml
@@ -1,0 +1,58 @@
+lcr-modules:
+    
+    mixcr:
+
+        inputs:
+            # Available wildcards: {seq_type} {sample_id}
+            sample_fastq_1: "__UPDATE__" 
+            sample_fastq_2: "__UPDATE__"
+            # Path to the directory where MIXCR should be installed
+            mixcr_exec: "__UPDATE__"
+
+        scratch_subdirectories: []
+
+        receptors: # Set in snakefile
+        ###### Desired IG/TCR chains in list format or keyword:
+        ###### "ALL" = ['IGH', 'IGL', 'IGK', 'TRA', 'TRB', 'TRD', 'TRG']
+        ###### "BCR" = ['IGH', 'IGL', 'IGK']
+        ###### "TCR" = ['TRA', 'TRB', 'TRD', 'TRG']
+
+        igblastn: "__UPDATE__"
+        #Boolean
+        # Set to True to perform igblastn alignment to IMGT database
+
+        igblast_scripts:
+            mixcr2fasta: "{MODSDIR}/src/mixcr_to_fasta.py"
+            igblastn2mixcr: "{MODSDIR}/src/igblast_to_mixcr.py"
+
+        options:
+            mixcr_run: 
+                mrna:
+                    --starting-material rna
+                genome:
+                    --starting-material dna
+            igblast_run:
+                run_flags: ""
+                form: "qseq sseq"
+
+        conda_envs: 
+            java: "{MODSDIR}/envs/java-jdk-8.0.yaml"
+            igblast: "{MODSDIR}/envs/igblast.yaml"
+            
+        threads:
+            mixcr_run: 4
+
+        resources:
+            mixcr_run: 
+                mem_mb: 10000
+                fastq: 1
+
+        pairing_config:
+            genome:
+                run_paired_tumours: False
+                run_unpaired_tumours_with: "no_normal"
+                run_paired_tumours_as_unpaired: True
+            mrna:
+                run_paired_tumours: False
+                run_unpaired_tumours_with: "no_normal"
+                run_paired_tumours_as_unpaired: True
--- a/modules/mixcr/1.2/config/default.yaml
+++ b/modules/mixcr/1.2/config/default.yaml
@@ -11,14 +11,14 @@ lcr-modules:
 
         scratch_subdirectories: []
 
-        receptors: # Set in snakefile
+        receptors: "__UPDATE__"
         ###### Desired IG/TCR chains in list format or keyword:
         ###### "ALL" = ['IGH', 'IGL', 'IGK', 'TRA', 'TRB', 'TRD', 'TRG']
         ###### "BCR" = ['IGH', 'IGL', 'IGK']
         ###### "TCR" = ['TRA', 'TRB', 'TRD', 'TRG']
 
-        igblastn: "__UPDATE__"
-        #Boolean
+        igblastn: False
+        # Boolean
         # Set to True to perform igblastn alignment to IMGT database
 
         igblast_scripts:
--- a/modules/mixcr/1.2/mixcr.smk
+++ b/modules/mixcr/1.2/mixcr.smk
@@ -208,7 +208,7 @@ if CFG["igblastn"]:
     rule _mixcr_to_fasta:
         input:
             mixcr_finished = str(rules._mixcr_run.output.txt),
-            mixcr_chains = str(rules._mixcr_run.output.results),
+            mixcr_chains = rules._mixcr_run.output.results,
             mixcr_results = CFG["dirs"]["mixcr"] + "{seq_type}/{sample_id}/mixcr.{sample_id}.clonotypes.{chain}.txt",
             script = CFG["igblast_scripts"]["mixcr2fasta"]
         output:
--- a/modules/mixcr/1.2/mixcr.smk
+++ b/modules/mixcr/1.2/mixcr.smk
@@ -1,0 +1,319 @@
+#!/usr/bin/env snakemake
+
+
+##### ATTRIBUTION #####
+
+
+# Original Author:  Anita dos Santos
+# Module Author:    Manuela Cruz
+# Contributors:     Laura Hilton
+
+
+##### SETUP #####
+
+
+# Import package with useful functions for developing analysis modules
+import oncopipe as op
+
+# Check that the oncopipe dependency is up-to-date. Add all the following lines to any module that uses new features in oncopipe
+min_oncopipe_version="1.0.11"
+import pkg_resources
+try:
+    from packaging import version
+except ModuleNotFoundError:
+    sys.exit("The packaging module dependency is missing. Please install it ('pip install packaging') and ensure you are using the most up-to-date oncopipe version")
+
+# To avoid this we need to add the "packaging" module as a dependency for LCR-modules or oncopipe
+
+current_version = pkg_resources.get_distribution("oncopipe").version
+if version.parse(current_version) < version.parse(min_oncopipe_version):
+    logger.warning(
+                '\x1b[0;31;40m' + f'ERROR: oncopipe version installed: {current_version}'
+                "\n" f"ERROR: This module requires oncopipe version >= {min_oncopipe_version}. Please update oncopipe in your environment" + '\x1b[0m'
+                )
+    sys.exit("Instructions for updating to the current version of oncopipe are available at https://lcr-modules.readthedocs.io/en/latest/ (use option 2)")
+
+# End of dependency checking section 
+
+# Setup module and store module-specific configuration in `CFG`
+# `CFG` is a shortcut to `config["lcr-modules"]["mixcr"]`
+CFG = op.setup_module(
+    name = "mixcr",
+    version = "1.1",
+    subdirectories = ["inputs", "mixcr", "outputs"],
+)
+
+assert type(CFG["igblastn"])==bool, (
+    "Ensure 'igblastn' is set to either True or False in config. "
+    "True: Runs igblastn alignment to retrieve % identity to IMGT sequences. This requires the '--contig-assembly' parameter when running MiXCR and may increase running time. "
+    )
+
+run_parameters = CFG["options"]["mixcr_run"]["mrna"]
+
+if CFG["igblastn"]:
+    assert "--contig-assembly" in run_parameters, (
+        "Config 'igblastn' value is set to True, but ['mixcr_run']['options'] does not include '--contig-assembly'. "
+        "Include '--contig-assembly' in CFG['mixcr_run']['options'] to run igblastn.\n "
+        "\n***** NOTE: using '--contig-assembly' may increase run duration. *****\n"
+    )
+
+RECEPTORS = CFG["receptors"]
+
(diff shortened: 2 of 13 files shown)


modules/mixcr/1.2/mixcr.smk
```diff
@@ -0,0 +1,320 @@
...
+run_parameters = CFG["options"]["mixcr_run"]["mrna"]
+
+if CFG["igblastn"]:
+    assert "--contig-assembly" in run_parameters, (
+        "Config 'igblastn' value is set to True, but ['mixcr_run']['options'] does not include '--contig-assembly'. "
+        "Include '--contig-assembly' in CFG['mixcr_run']['options'] to run igblastn.\n "
+        "\n***** NOTE: using '--contig-assembly' may increase run duration. *****\n"
+    )
+
+RECEPTORS = CFG["receptors"]
```
lkhilton, line 63: Why do you set this from the config, but the config says "Set in Snakefile" for the `receptors` key?

modules/mixcr/1.2/mixcr.smk
```diff
@@ -0,0 +1,320 @@
...
+
+RANGE = ["IGH","IGL","IGK","TRA","TRB","TRD","TRG"]
```
lkhilton: You could avoid some redundancy here by saying `RANGE = ig_type + tr_type`

modules/mixcr/1.2/config/default.yaml
```diff
@@ -0,0 +1,58 @@
...
+        inputs:
+            # Available wildcards: {seq_type} {sample_id}
+            sample_fastq_1: "__UPDATE__" 
+            sample_fastq_2: "__UPDATE__"
+            # Path to the directory where MIXCR should be installed
+            mixcr_exec: "__UPDATE__"
+
+        scratch_subdirectories: []
+
+        receptors: # Set in snakefile
```
lkhilton: Should this say `__UPDATE__` and require the user to set a value?

modules/mixcr/1.2/mixcr.smk
```diff
@@ -0,0 +1,320 @@
...
+    version = "1.2",
+    subdirectories = ["inputs", "mixcr", "outputs"],
+)
+
+assert type(CFG["igblastn"])==bool, (
+    "Ensure 'igblastn' is set to either True or False in config. "
+    "True: Runs igblastn alignment to retrieve % identity to IMGT sequences. This requires the '--contig-assembly' parameter when running MiXCR and may increase running time. "
+    )
+
+run_parameters = CFG["options"]["mixcr_run"]["mrna"]
```
lkhilton: If someone was running a mix of seq_types, how would this work?

modules/mixcr/1.2/mixcr.smk
```diff
@@ -0,0 +1,320 @@
...
+    params:
+        opts = op.switch_on_wildcard("seq_type", CFG["options"]["mixcr_run"]),
+        prefix = CFG["dirs"]["mixcr"] + "{seq_type}/{sample_id}/mixcr.{sample_id}", 
+        mixcr = CFG["inputs"]["mixcr_exec"] + "/mixcr", 
+        jvmheap = lambda wildcards, resources: int(resources.mem_mb * 0.8) 
+    conda: CFG["conda_envs"]["java"]
+    threads:
+        CFG["threads"]["mixcr_run"]
+    message:
+        "{params.mixcr}"
```
lkhilton: Can you make this more informative, e.g. `Using MiXCR installation at {params.mixcr}.` Or delete it, I'm not sure it's necessary.

modules/mixcr/1.2/mixcr.smk
```diff
@@ -0,0 +1,320 @@
...
+RECEPTORS = CFG["receptors"]
+
+RANGE = ["IGH","IGL","IGK","TRA","TRB","TRD","TRG"]
+
+ig_type = ["IGH","IGK","IGL"]
+tr_type = ["TRA", "TRB", "TRD", "TRG"]
+
+if isinstance(RECEPTORS, str):
+    RECEPTORS = RECEPTORS.split(" ")
+if isinstance(RECEPTORS, list) and len(RECEPTORS) == 1:
```
lkhilton, line 72: What about the case where `RECEPTORS` is a list of length > 1?

modules/mixcr/1.2/mixcr.smk
```diff
@@ -0,0 +1,320 @@
...
+            """)
+
+    rule _update_mixcr_results:
+        input:
+            db = rules._igblastn_run.output.db,
+            og_mixcr = CFG["dirs"]["mixcr"] + "{seq_type}/{sample_id}/mixcr.{sample_id}.clonotypes.{chain}.txt",
+            seq_info = rules._mixcr_to_fasta.output.seq_info,
+            script = CFG["igblast_scripts"]["igblastn2mixcr"]
+        output:
+            txt = CFG["dirs"]["mixcr"] + "{seq_type}/{sample_id}/mixcr.{sample_id}.clonotypes.{chain}.status.txt"
```
lkhilton: I'd like to suggest that `txt = CFG["dirs"]["mixcr"] + "{seq_type}/{sample_id}/mixcr.{sample_id}.clonotypes.{chain}.igblast.txt"` would be a more informative name for this output file.

modules/mixcr/1.2/mixcr.smk
```diff
@@ -0,0 +1,320 @@
...
+        run:
+            op.relative_symlink(input.txt, output.txt, in_module=True)
+
+# Symlinks the final output files into the module results directory (under '99-outputs/')
+rule _mixcr_output_txt:
+    input:
+        txt = rules._mixcr_run.output.txt,
+        report = rules._mixcr_run.output.report
+    output:
+        txt = CFG["dirs"]["outputs"] + "txt/{seq_type}/mixcr.{sample_id}.clonotypes.ALL.txt",
```
lkhilton: IIRC MiXCR creates an empty `clonotypes.ALL.txt` file when you specify either `bcr` or `tcr` for `--receptor-type`, so I would say it makes sense to change this rule to symlink the individual chain files instead of `ALL.txt`. Or you can do both if you really want to.

modules/mixcr/1.2/mixcr.smk
```diff
@@ -0,0 +1,320 @@
...
+        report = CFG["dirs"]["outputs"] + "txt/{seq_type}/mixcr.{sample_id}.report"
+    run:
+        op.relative_symlink(input.txt, output.txt, in_module=True)
+        op.relative_symlink(input.report, output.report, in_module=True)
+
+rule _mixcr_chains:
```
lkhilton: Is this rule needed?

lkhilton: If you symlink the `clonotypes.{chain}.txt` files in the rule above, this would be unnecessary.

modules/mixcr/1.2/mixcr.smk
```diff
@@ -0,0 +1,320 @@
...
+    input:
+        results = CFG["dirs"]["mixcr"] + "{seq_type}/{sample_id}/mixcr.{sample_id}.clonotypes.{chain}.txt"
+
+
+# Generates the target sentinels for each run, which generate the symlinks
+
+if CFG["igblastn"]:
```
lkhilton, line 282: I think you want all the mixcr output files symlinked regardless of whether igblastn is run, so I think your if statement should only add an output file if CFG["igblastn"] == True

modules/mixcr/CHANGELOG.md
```diff
@@ -5,6 +5,18 @@ All notable changes to the `mixcr` module will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.2] - 2022-06-6
+
+This release was updated by Manuela Cruz.
+
+- Removes dependency on `{genome_build}`wildcard.
```
lkhilton, line 12: YAY

workflows/reference_files/2.4/reference_files_header.smk
```diff
@@ -75,12 +75,6 @@ for build_name, build_info in config["genome_builds"].items():
     assert "provider" in build_info and genome_provider in possible_providers, (
         f"`provider` not set for `{build_name}` or `provider` not among {possible_providers}."
     )
-    if "genome_fasta_url" in build_info:
```
lkhilton, line 78: Could you delete the similar lines below starting from `if "capture_bed_url" in build_info:` please?

workflows/reference_files/2.4/reference_files.smk
```diff
@@ -598,3 +598,105 @@ rule install_sigprofiler_genome:
...
+        database = expand("genomes/no_build/igblast/database/imgt_human_{{chain}}.{ext}", ext = ['ndb','nhr','nin','nog','nos','not','nsq','ntf','nto'])
+    conda: CONDA_ENVS["igblast"]
+    shell:
+        op.as_one_line("""
+        fasta_file={input.fasta} && 
+        output_file="genomes/no_build/igblast/database/imgt_human_{wildcards.chain}" && 
+        makeblastdb -parse_seqids -dbtype nucl -in $fasta_file -out $output_file
+        """)
+
+rule _imgt_db_success:
```
lkhilton, line 696: There's another Snakefile that can be used to generate all reference files prior to doing any runs. 
https://github.com/LCR-BCCRC/lcr-modules/blob/0cf2b50b501806be1d1a507a7c0e1eb91323c8ae/workflows/reference_files/2.4/prepare_reference_files.smk#L35-L68

Can you please add the output of this rule to the input of that one? You can add  the possible chain values to the config following the example of the other `config["wildcard_values"]["..."]` examples there.

lkhilton (changes requested):
I'm a bit concerned about the complexity of how you're parsing the user's specification of which chains they want to run. There's also the issue that the user might have the same snakefile set up to run different seq_types, and part of your function to set the RECEPTORS value assumes that the value under the `mrna` key is correct.

IMO you should have only one way to set which receptor type you want. If you want to use the `CFG["receptors"]` value, drop the `--receptor-type` option from `CFG["options"]["mixcr_run"]` and make it explicit that users should not specify a `--receptor-type` option. Include an assert statement to catch it and throw an error if the receptor-type is set there. OTOH if you want to maintain the current functionality, parse the  `CFG["options"]["mixcr_run"]`  to get the `--receptor-type` and drop the `CFG["receptors"]` step altogether. 

I also don't see a reason to let users specify anything besides xcr|tcr|bcr (i.e. you don't need to give users the option to only get IGH outputs). The files aren't big enough and it doesn't make the pipeline any more computationally intensive to just require all possible chains for a given receptor type. 

Lastly, you can preserve the ability to request different receptor types depending on the seq_type by using `op.switch_on_wildcard` in the final symlinking rules to define the target chains and expand on that wildcard (I think). It might require an input function. 

Otherwise this is a major improvement to this module, thank you so much for doing this!

modules/mixcr/1.2/config/default.yaml
mannycruz: Oops, yes it should!

modules/mixcr/1.2/mixcr.smk
mannycruz: Hmmm here I am assuming that if the length is longer than > 1, that the receptor input is multiple receptors for example ["IGH","IGK"] or ["TRA","TRB"]. The assert statement below makes sure that all receptors specified are within the accepted values, so I think that will handle the cases where len(RECEPTORS) > 1 without having to handle them in this block of code

modules/mixcr/1.2/mixcr.smk
mannycruz: I don't think it's necessary either, I'll delete it

modules/mixcr/1.2/mixcr.smk
mannycruz: Fixed by adding  `receptors: "__ UPDATE __"` to the default.yaml config

modules/mixcr/1.2/mixcr.smk
mannycruz: Good idea, updated to `RANGE = ig_type + tr_type`

modules/mixcr/1.2/mixcr.smk
mannycruz: Updated output file to `txt = CFG["dirs"]["mixcr"] + "{seq_type}/{sample_id}/mixcr.{sample_id}.clonotypes.{chain}.igblast.txt"`

modules/mixcr/1.2/mixcr.smk
mannycruz: Changed `_mixcr_output_txt` input to a single `_mixcr_run` output file: `CFG["dirs"]["mixcr"] + "{seq_type}/{sample_id}/mixcr.{sample_id}.clonotypes.{chain}.txt"` and changed output to `CFG["dirs"]["outputs"] + txt/{seq_type}/mixcr.{sample_id}.clonotypes.{chain}.txt"`

modules/mixcr/1.2/mixcr.smk
mannycruz: `rule _mixcr_chains` is not needed anymore with the changes made to `rule _mixcr_output_txt`

modules/mixcr/1.2/mixcr.smk
mannycruz: I updated the `_mixcr_all` rules so that when `CFG["igblastn"] == True`, the extra file created by `_symlink_mixcr_update` is added to the required inputs

workflows/reference_files/2.4/reference_files.smk
mannycruz: Added the `rules._imgt_db_success.output.txt` rule to `prepare_reference_files.smk` rule all

workflows/reference_files/2.4/reference_files_header.smk
mannycruz: Done 👍

modules/mixcr/1.2/mixcr.smk
```diff
@@ -245,40 +247,36 @@ if CFG["igblastn"]:
...
 
     rule _symlink_mixcr_update:
         input:
-            txt = rules._update_mixcr_results.output.txt
+            txt = str(rules._update_mixcr_results.output.txt)
         output:
-            txt = CFG["dirs"]["outputs"] + "txt/{seq_type}/mixcr.{sample_id}.clonotypes.{chain}.status.txt"
+            txt = CFG["dirs"]["outputs"] + "txt/{seq_type}/mixcr.{sample_id}.clonotypes.{chain}.igblast.txt"
+        wildcard_constraints:
+            chain = '[A-Z]+'
```
lkhilton: A better wildcard constraint here is: 
```
chain = "|".join(RECEPTORS)
```
This makes a string of the possible values of the chain wildcards separated by the `|` symbol

lkhilton (approved):

PR #241, opened 2022-06-15
Module/mutsig/1.0

Kdreval:
# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable

- [X] I added more granular output subdirectories.

- [X] I updated the list of available wildcards for the input files in the `default.yaml` configuration file.

### DIFF
--- a/modules/mutsig/1.0/mutsig.smk
+++ b/modules/mutsig/1.0/mutsig.smk
@@ -96,6 +96,24 @@ rule _mutsig_prepare_maf:
         """)
 
 
+# Download the MutSig2CV
+rule _mutsig_download_mutsig:
+    output:
+        mutsig_compressed = temp(CFG["dirs"]["inputs"] + "MutSig2CV/MutSig2CV.tar.gz")
+        mutsig = CFG["dirs"]["inputs"] + "MutSig2CV/MutSig2CV.success"
+    conda:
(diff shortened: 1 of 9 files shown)


modules/mutsig/1.0/mutsig.smk
```diff
@@ -0,0 +1,295 @@
...
+        ln -sf MutSig2CV/mutsig2cv/reference .
+            &&
+        touch configure.success
+        """)
+
+
+# Actual MutSig2CV run
+rule _mutsig_run:
+    input:
+        mcr = str(rules._mutsig_download_mcr.output.local_mcr),
```
lkhilton: I recommend wrapping all references to software installation outputs in `ancient()`.

Kdreval:
I have confirmed it works in GAMBL settings :heavy_check_mark:

lkhilton (changes requested):
One minor change requested!

Kdreval:
I wrapped the installation inputs into `ancient()` and this is ready for re-review (or approval)

Kdreval:
I have also wrapped the sample subsets file into ancient because I realised once you add a new subset to the table, it will want to re-run all previous subsets as well. This modification prevents from unnecessary re-runs, confirmed :heavy_check_mark:

lkhilton (approved):

PR #242, opened 2022-06-29
Module/oncodriveclustl/1.0

mannycruz:
# Pull Request Checklists

### Required

  - [x] All references to the `rules` object (_e.g._ for input files) are wrapped with `str()`. **(I forgot again)**

- [x] I updated the final target rule (`*_all`) to include every output rule. **(I think?)**

- [x] I explained important module design decisions in `CHANGELOG.md`. **(I forgot, will do!!!)**

  ```bash
  ```

### If applicable

Example:
```
```


mannycruz:
There are 3 commits that should not be merged, I just added them to keep track of the previous script I had made: [c26cbd7], [718ec81], [150788e]

lkhilton:
> There are 3 commits that should not be merged, I just added them to keep track of the previous script I had made: [[c26cbd7](https://github.com/LCR-BCCRC/lcr-modules/commit/c26cbd7ee97b4a52e65c6604a4cc2b50eb7227d6)], [[718ec81](https://github.com/LCR-BCCRC/lcr-modules/commit/718ec81673d88a00ea13bf288ebbbca2011252a4)], [[150788e](https://github.com/LCR-BCCRC/lcr-modules/commit/150788e85fccdc36fd056d8774cb7c3887300ba3)]

Unfortunately you can't omit specific commits when you merge a branch with master - the complete commit history becomes part of the master branch when you merge. As long as the unnecessary files are deleted in the most recent commit they won't be part of the available code though.

modules/oncodriveclustl/1.0.1/config/default.yaml
```diff
@@ -0,0 +1,63 @@
...
+    oncodriveclustl:
+
+        dirs:
+            _parent:
+
+        inputs:
+            samples:
+            maf_path: # Leave blank if using sample subset
+
+        scratch_subdirectories: [] # Recommended: mutect2
```
lkhilton: Is the comment here carried over from another module?

modules/oncodriveclustl/1.0.1/oncodriveclustl.smk
```diff
@@ -0,0 +1,428 @@
...
+config["pipeline_name"] = "oncodriveclustl"
+
+include: "header_2.1.smk"
+
+config["lcr-modules"]["oncodriveclustl"]["dirs"]["_parent"] = config['lcr-modules']['_shared']['root_output_dir'] + "oncodriveclustl_module_test/"
+
+#### FOR CHECKING MODULE RUNS:
+RUN_SAMPLES = PIPELINE_SAMPLES["SAMPLES"]["oncodriveclustl"]
+#### add this to code below? kinda necessary to run?
+####
```
lkhilton: This stuff is all GAMBL-specific and really belongs in the Snakefile that you would use to launch this module. In that Snakefile you would store the samples table in the config as follows: 
```
config["lcr-modules"]["oncodriveclustl"]["samples"] = PIPELINE_SAMPLES["SAMPLES"]["oncodriveclustl"]
```

Then here in the module you can create the `RUN_SAMPLES` variable like this: 
```
RUN_SAMPLES = CFG["samples']
```

modules/oncodriveclustl/1.0.1/config/default.yaml
```diff
@@ -0,0 +1,63 @@
...
+
+        options:
+            mode: pathology  # pathology, samples, subset, maf           
+            genome_build: "grch37"
+            seq_type: ["genome","capture"]
+            
+            pathology: [["DLBCL","BL"],["FL","MCL","CLL"]]
```
lkhilton: This is going to require `pathology` to be a column in the input metadata frame. Could you please add a `pathology-1.0.yaml` schema to the `schema` directory in your module directory? There's an example in the `star/1.4` module for how to do this.

modules/oncodriveclustl/1.0.1/config/default.yaml
```diff
@@ -0,0 +1,63 @@
...
+        pairing_config:
+            genome:
+                run_paired_tumours: False
+                run_unpaired_tumours_with: "no_normal"
+                run_paired_tumours_as_unpaired: True
+            capture:
+                run_paired_tumours: False
+                run_unpaired_tumours_with: "no_normal"
+                run_paired_tumours_as_unpaired: True
+            mrna:
```
lkhilton: I doubt you would run this module on mrna data, so you could drop this from your pairing config.

lkhilton (commented):
I've made some comments but I see there are duplicate files in here. Can you please delete the ones I'm not supposed to see, then commit and push those changes? If you want to keep the files you can move them out of your directory, commit the deletion, then move them back and leave them untracked after moving them back.

mannycruz:
Thanks Laura! I moved the first draft version of my pipeline out of the module directory and committed, now this pull request should just contain the most recent 'almost finalized' version

modules/oncodriveclustl/1.0/config/default.yaml
```diff
@@ -0,0 +1,45 @@
...
+            hg19: "__UPDATE__"
+            hg38: "__UPDATE__"
+
+        scratch_subdirectories: []
+
+        options:
+            clustl: ""
+
+        conda_envs:
+            prepare_mafs: "{REPODIR}/envs/gatk/gatkR.yaml"
```
lkhilton, line 53: The typical way to handle conda envs is to symlink them into the module directory and reference them with the `MODSDIR` shortcut instead of `REPODIR`. However since the yaml file for this conda env is packaged with the script, I think it would be better to leave this conda env file out of the LCR-modules directory system and instead refer to its location in the `SCRIPTSDIR`.

modules/oncodriveclustl/1.0/config/default.yaml
```diff
@@ -0,0 +1,45 @@
...
+        inputs:
+            # Available wildcards: {seq_type} {genome_build}
+            master_maf: "__UPDATE__"
+            sample_sets: "__UPDATE__"
+
+        sample_set: "__UPDATE__"
+        genome_build: "__UPDATE__"
+        seq_types: ["genome","capture"]
+        prepare_mafs: "{SCRIPTSDIR}/generate_smg_inputs/1.0/generate_smg_inputs.R"
+        include_non_coding: False # True or False
```
lkhilton: The structure of these config values doesn't match what happens in most other modules. Scripts should be contained under a `scripts` key; the `sample_set` file should be under `inputs`, and things that are simple parameters (e.g. `sample_sets`, `genome_build`) should really be under `options`. 

Also what if a user wants different genome builds per sample_set? Perhaps a good way to structure this is as key:value pairs, e.g.: 
```
lcr-modules:
    oncodriveclustl:
        options: 
            sample_sets: 
                fl--dlbcl: "grch37"
```

It would also be really helpful to point the user to some documentation on how to configure the sample_sets file for compatibility with the module if it's already documented elsewhere.

lkhilton: Actually, ideally the seq_types would be captured under this key:value pair model as well... something to think about how to structure.

lkhilton (changes requested):
Looks great! I have one suggestion for how to improve the generalizability of the sample_sets and some requests regarding the config structure.

mannycruz:
Created a new branch with updated OncodriveCLUSTL code because this branch is behind master by too many commits

PR #244, opened 2022-08-02
dNdS module for SMGs

Kdreval:
# Pull Request Checklists

### Required

  ```bash
  ```


modules/dnds/1.0/config/default.yaml
```diff
@@ -0,0 +1,42 @@
...
+        options:
+            target_genes: "" # optionally, specify path to the genes of interest if this is a targeted sequencing
+                            # this will restrict SMG search to only these genes
+                            # this is a tsv file that contain one gene symbol/row and is expected to be without header
+            max_muts_per_gene_per_sample: Inf # How many mutations/gene dNdS should consider?
+                                            # This default value will consider all mutations/gene.
+                                            # When modified, this should be integer number.
+
+        conda_envs:
+            prepare_mafs: "{REPODIR}/envs/gatk/gatkR.yaml"
```
Jwong684, line 24: is REPODIR a new variable now?

Kdreval: Not exactly, it is just not used very often. This is a shortcut to the parent directory of lcr-modules. More details here: https://lcr-modules.readthedocs.io/en/latest/for_users.html?highlight=repodir#directory-placeholders

modules/dnds/1.0/src/R/run_dnds.R
```diff
@@ -0,0 +1,70 @@
...
+# Load packages -----------------------------------------------------------
+message("Loading packages...")
+suppressWarnings(
+suppressPackageStartupMessages({
+    library(dndscv)
+})
+)
+
+# Read the maf file
+maf = read.table(
```
Jwong684, line 22: would fread be better in this case?

Jwong684 (approved):
Seems good

modules/dnds/1.0/src/R/run_dnds.R
Kdreval: I opted for the base R alternative here to reduce the number of dependencies and make the conda environment more light. But in Rstudio will definitely opt for fread or read_tsv 😄

PR #246, opened 2022-08-29
Module/nanomethphase/1.0

nthomas11:
# Pull Request Checklists

### Required

- [x ] I used the cookiecutter template and updated the placeholder rules.

- [ x] The snakemake rules follow the [design guidelines](https://lcr-modules.readthedocs.io/en/latest/for_developers.html#module-rules). 

  - [x ] All references to the `rules` object (_e.g._ for input files) are wrapped with `str()`.

- [x ] Every rule in the module is either listed under `localrules` or has the `threads` and `resources` directives.

- [ x] Input and output files are being symlinked into the `CFG["inputs"]` and `CFG["outputs"]` subdirectories, respectively.

- [x ] I grouped the input symlinking rule to the next job that uses the input files. 

- [x ] I updated the final target rule (`*_all`) to include every output rule.

- [x ] I explained important module design decisions in `CHANGELOG.md`.

- [x ] I tested the module on real data for all supported `seq_type` values.

- [x ] I updated the `default.yaml` configuration file to provide default values for each rule in the module snakefile.

- [x ] I did not set any global wildcard constraints. Any/all wildcard constraints are set on a per-rule basis. 

- [x ] I ensured that all symbolic links are relative and self-contained (_i.e._ do not point outside of the repository).

- [ x] I replaced every value that should (or might need to) be updated in the default configuration file with `__UPDATE__`.

- [x ] I recursively searched for all comments containing `TODO` to ensure none were left. For example:

  ```bash
  ```

### If applicable

- [x ] I added more granular output subdirectories.

- [NA ] I added rules to the `reference_files` workflow to generate any new reference files.

- [x ] I added subdirectories with large intermediate files to the list of `scratch_subdirectories` in the `default.yaml` configuration file.

- [x ] I updated the list of available wildcards for the input files in the `default.yaml` configuration file.

Example:
```
```


modules/nanomethphase/1.0/config/default.yaml
```diff
@@ -0,0 +1,39 @@
...
+            nanomethphase: 50
+            clair3: 50
+            
+        mem_mb:
+            methyl_call: 25000 
+            clair3: 25000
+            whatshap: 25000
+            
+        clair3:
+            model: "{CONDA_PREFIX}/bin/models/{model_name}" #Update to conda prefix of Clair3 environment and select appropriate model recommended r941_prom_hac_g360+g422 for GSC bams
```
lkhilton: This can be pretty hard to track down because of the way Snakemake names Conda directories. 

In the vcf2maf module we have a function that identifies the path to the GAMBLR conda environment: 

```
import hashlib
# Obtain the path to the GAMBLR conda environment
md5hash = hashlib.md5()
if workflow.conda_prefix: 
    conda_prefix = workflow.conda_prefix
else: 
    conda_prefix = os.path.abspath(".snakemake/conda")

md5hash.update(conda_prefix.encode())
f = open("config/envs/GAMBLR.yaml", 'rb')
md5hash.update(f.read())
f.close()
h = md5hash.hexdigest()
GAMBLR = glob.glob(conda_prefix + "/" + h[:8] + "*")[0]
```

You could use this to automagically retrieve the path to the Conda env, then the config value would only have to be the model name.

modules/nanomethphase/1.0/nanomethphase.smk
```diff
@@ -0,0 +1,238 @@
...
+
+rule _filter_clair3:
+    input:
+        vcf = str(rules._clair3.output.vcf)
+    output: 
+        filtered = CFG["dirs"]["filter_clair3"] + "{seq_type}--{genome_build}/{sample_id}.filtered_phased_merged.vcf.gz"
+    log:
+        stderr = CFG["logs"]["filter_clair3"] + "{seq_type}--{genome_build}/{sample_id}/filter_clair3.stderr.log"      
+    shell :
+        "zcat {input.vcf} | awk '$6 > 20' | gzip > {output.filtered}"
```
lkhilton: Is this value of 20 something that could/should be user-configurable via the config?

lkhilton (changes requested):
Looks fantastic, Nicole! One question and one request re: the `{CONDA_PREFIX}` for the clair3 model.

lkhilton:
Now that I've reviewed the WhatsHap module, I see that you're running Clair3 and WhatsHap as part of both modules. Does it make sense to integrate the WhatsHap module and its outputs into this module so the rules and output files aren't duplicated?

modules/nanomethphase/1.0/config/default.yaml
nthomas11: Clair3 has been moved to phase_vcf module to avoid duplicating vcf files used by this module and whatshap module

lkhilton (approved):
Looks good!

PR #247, opened 2022-08-31
Add and implement CI unit tests

ckrushton:
# Lcr-modules continuous integration (CI) via Github actions

## About
This document briefly outlines the automatic unit tests configured for lcr-modules. This should reduce the likelihood that a module is merged into master in a broken state, and improve user experience by ensuring that a functional test setup is always available.

## How it works
The continuous integration (CI) tests are implemented via Github Actions, and are triggered every time a push is made to the lcr-modules repository. These tests are executed by a “runner” (i.e. a computer/server), which clones the repository and associated dependencies, sets up the environment, and perfoms the specified tests. These tests include:
1)	Execute a dry run of the capture demo workflow
2)	Run the capture demo workflow

To minimize runtime, a synthetic Tumour-Normal pair is used representing ultra-deep targeted sequencing.

NOTE: The pathseq module is currently not tested, as the bacterial database takes an exceptionally long time to prepare.  Also note that the test environment state is NOT cleaned between runs (i.e. python, snakemake, and oncopipe installs will persist, and packages used via snakemake conda environments will be cached)

## Checking the test results

The result of all tests can be found under Actions > lcr-modules on Github: 
https://github.com/LCR-BCCRC/lcr-modules/actions/workflows/lcr-modules-test-actions.yml. 
Click on a given test for more details about what was executed, and what steps failed.

### Configuring the runner
The CI tests are executed by a self-hosted runner (see https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners). This is both free and allows for improved flexibility compared to cloud-based runners. It is highly recommended that a fresh user account for the runner be created on the desired server. Conda and python should also be installed for this user account, due to MANY painful 
limitations of Github Actions

Currently, the runner is hosted on Thanos (thanos.mbb.sfu.ca), under the user account `github-runner`

To configure the runner, run `actions-runner/config.sh` (this only needs to be done once), then start the runner using `actions-runner/run.sh`. It is highly recommended that the latter command be initialized in `screen`

### The configuration
The steps the runner executes are configured via Github Actions, which is extensively documented (https://docs.github.com/en/actions).  The corresponding actions file for these tests can be found under `.github/workflows/lcr-modules-test-actions.yml`. 

Note the ”runs-on” tags should match those assigned to the runner during runner configuration.

### Demo data
Demo data, including sample BAM files and capture regions, can be found in https://github.com/LCR-BCCRC/lcr-modules-demo. The sample config and snakemake config for these tests files are found under `demo/capture_config_ci.yaml` and `demo/data/samples_ci.tsv`.

### The results
All tests are executed under `~/actions-runner/runner-work/lcr-modules/lcr-modules/`, which includes clones of all three repositories (lcr-modules, lcr-scripts, and lcr-modules-demo). If the tests fails, the corresponding logs can be found under 
`/home/github-runner/actions-runner/runner-work/lcr-modules/lcr-modules/lcr-modules/demo/.snakemake`. If the tests are successful, the `results` folder iscleared.

## Other info
Note this PR all contains bugfixes for:
- cnvkit: Fixed a bug where cnvkit would crash if a genome build was not specified for the capture space BED file
- qc: Now explicitly provides the reference genome FASTA to `samtools stats`, which aids in decompressing CRAM files if the reference is not cached
- LoFreq; Now includes a pre-compiled binary that works on Thanos (used for CIs).


lkhilton:
I actually think @Kdreval is better suited to review this because he's spent more time with the demo and on trying unit tests than I have.

ckrushton:
I've made quite a few performance improvements and bugfixes (LoFreq now works on different OS's), and have the runner working entirely in a docker container. I plan to squash these commits before merging anything into master, but let me know if these changes look good

.github/workflows/README.md
```diff
@@ -0,0 +1,58 @@
+# Lcr-modules continuous integration (CI) via Github actions
+
+HELLO HAYA AND ADAM
```
ckrushton: Note to self: Remove this

.github/workflows/lcr-modules-test-actions.yml
```diff
@@ -0,0 +1,60 @@
+name: lcr-modules
+on: [push]  # When should the tests be run? "push" is following every push to github
```
Kdreval: can we change it to be triggered at the PR?

.github/workflows/lcr-modules-test-actions.yml
```diff
@@ -0,0 +1,60 @@
...
+jobs:
+    run-capture-demo:
+        # INCLUDE SELF-HOSTED, OTHERWISE IT USES THE CLOUD AND WILL START COSTING US $$$$
+        # These tags are specified by the runner
+        runs-on: [self-hosted, linux]
+        steps:
+            # SETUP SOFTWARE
+            # Clone repos
+          - name: Checkout lcr-modules repo
+            uses: actions/checkout@v3
```
Kdreval, line 21: What does the `v3` part stands for here?

.github/workflows/lcr-modules-test-actions.yml
```diff
@@ -0,0 +1,60 @@
...
+            with:
```
Kdreval, line 22: so in this `with` we don't need the `repository` key? Is it because it is "knows about itself" by default?

.github/workflows/lcr-modules-test-actions.yml
```diff
@@ -0,0 +1,60 @@
...
+                repository: LCR-BCCRC/lcr-scripts
+                path: lcr-scripts
+            # As the github actions for installing conda are GARBAGE, here we will use a user-account install
+            # I am assuming the user has installed miniconda3 in their home directory
+          - name: Setup conda and mamba
+            run: $HOME/miniconda3/bin/conda install -c conda-forge mamba 'python>3.8'
+            # Install python dependencies
+          - name: Install snakemake
```
Kdreval: This step we can drop because snakemake is installed by oncopipe

.github/workflows/lcr-modules-test-actions.yml
```diff
@@ -0,0 +1,60 @@
...
+            run: pip install 'snakemake>6'
+          - name: Install oncopipe
+            run: pip install -e $GITHUB_WORKSPACE/lcr-modules/oncopipe/
```
Kdreval, line 35: Is the `$GITHUB_WORKSPACE` defined by Thanos as part of that `run.sh` script?

.github/workflows/lcr-modules-test-actions.yml
```diff
@@ -0,0 +1,60 @@
...
+          - name: Symlink reference directory
+            run: ln -s $LCR_MODULE_REF $GITHUB_WORKSPACE/lcr-modules/demo/reference
+
+            # SETUP DEMO DATA
+          - name: Download demo data
+            uses: actions/checkout@v3
+            with:
+                    repository: LCR-BCCRC/lcr-modules-demo
```
Kdreval, line 43: Can you please make sure someone else has permissions to edit/manage this repo? In case it will need to be modified in the future :smile:

.github/workflows/python-publish.yml
```diff
@@ -1,32 +0,0 @@
-# This workflows will upload a Python Package using Twine when a release is created
```
Kdreval, line 1: My understanding is that this PR replaces what was there before, so this file can be dropped?

demo/capture_Snakefile.smk
```diff
@@ -36,7 +36,7 @@ configfile: "../modules/sequenza/1.4/config/default.yaml"
 configfile: "../modules/bwa_mem/1.1/config/default.yaml"
```
Kdreval: What if we keep the capture Snakefile as is, but create something like cli_Snakefile and associated config for the purpose of this testing?

envs/r-sequenza/r-sequenza-3.0.0.yaml
```diff
@@ -12,7 +12,7 @@ dependencies:
   - bioconductor-biocgenerics=0.34.0
   - bioconductor-copynumber=1.28.0
   - bioconductor-genomeinfodb=1.24.0
-  - bioconductor-genomeinfodbdata=1.2.3
```
Kdreval, line 15: This will trigger re-run of all sequenza results in GAMBL because the change to env file will be recognized and snakemake will want to build new environment. Can we instead here have this environment as separate file, which will become default in the config of sequenza module?

modules/lofreq/1.0/lofreq.smk
```diff
@@ -113,6 +113,7 @@ rule _lofreq_run:
         else echo "WARNING: PATH is not set properly, using $(which lofreq2_call_pparallel.py)"; fi
         """)
 
+
```
Kdreval: this line can probably be dropped so we introduce less changes to the master

modules/lofreq/1.1/lofreq.smk
```diff
@@ -82,6 +83,25 @@ def _lofreq_get_capspace(wildcards):
 
 ##### RULES #####
 
+# Setup LoFreq to use system-specific binaries, but custom python scripts
+# By default, the LoFreq binary (from my understanding) uses python scripts in the same directory as the binary
+# So lets put the conda binary with our python scripts
+rule _lofreq_configure_software:
```
Kdreval, line 89: I think I am not following here how it will fall back to the default lofreq-provided somatic and call_parallel scripts if it can't find custom binaries. The behaviour that was currently in place is that if there is no custom binary, the default lofreq is used (with a warning message `"WARNING: PATH is not set properly, using $(which lofreq2_call_pparallel.py)"`). This and the other rules downstream seems to be missing this functionality

modules/lofreq/1.1/src/lofreq2_somatic.py
```diff
@@ -377,7 +377,6 @@ def call_rlx(self, sample_type):
                 #return self.num_tests_from_log(elines)
             else:
                 assert not os.path.exists(out_log)
-        print(f"HERE! {cmd}")
```
Kdreval: I think this was an important line to troubleshoot and last time there was a discussion on Slack it was important to see this message. Can we keep `HERE` here? :smile:

modules/slms_3/1.0/slms_3.smk
```diff
@@ -86,72 +87,94 @@ for tool in CFG_SLMS3["module_versions"].keys():
...
     output:
-        vcf = CFG_SLMS3["dirs"]["inputs"] + "vcf/{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}.lofreq.snvs.vcf.gz", 
+        vcf = CFG_SLMS3["dirs"]["inputs"] + "vcf/{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}.lofreq.snvs.vcf.gz",
         tbi = CFG_SLMS3["dirs"]["inputs"] + "vcf/{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}.lofreq.snvs.vcf.gz.tbi"
     run:
         op.relative_symlink(input.vcf, output.vcf, in_module = True)
         op.relative_symlink(input.vcf + ".tbi", output.tbi, in_module = True)
 
 # Annotate Strelka VCF and remove common GnomAD variants
+# bcftools annotate is very slow if one of the VCF files is quite large (see https://github.com/samtools/bcftools/issues/1199#issuecomment-624713745)
```
Kdreval, line 148: Feel like this is not a problem when the cluster is used? Maybe we should drop this to avoid possible bugs in the future and avoid possible effect of snakefile modification on the samples already run through SLMS-3? This will also revert all modifications in SLMS-3 to master version

workflows/reference_files/2.4/reference_files_header.smk
```diff
@@ -76,11 +76,17 @@ for build_name, build_info in config["genome_builds"].items():
         f"`provider` not set for `{build_name}` or `provider` not among {possible_providers}."
     )
     if "genome_fasta_url" in build_info:
```
Kdreval, line 78: this part has been a problem needing an update already for long! Since you modified it here, can you please instead drop the whole url_code ping?

workflows/reference_files/2.4/reference_files_header.smk
```diff
@@ -97,11 +103,15 @@ for build_name, build_info in config["capture_space"].items():
     assert "genome" in build_info and build_info["genome"] in possible_versions,(
         f"`genome` not set for `{build_name}` or `genome` not among {possible_versions}." )
     if "capture_bed_url" in build_info: 
-        url_code = urllib.request.urlopen(build_info["capture_bed_url"]).getcode()
-        assert url_code == 200, (
-            f"Pinging `capture_bed_url` for {build_name} returned HTTP code {url_code} "
-            f"(rather than 200): \n{build_info['capture_bed_url']}"
-            )
+        try:
```
Kdreval, line 106: same here

demo/data/samples_ci.tsv
```diff
@@ -0,0 +1,3 @@
+sample_id	seq_type	patient_id	tissue_status	genome_build	strand	read_length	capture_space
```
Kdreval: The snakefile reading these samples is missing I think

demo/capture_config_ci.yaml
```diff
@@ -0,0 +1,100 @@
+lcr-modules:
```
Kdreval: I think there is no Snakefile that uses this config

Kdreval (commented):
Thanks Chris!

.github/workflows/lcr-modules-test-actions.yml
ckrushton: This uses a github action https://github.com/actions/checkout to check out the repos. @v3 is effectively the "version" of that repo

.github/workflows/lcr-modules-test-actions.yml
ckrushton: Exactly! By default the action refers to the "current" workflow

.github/workflows/lcr-modules-test-actions.yml
ckrushton: Aye, this is configured by the action-runner indicating the location of the "work" directory

.github/workflows/python-publish.yml
ckrushton: Aye, this file has been deleted entirely in the PR

.github/workflows/lcr-modules-test-actions.yml
ckrushton: Fixed, although I currently still have this as "push" so I can test that the CI still works after my tweaks

.github/workflows/lcr-modules-test-actions.yml
ckrushton: Done

modules/slms_3/1.0/slms_3.smk
ckrushton: As the entire dbSNP file is iterated through every time a VCF is annotated, this would definitely also be an issue on the GSC, perhaps even more so due to increased I/O bottlenecks. This probably hasn't been noticed as 1) The runtime of this step is relatively small when running exome or genome data through other pipeline steps, and 2) There are more variants, and over a wider range, compared to CAPPSeq data. This is most noticeable when running samples with only a handful (ex. 12) SNV calls for annotation, as this step is actually the bottleneck for both the CI and running capture data through

ckrushton: I did test this to help ensure nothing is dropped, but I was unable to test this fully, as the annotation step also includes filtering steps. Will do one final comparison between the old version and new version

envs/r-sequenza/r-sequenza-3.0.0.yaml
ckrushton: I'm not sure I totally follow. If we create a separate file for this environment and use that as the default for this rule, won't we run into the same problem?

Kdreval: No, because the GAMBL-specific config can keep referring to the yml file for the environment that is already built.

modules/lofreq/1.1/lofreq.smk
ckrushton: It won't fall back to the default binaries, since the python scripts packaged with LoFreq don't have some of the modifications we've made (ex. the --normal-only). As it appears the LoFreq binary appears to use the python scripts located in the same directory as the LoFreq binary by default (from my understanding, could be wrong), the "general" binary has to be copied to the same location as these modified scripts

ckrushton: P.S. the "default binary" portion wouldn't work, as it would just print that message (WARNING: PATH is not set properly, using $(which lofreq2_call_pparallel.py)") and then "complete" the rule

modules/lofreq/1.1/src/lofreq2_somatic.py
ckrushton: Added back

workflows/reference_files/2.4/reference_files_header.smk
ckrushton: I've deleted all the HTTP 200 checks

demo/capture_config_ci.yaml
ckrushton: I've modified the structure of the CI tests to be self-contained in their own folder (ci), so its easier to understand and follow

modules/lofreq/1.1/lofreq.smk
Kdreval: Ah, I see - you are right! There was nothing after the message, so the "default" lofreq will not run anyways. Can you at least please include the echo statement which will print the $PATH and $SCRIPT for future debugging purposes?

Kdreval:
will close this for now as the CI has to be re-thinked to not rely on thanos

PR #249, opened 2022-09-13
Phylogenetics tools! 

lkhilton:
# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable


envs/phylowgs/fill_battenberg.yaml
```diff
@@ -0,0 +1,41 @@
+name: fill_segments
```
Kdreval: Can this be a symlink to the yaml file in lcr-scripts? I think they are same recipes
https://github.com/LCR-BCCRC/lcr-scripts/blob/master/fill_segments/1.0/fill_segments.yaml

modules/phylowgs/1.0/etc/noncoding.txt
```diff
@@ -0,0 +1,8 @@
+Hugo_Symbol
```
Kdreval, line 1: Just curious why is the Hugo_Symbol here?

modules/phylowgs/1.0/phylowgs.smk
```diff
@@ -0,0 +1,391 @@
...
+
+# Symlinks the input files into the module results directory (under '00-inputs/')
+rule _phylowgs_input_maf:
+    input:
+        maf = CFG["inputs"]["maf"],
+    output:
+        maf = CFG["dirs"]["inputs"] + "maf/{seq_type}--{genome_build}/{tumour_id}--{normal_id}--{pair_status}.maf",
+    group: "input_maf"
+    run:
+        op.absolute_symlink(input.maf, output.maf)
```
Kdreval, line 199: I think absolute_symlink needs an oncopipe version check at the top of the snakefile? It is the one introsuced in the version 11 I think and was not supported earelier - but probably it is not so important because I don't think anyone will deliberately install old version of oncopipe 😄

modules/phylowgs/1.0/etc/noncoding.txt
```diff
@@ -0,0 +1,8 @@
...
+Silent
+RNA
+IGR
+Intron
+5'Flank
+3'Flank
+5'UTR
```
Kdreval, line 8: Should the 3'UTR be included here?

modules/phylowgs/1.0/phylowgs.smk
```diff
@@ -0,0 +1,391 @@
...
+if workflow.conda_prefix:
+    conda_prefix = workflow.conda_prefix
+else:
+    conda_prefix = os.path.abspath(".snakemake/conda")
+md5hash.update(conda_prefix.encode())
+f = open(CFG['conda_envs']['phylowgs'], 'rb')
+md5hash.update(f.read())
+f.close()
+h = md5hash.hexdigest()
+PHYLO = "".join(glob.glob(conda_prefix + "/" + h[:8] + "*/share/phylowgs/"))
```
Kdreval, line 93: I think this part only works when the environment is built with conda, since the directories are stored with the 8-digit code from md5 hash. When it is created with mamba, I think the full string `h` is needed here. Did you happen to know if the env was created with mamba or conda?

modules/phylowgs/1.0/phylowgs.smk
```diff
@@ -0,0 +1,391 @@
...
+         CFG["conda_envs"]["phylowgs"]
+    threads:
+        CFG["threads"]["create_inputs"]
+    resources:
+        **CFG["resources"]["create_inputs"]
+    group: "input_battenberg"
+    shell:
+        op.as_one_line("""
+        cellularity=$(tail -n +2 {input.cellularity} | cut -f 1);
+        python2 {params.script} -f battenberg-smchet -c $cellularity --cnv-output {output.txt} {input.subclones}
```
Kdreval, line 236: python2 😱

modules/phylowgs/1.0/phylowgs.smk
```diff
@@ -0,0 +1,391 @@
...
+        ssms = CFG["dirs"]["preprocess_inputs"] + "{seq_type}--{genome_build}/{patient_id}/ssm_data.txt",
+        cnvs = CFG["dirs"]["preprocess_inputs"] + "{seq_type}--{genome_build}/{patient_id}/cnv_data.txt",
+        params = CFG["dirs"]["preprocess_inputs"] + "{seq_type}--{genome_build}/{patient_id}/params.json"
+    log:
+        stderr = CFG["logs"]["preprocess_inputs"] + "{seq_type}--{genome_build}/{patient_id}/create_phylowgs_inputs.stderr.log",
+        stdout = CFG["logs"]["preprocess_inputs"] + "{seq_type}--{genome_build}/{patient_id}/create_phylowgs_inputs.stdout.log"
+    params:
+        cli = create_phylowgs_inputs_cli,
+        opts = CFG["options"]["create_inputs"]["opts"],
+        sex = lambda w: config["lcr-modules"]["phylowgs"]["switches"]["sex"][op.filter_samples(PATIENTS, tumour_patient_id = w.patient_id)["tumour_sex"].values[0]] if op.filter_samples(PATIENTS, tumour_patient_id = w.patient_id)["tumour_sex"].values[0] in config["lcr-modules"]["phylowgs"]["switches"]["sex"].keys() else "auto",
```
Kdreval, line 298: Do you think this can be a separate designated function?

modules/phylowgs/1.0/src/process_phyloWGS_outputs.R
```diff
@@ -0,0 +1,536 @@
...
+mut_file = snakemake@input[["muts"]]
+driver_genes = snakemake@params[["drivers"]]
+sample_order = unlist(strsplit(snakemake@params[["sample_order"]], ","))
+genome_build = snakemake@wildcards[["genome_build"]]
+
+# Define the chr_prefix parameter based on the genome_build
+chr_prefixed = str_detect(genome_build, "hg")
+
+
+# option_list = list(
```
Kdreval: This section probably can go?
(and another section below at line 112)

modules/phylowgs/1.0/src/process_phyloWGS_outputs.R
```diff
@@ -0,0 +1,536 @@
...
+tree_structure_long$num_ssms <- get_ssms(tree_structure_long, best_focal, best_tree_fileID)[as.character(tree_structure_long$node)]
+
+tree_structure_long <- tree_structure_long %>% 
+  mutate(parent = as.numeric(parent)) %>%
+  left_join(select(tree_structure_long, node, xstart = x, ystart = y), 
+            by = c("parent" = "node"))
+
+
+
+ggplot(tree_structure_long,
```
Kdreval: How does this run on the command line without X-forwarding? Not sure if this errors out because it is not saved to a variable here and tries to open graphic device, or it is just goes through? I think `ggsave` below needs plot to be displayed before it can be saved

modules/pyclone_vi/1.0/config/default.yaml
```diff
@@ -0,0 +1,64 @@
+lcr-modules:
+
+    pyclone_vi:
+
+        # TODO: Update the list of available wildcards, if applicable
+        inputs:
+            # Available wildcards: {seq_type} {genome_build} {tumour_id} {normal_id} {pair_status}
+            sample_maf: "__UPDATE__"
+            sample_subclones: "__UPDATE__"
```
Kdreval: Can the commented example of file path with wild cards be used here, like in the phylowgs config?

modules/pyclone_vi/1.0/envs/gamblr.yaml
```diff
@@ -0,0 +1 @@
+/home/lhilton/repos/lcr-modules-starfish/envs/GAMBLR/gamblr.yaml
```
Kdreval, line 1: This seems to point to home directory. Maybe this file needs to be updated (or removed)?

modules/pyclone_vi/1.0/pyclone_vi.smk
```diff
@@ -0,0 +1,397 @@
...
+# Setup module and store module-specific configuration in `CFG`
+# `CFG` is a shortcut to `config["lcr-modules"]["pyclone_vi"]`
+CFG = op.setup_module(
+    name = "pyclone_vi",
+    version = "1.0",
+    subdirectories = ["inputs", "build_inputs", "fit", "og-pyclone", "phyclone", "outputs"],
+)
+
+# Define rules to be run locally when using a compute cluster
+# TODO: Replace with actual rules once you change the rule names
```
Kdreval: TODO here that probably can be dropped?

modules/pyclone_vi/1.0/pyclone_vi.smk
```diff
@@ -0,0 +1,397 @@
...
+                str(rules._pyclone_vi_output_tsv.output.phyclone),
+                str(rules._pyclone_vi_output_tsv.output.pyclone),
+                str(rules._pyclone_vi_output_tsv.output.tree),
+                str(rules._pyclone_vi_output_tsv.output.clusters)
+            ],
+            zip,  # Run expand() with zip(), not product()
+            seq_type=PATIENTS["tumour_seq_type"],
+            genome_build=PATIENTS["tumour_genome_build"],
+            patient_id=PATIENTS["tumour_patient_id"])
+        # expand(
```
Kdreval: can probably be removed?

modules/pyclone_vi/1.0/pyclone_vi.smk
```diff
@@ -0,0 +1,397 @@
...
+        phyclone = str(rules._pyclone_vi_compute_tree_stats.output.stats),
+        tree = str(rules._pyclone_vi_phyclone_consensus.output.tree),
+        clusters = str(rules._pyclone_vi_phyclone_consensus.output.clusters),
+    output:
+        pyclone = CFG["dirs"]["outputs"] + "{seq_type}--{genome_build}/{patient_id}.pyclone.results.tsv",
+        phyclone = CFG["dirs"]["outputs"] + "{seq_type}--{genome_build}/{patient_id}.phyclone.stats.tsv",
+        tree = CFG["dirs"]["outputs"] + "{seq_type}--{genome_build}/{patient_id}.phyclone.tree.nwk",
+        clusters = CFG["dirs"]["outputs"] + "{seq_type}--{genome_build}/{patient_id}.phyclone.clusters.tsv"
+    run:
+        op.relative_symlink(input.pyclone, output.pyclone)
```
Kdreval: Do you want to add in_module=True here?

Kdreval (commented):
Thanks! Important update!

envs/phylowgs/fill_battenberg.yaml
lkhilton: I removed this since it isn't referenced in the module any longer now that the Battenberg module outputs filled seg files.

modules/phylowgs/1.0/etc/noncoding.txt
lkhilton: It's used to remove the header from the output file that's generated by the rule that uses this.

modules/phylowgs/1.0/phylowgs.smk
lkhilton: Added!

modules/phylowgs/1.0/etc/noncoding.txt
lkhilton: Probably. Added!

modules/phylowgs/1.0/phylowgs.smk
lkhilton: The glob takes care of this to make sure it works with either conda- or mamba-installed environments 😄

modules/phylowgs/1.0/phylowgs.smk
lkhilton: :backaway:

modules/phylowgs/1.0/phylowgs.smk
lkhilton: It's a bit verbose but it's IMO a good use of a lambda function.

modules/phylowgs/1.0/src/process_phyloWGS_outputs.R
lkhilton: Implemented.

modules/pyclone_vi/1.0/config/default.yaml
lkhilton: Done.

modules/pyclone_vi/1.0/envs/gamblr.yaml
lkhilton: The prefix is ignored when building envs.

Kdreval:
@lkhilton  this is merge-able?

PR #253, opened 2022-10-25
Maf2maf support for variant re-annotation in vcf2maf module

Kdreval:
This PR introduces new functionality to re-annotate crossmaped mafs in the vcf2maf module. This will ensure the consistency within particular genome build of variant annotation, regardless whether the sample is aligned to this build initially or was crossmaped. I have:
1. Ensured the re-annotation actually works by checking the HIST1H3B gene (ENSG00000274267 vs ENSG00000124693).
2. Ensured the columns from original maf are propagated to the re-annotated version, and their order is the same, which should maintain existing downstream applications for maf file merging.

The current version of the vcf2maf family of scripts is different from what we have in our production conda environment, so I have bundled the compatible versions with the module. The bundled versions include some debugging features too - and also dropped the annoying `#version 2.4` first line of maf files.

Some variants are being lost during the re-annotation due to the differences in nucleotide sequence between grch37 and hg38. More details discussed on Slack [here](https://morinlabsfu.slack.com/archives/C0GF04TCL/p1666293881045829)

### DIFF
--- a/modules/vcf2maf/1.3/vcf2maf.smk
+++ b/modules/vcf2maf/1.3/vcf2maf.smk
@@ -402,8 +402,6 @@ rule _vcf2maf_reannotate:
         opts = CFG["options"]["maf2maf"],
         build = lambda w: VCF2MAF_VERSION_MAP[w.target_build],
         custom_enst = lambda w: "--custom-enst " + str(config['lcr-modules']["vcf2maf"]["switches"]["custom_enst"][VCF2MAF_GENOME_VERSION[w.target_build]]) if config['lcr-modules']["vcf2maf"]["switches"]["custom_enst"][VCF2MAF_GENOME_VERSION[w.target_build]] != "" else ""
-    #wildcard_constraints:
-    #    target_build = lambda w: "grch37" if "38" in str(get_original_genome(wildcards)[0]) else "hg38"
     conda:
         CFG["conda_envs"]["vcf2maf"]
     threads:
--- a/modules/vcf2maf/1.3/vcf2maf.smk
+++ b/modules/vcf2maf/1.3/vcf2maf.smk
@@ -28,11 +28,11 @@ CFG = op.setup_module(
 # Define rules to be run locally when using a compute cluster
 localrules:
     _vcf2maf_input_vcf,
-    _vcf2maf_input_bam, 
+    _vcf2maf_input_bam,
     _vcf2maf_output_maf,
-    _vcf2maf_gnomad_filter_maf, 
+    _vcf2maf_gnomad_filter_maf,
     _vcf2maf_install_GAMBLR,
-    _vcf2maf_output_original, 
+    _vcf2maf_output_original,
     _vcf2maf_normalize_prefix,
     _vcf2maf_all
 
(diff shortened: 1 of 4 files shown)


modules/vcf2maf/1.3/vcf2maf.smk
```diff
@@ -387,6 +387,59 @@ rule _vcf2maf_crossmap:
         """)
 
 
+# Re-annotate variants according to the new genome build
```
Kdreval, line 390: This is a new rule that does re-annotation of crossmaped mafs

modules/vcf2maf/1.3/vcf2maf.smk
```diff
@@ -397,44 +450,44 @@ def get_normalize_input(wildcards, todo_only = False):
...
         else:
             todo = "normalize"
     else:
         # Source doesn't match. CrossMap and normalization necessary.
-        target_build = "hg38" if new_genome_version == "grch38" else "hg19" 
+        target_build = "hg38" if new_genome_version == "grch38" else "hg19"
         maf = expand(
-            rules._vcf2maf_crossmap.output.maf, 
-            target_build = target_build, 
+            rules._vcf2maf_reannotate.output.maf,
```
Kdreval, line 478: Meaningful change here: instead of the output of crossmap rule, use the outputs after re-annotation

lkhilton:
> See there is lots of trailing space edits again - I have highlighted what was actually changed in the snakefile

No worries you can turn off white space changes in the diff viewer 👍🏻

Kdreval:
Yeah I probably have this in VS Code - they were not shown in diff but showed up on Github :sweat_smile:

modules/vcf2maf/1.3/vcf2maf.smk
```diff
@@ -387,6 +387,59 @@ rule _vcf2maf_crossmap:
...
+        maf = temp(CFG["dirs"]["crossmap"] + "{seq_type}--{target_build}/{tumour_id}--{normal_id}--{pair_status}/{base_name}.{filter}_reannotated.{maf}")
+    log:
+        stdout = CFG["logs"]["normalize"] + "{seq_type}--{target_build}/{tumour_id}--{normal_id}--{pair_status}/{base_name}.{filter}_{maf}_reannotated.stdout.log",
+        stderr = CFG["logs"]["normalize"] + "{seq_type}--{target_build}/{tumour_id}--{normal_id}--{pair_status}/{base_name}.{filter}_{maf}_reannotated.stderr.log",
+    params:
+        opts = CFG["options"]["maf2maf"],
+        build = lambda w: VCF2MAF_VERSION_MAP[w.target_build],
+        custom_enst = lambda w: "--custom-enst " + str(config['lcr-modules']["vcf2maf"]["switches"]["custom_enst"][VCF2MAF_GENOME_VERSION[w.target_build]]) if config['lcr-modules']["vcf2maf"]["switches"]["custom_enst"][VCF2MAF_GENOME_VERSION[w.target_build]] != "" else ""
+    #wildcard_constraints:
+    #    target_build = lambda w: "grch37" if "38" in str(get_original_genome(wildcards)[0]) else "hg38"
```
lkhilton: Is this still useful if you change it to just `target_build = "grch37|hg38"`?

lkhilton (dismissed):
I just had one question about the commented-out wildcard constraints, but otherwise approved.

modules/vcf2maf/1.3/vcf2maf.smk
Kdreval: ah that was from early troubleshooting - can go away in the clean version. Will drop it!

Kdreval: This is now fixed and the new commit dismisses the previous approval

lkhilton (approved):

PR #262, opened 2023-03-21
Module/minimap2/1.0

hayashaalan:
# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable

Example:
```
```

### DIFF
--- a/modules/minimap2/1.0/config/default.yaml
+++ b/modules/minimap2/1.0/config/default.yaml
@@ -6,21 +6,31 @@ lcr-modules:
         inputs:
             # Available wildcards: {seq_type} {genome_build} {sample_id}
             sample_fastq: "__UPDATE__"
+            reference_build: "__UPDATE__"
 
         scratch_subdirectories: []
 
         options:
-            step_1: ""
-            step_2: ""
+            minimap2: '-ax map-ont -L --MD -Y -R "@RG\tID:{sample_id}\tLB:{sample_id}\tPL:ONT\tSM:{sample_id}"'
+            samtools: '-bhS'
 
         conda_envs:
+            minimap2: "{MODSDIR}/envs/minimap2-2.24.yaml"
             samtools: "{MODSDIR}/envs/samtools-1.9.yaml"
             
         threads:
-            step_1: 4
+            minimap2: 4
+            samtools: 1
 
         resources:
-            step_1: 
-                mem_mb: 2000
+            minimap2: 
+                mem_mb: 30000
+            samtools:
+                mem_mb: 10000
             
         pairing_config:
+            promethION:
+                run_paired_tumours: False
+                run_unpaired_tumours_with: "no_normal"
+                run_paired_tumours_as_unpaired: True
+
--- a/modules/minimap2/1.0/config/default.yaml
+++ b/modules/minimap2/1.0/config/default.yaml
@@ -1,0 +1,26 @@
+lcr-modules:
+    
+    minimap2:
+
+        # TODO: Update the list of available wildcards, if applicable
+        inputs:
+            # Available wildcards: {seq_type} {genome_build} {sample_id}
+            sample_fastq: "__UPDATE__"
+
+        scratch_subdirectories: []
+
+        options:
+            step_1: ""
+            step_2: ""
+
+        conda_envs:
+            samtools: "{MODSDIR}/envs/samtools-1.9.yaml"
+            
+        threads:
+            step_1: 4
+
+        resources:
+            step_1: 
+                mem_mb: 2000
+            
+        pairing_config:
--- a/modules/minimap2/1.0/envs/minimap2-2.24.yaml
+++ b/modules/minimap2/1.0/envs/minimap2-2.24.yaml
@@ -1,0 +1,16 @@
+name: minimap2
+channels:
+  - conda-forge
+  - bioconda
+  - defaults
+dependencies:
+  - _libgcc_mutex=0.1
+  - _openmp_mutex=4.5
+  - k8=0.2.5
+  - libgcc-ng=12.2.0
+  - libgomp=12.2.0
+  - libstdcxx-ng=12.2.0
+  - libzlib=1.2.13
+  - minimap2=2.24
+  - zlib=1.2.13
+prefix: /home/hshaalan/miniconda3/envs/minimap2
--- a/modules/minimap2/1.0/minimap2.smk
+++ b/modules/minimap2/1.0/minimap2.smk
@@ -1,0 +1,141 @@
+#!/usr/bin/env snakemake
+
+
+##### ATTRIBUTION #####
+
+
+# Original Author:  Haya Shaalan
+# Module Author:    Haya Shaalan
+# Contributors:     N/A
+
+
+##### SETUP #####
+
+# Import package with useful functions for developing analysis modules
+import oncopipe as op
+
+# Check that the oncopipe dependency is up-to-date. Add all the following lines to any module that uses new features in oncopipe
+min_oncopipe_version="1.0.11"
+import pkg_resources
+try:
+    from packaging import version
+except ModuleNotFoundError:
+    sys.exit("The packaging module dependency is missing. Please install it ('pip install packaging') and ensure you are using the most up-to-date oncopipe version")
+
+# To avoid this we need to add the "packaging" module as a dependency for LCR-modules or oncopipe
+
+current_version = pkg_resources.get_distribution("oncopipe").version
+if version.parse(current_version) < version.parse(min_oncopipe_version):
+    print('\x1b[0;31;40m' + f'ERROR: oncopipe version installed: {current_version}' + '\x1b[0m')
+    print('\x1b[0;31;40m' + f"ERROR: This module requires oncopipe version >= {min_oncopipe_version}. Please update oncopipe in your environment" + '\x1b[0m')
+    sys.exit("Instructions for updating to the current version of oncopipe are available at https://lcr-modules.readthedocs.io/en/latest/ (use option 2)")
+
+# End of dependency checking section
+
+# Setup module and store module-specific configuration in `CFG`
+# `CFG` is a shortcut to `config["lcr-modules"]["minimap2"]`
+CFG = op.setup_module(
+    name = "minimap2",
+    version = "1.0",
+    # TODO: If applicable, add more granular output subdirectories
+    subdirectories = ["inputs", "minimap2", "outputs"],
+)
+
+# Define rules to be run locally when using a compute cluster
+# TODO: Replace with actual rules once you change the rule names
+localrules:
+    _minimap2_input_fastq,
+    _minimap2_step_2,
+    _minimap2_output_bam,
+    _minimap2_all,
+
+
+##### RULES #####
+
+
+# Symlinks the input files into the module results directory (under '00-inputs/')
+# TODO: If applicable, add an input rule for each input file used by the module
+# TODO: If applicable, create second symlink to .crai file in the input function, to accomplish cram support
+rule _minimap2_input_fastq:
+    input:
+        fastq = CFG["inputs"]["sample_fastq"]
+    output:
+        fastq = CFG["dirs"]["inputs"] + "fastq/{seq_type}--{genome_build}/{sample_id}.fastq"
+    group: 
+        "input_and_step_1"
+    run:
+        op.absolute_symlink(input.fastq, output.fastq)
+
+
+# Example variant calling rule (multi-threaded; must be run on compute server/cluster)
+# TODO: Replace example rule below with actual rule
+rule _minimap2_step_1:
+    input:
+        fastq = str(rules._minimap2_input_fastq.output.fastq),
+        fasta = reference_files("genomes/{genome_build}/genome_fasta/genome.fa")
+    output:
+        bam = CFG["dirs"]["minimap2"] + "{seq_type}--{genome_build}/{sample_id}/output.bam"
+    log:
+        stdout = CFG["logs"]["minimap2"] + "{seq_type}--{genome_build}/{sample_id}/step_1.stdout.log",
+        stderr = CFG["logs"]["minimap2"] + "{seq_type}--{genome_build}/{sample_id}/step_1.stderr.log"
+    params:
+        opts = CFG["options"]["step_1"]
+    conda:
+        CFG["conda_envs"]["samtools"]
+    threads:
+        CFG["threads"]["step_1"]
+    resources:
+        **CFG["resources"]["step_1"]    # All resources necessary can be included and referenced from the config files.
+    shell:
+        op.as_one_line("""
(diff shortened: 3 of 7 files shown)


modules/minimap2/1.0/config/default.yaml
```diff
@@ -0,0 +1,36 @@
+lcr-modules:
+    
+    minimap2:
+
+        # TODO: Update the list of available wildcards, if applicable
```
Kdreval: Can you address all the TODOs (here and elsewhere in the module) and then remove these commented lines?

modules/minimap2/1.0/config/default.yaml
```diff
@@ -0,0 +1,36 @@
...
+        inputs:
+            # Available wildcards: {seq_type} {genome_build} {sample_id}
+            sample_fastq: "__UPDATE__"
```
Kdreval: Are any indexes required for this tool to work? Maybe we should add them here as inputs as well?

modules/minimap2/1.0/config/default.yaml
```diff
@@ -0,0 +1,36 @@
...
+        options:
+            minimap2: '-ax map-ont -L --MD -Y -R "@RG\tID:{sample_id}\tLB:{sample_id}\tPL:ONT\tSM:{sample_id}"'
+            samtools: '-bhS'
+
+        conda_envs:
+            minimap2: "{MODSDIR}/envs/minimap2-2.24.yaml"
+            samtools: "{MODSDIR}/envs/samtools-1.9.yaml"
+            
+        threads:
+            minimap2: 4
```
Kdreval: Just wonder how long this tool takes to run and if we should consider giving it more resources here?

modules/minimap2/1.0/envs/minimap2-2.24.yaml
Kdreval, line 1: Van you please put this file under `lcr-modules/envs` (and create there designated directory for it  similar to other recipes) and then symlink here that file instead? So the set up is similar to how the samtools recipe is specified

modules/minimap2/1.0/minimap2.smk
```diff
@@ -0,0 +1,168 @@
...
+# Setup module and store module-specific configuration in `CFG`
+# `CFG` is a shortcut to `config["lcr-modules"]["minimap2"]`
+CFG = op.setup_module(
+    name = "minimap2",
+    version = "1.0",
+    # TODO: If applicable, add more granular output subdirectories
+    subdirectories = ["inputs", "minimap2", "sort_bam", "outputs"],
+)
+
+#include: "../../utils/2.1/utils.smk"
```
Kdreval: Not sure if this is supposed to be here or should be removed?

modules/minimap2/1.0/minimap2.smk
```diff
@@ -0,0 +1,168 @@
...
+    run:
+        op.absolute_symlink(input.fastq, output.fastq)
+
+
+rule _minimap2_run:
+    input:
+        fastq = str(rules._minimap2_input_fastq.output.fastq),
+        fasta = reference_files("genomes/{genome_build}/genome_fasta/genome.fa")
+    output:
+        sam = pipe(CFG["dirs"]["minimap2"] + "{seq_type}--{genome_build}/{sample_id}_out.sam")
```
Kdreval, line 96: Curious about what this does. Haven't seen this before! What is this `pipe()` wrapping?

modules/minimap2/1.0/config/default.yaml
```diff
@@ -0,0 +1,36 @@
...
+        # TODO: Update the list of available wildcards, if applicable
+        inputs:
+            # Available wildcards: {seq_type} {genome_build} {sample_id}
+            sample_fastq: "__UPDATE__"
+            reference_build: "__UPDATE__"
+
+        scratch_subdirectories: []
+
+        options:
+            minimap2: '-ax map-ont -L --MD -Y -R "@RG\tID:{sample_id}\tLB:{sample_id}\tPL:ONT\tSM:{sample_id}"'
```
Kdreval: Can you please give in-line documentation what these options do? Something similar to other documentation, like [this](https://github.com/LCR-BCCRC/lcr-modules/blob/master/modules/liftover/2.0/config/default.yaml) for example

modules/minimap2/1.0/minimap2.smk
```diff
@@ -0,0 +1,168 @@
...
+        2> {log.stderr}
+        """)
+
+
+rule _minimap2_samtools:
+    input:
+        sam = str(rules._minimap2_run.output.sam)
+    output:
+        bam = CFG["dirs"]["minimap2"] + "{seq_type}--{genome_build}/{sample_id}_out.bam",
+        complete = touch(CFG["dirs"]["minimap2"] + "{seq_type}--{genome_build}/{sample_id}_out.bam.complete")
```
Kdreval, line 125: Not sure how this works here because this output is not used in the shell command or elsewhere in the module. Maybe this is left from the earlier versions and needs to be dropped?

modules/minimap2/1.0/minimap2.smk
```diff
@@ -0,0 +1,168 @@
...
+    run:
+        op.absolute_symlink(input.bam, output.bam)
+
+
+# Symlinks the final output files into the module results directory (under '99-outputs/')
+rule _minimap2_output_bam:
+    input:
+        bam = CFG["dirs"]["sort_bam"] + "{seq_type}--{genome_build}/{sample_id}.sort.bam",
+        bai = CFG["dirs"]["sort_bam"] + "{seq_type}--{genome_build}/{sample_id}.sort.bam.bai",
+        sorted_bam = str(rules._minimap2_symlink_bam.input.bam)
```
Kdreval, line 176: usually the output of the rule is requested as input of the following rule, not the input like in this case

modules/minimap2/1.0/minimap2.smk
```diff
@@ -0,0 +1,168 @@
...
+    wildcard_constraints: 
+        sample_id = "|".join(sample_ids_minimap2)
+    run:
+        op.absolute_symlink(input.bam, output.bam)
+
+
+# Symlinks the final output files into the module results directory (under '99-outputs/')
+rule _minimap2_output_bam:
+    input:
+        bam = CFG["dirs"]["sort_bam"] + "{seq_type}--{genome_build}/{sample_id}.sort.bam",
```
Kdreval, line 174: Is this supposed to be created by the utils module? It is commented above so is not imported. Where does this come from then?

modules/minimap2/1.0/minimap2.smk
```diff
@@ -0,0 +1,168 @@
...
+        **CFG["resources"]["samtools"]
+    shell:
+        op.as_one_line("""
+        samtools view {params.opts}
+        {input.sam} > {output.bam} 
+        2> {log.stderr}
+        """)
+
+
+rule _minimap2_symlink_bam:
```
Kdreval, line 145: why not symlink directly to 99-outputs?

modules/minimap2/CHANGELOG.md
Kdreval, line 1: Can you add some basic info here to the readme similar to [other](https://github.com/LCR-BCCRC/lcr-modules/blob/master/modules/star/CHANGELOG.md) modules?

modules/utils/2.1/config/default.yaml
```diff
@@ -13,7 +13,7 @@ lcr-modules:
 ## Lines commented out with `#?` can optionally be user-configured
 ## Lines commented out with `##` act as regular comments
 
-        paired_modules: ["bwa_mem", "star"]
+        paired_modules: ["bwa_mem", "star", minimap2]
```
Kdreval: The utils is not imported here, should we need this modification? Does it work without quotation marks?

Kdreval (commented):
Thanks @hayashaalan !! Great!! I think my biggest confusion is whether (or not) we need to include the utils here - not clear if you run it with or without that module. I assumed without since it is commented. Maybe I am not right?

modules/minimap2/1.0/config/default.yaml
lkhilton: Also minimap can work with paired end data - can you think of a way you might be able to adapt this so that it can use paired or unpaired fastq files?

modules/minimap2/1.0/config/default.yaml
```diff
@@ -0,0 +1,36 @@
+lcr-modules:
+    
+    minimap2:
+
+        # TODO: Update the list of available wildcards, if applicable
+        inputs:
+            # Available wildcards: {seq_type} {genome_build} {sample_id}
+            sample_fastq: "__UPDATE__"
+            reference_build: "__UPDATE__"
```
lkhilton: Why is reference build being specified here instead of as a wildcard interpreted from the samples table?

modules/minimap2/1.0/config/default.yaml
lkhilton: This is somewhere you could use `switch_on_wildcard()` to allow the user to set different options for short read alignment.

modules/minimap2/1.0/minimap2.smk
```diff
@@ -0,0 +1,168 @@
...
+# Generates the target sentinels for each run, which generate the symlinks
+rule _minimap2_all:
+    input:
+        expand(
+            [
+                str(rules._minimap2_output_bam.output.bam),
+            ],
+            zip,  # Run expand() with zip(), not product()
+            seq_type=CFG["samples"]["seq_type"],
+            genome_build=CFG["inputs"]["reference_build"],
```
lkhilton: Can you please change this to be set based on the samples table? If you're looking to change the genome build from what's specified in the samples table, your wrapper Snakefile can just include a line like 
```
SAMPLES["genome_build"] = "hg38"
```

modules/minimap2/1.0/config/default.yaml
hayashaalan: No index files are required to run this tool

modules/minimap2/1.0/config/default.yaml
hayashaalan: I gave it more resources (10) and now it takes ~24 hours

modules/minimap2/1.0/minimap2.smk
hayashaalan: If an output is marked with pipe() then snakemake will execute the creating and consuming rule simultaneously like a group job https://snakemake.readthedocs.io/en/master/snakefiles/rules.html#piped-output

modules/minimap2/1.0/minimap2.smk
hayashaalan: Snakemake uses the touch flag to touch an empty file without having to write a shell command for it https://snakemake.readthedocs.io/en/master/snakefiles/rules.html#flag-files

modules/minimap2/1.0/minimap2.smk
hayashaalan: This is to delete the unsorted bam once the sorted bam is created by the utils module

modules/minimap2/1.0/minimap2.smk
hayashaalan: Yes it is! The utils module is imported by the wrapper snakefile instead of in the module

modules/minimap2/1.0/minimap2.smk
hayashaalan: I'm symlinking it to a new directory so that the utils module outputs the files into that directory once it has sorted and indexed the files

modules/minimap2/1.0/config/default.yaml
```diff
@@ -2,16 +2,23 @@ lcr-modules:
     
     minimap2:
 
-        # TODO: Update the list of available wildcards, if applicable
         inputs:
             # Available wildcards: {seq_type} {genome_build} {sample_id}
-            sample_fastq: "__UPDATE__"
-            reference_build: "__UPDATE__"
+            sample_fastq:
```
lkhilton, line 8: I think this needs some documentation about the `{number}` wildcard and how to correctly specify paired and unpaired fastq files.

modules/minimap2/1.0/config/default.yaml
lkhilton: Kostia's point is fair - it makes sense to set the default here MUCH higher. I think you should make 24 the default.

modules/minimap2/1.0/config/default.yaml
```diff
@@ -0,0 +1,43 @@
...
+                genome: '-ax sr -L --MD -Y -R "@RG\tID:{sample_id}\tLB:{sample_id}\tPL:Illumina\tSM:{sample_id}"'
+            samtools: '-bhS'
+
+        conda_envs:
+            minimap2: "{MODSDIR}/envs/minimap2-2.24.yaml"
+            samtools: "{MODSDIR}/envs/samtools-1.9.yaml"
+            
+        threads:
+            minimap2: 4
+            samtools: 1
```
lkhilton: Ditto for samtools - make the default higher please!

modules/minimap2/1.0/config/default.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+        resources:
+            minimap2:
```
lkhilton, line 36: Is this sufficient memory allocation for a PromethION whole genome?

modules/minimap2/1.0/envs/minimap2-2.24.yaml
```diff
@@ -0,0 +1 @@
+/projects/rmorin/projects/gambl-repos/gambl-hshaalan/src/lcr-modules/envs/minimap2/minimap2-2.24.yaml
```
lkhilton: Can you please update the symlink to point to the `lcr-modules/envs/minimap2` directory to parallel the samtools symlink below?

modules/minimap2/1.0/minimap2.smk
lkhilton: Kostia's point is that writing the `.complete` file alone isn't enough. For it to have the desired effect (forcing this rule to re-run if it's interrupted/the bam file is incomplete) it has to be an input to a subsequent rule. Can you add the `complete` file from this rule as an input to the next rule in the module?

modules/minimap2/1.0/minimap2.smk
lkhilton: To be clear - it's not the sorted bam that's being deleted but the un-sorted bam, right? 

Duplicate marking isn't necessary for PromethION, but I think it's still an important step for short-read WGS to include in this module. Can you add some rules to complete duplicate marking for short reads? You can use wildcard constraints and an input function to determine which output to symlink depending on the seq_type. (Also, this could be used for capture data in theory, right?)

lkhilton (changes requested):

Kdreval:
@hayashaalan just checking if you have any updates for this?

Kdreval:
@hayashaalan do you have any updates for this module?

PR #268, opened 2023-05-31
lymphgen 2.0 - run per sample

faranehym:
lymphgen 2.0 allows runs per sample, therefore inputting per sample maf/seg/sv and outputting per sample lymphgen calls

Example:
```
```


modules/lymphgen/1.0/config/default.yaml
```diff
@@ -3,18 +3,53 @@ lcr-modules:
...
+            #           sequenza: "results/sequenza-1.4/99-outputs/seg/{seq_type}--projection/{tumour_id}--{normal_id}--{pair_status}.{tool}.{projection}.seg"
+            #    It will first look for that sample's battenberg cnv file and then sequenza 
+            # Only uncomment the keys that have file paths that exist (always keep "empty" uncommented)
+            # Add any additional keys before "empty"
+            # If you don't have an CNV data, just include "empty" (an empty seg file), and the module will run without cnv data 
+                # battenberg: "__UPDATE__"
+                # sequenza: "__UPDATE__"
+                # controlfreec: "__UPDATE__"
+                # cnvkit: "__UPDATE__"
+                empty: "{MODSDIR}/etc/empty.seg"
```
lkhilton: I can envision people deleting the "empty" key. It might be a good idea to put an `assert` statement into your Snakefile to make sure this key is present and unmodified so you can output an interpretable error message if that happens.

modules/lymphgen/1.0/config/default.yaml
```diff
@@ -3,18 +3,53 @@ lcr-modules:
...
+                # controlfreec: "__UPDATE__"
+                # cnvkit: "__UPDATE__"
+                empty: "{MODSDIR}/etc/empty.seg" 
+            sample_sv_info: 
+            # All the SV files will be merged with a final SV file created for each sample so the order of keys don't matter 
+            # If there is no fish sv file put "{MODSDIR}/etc/empty_sv_fish.tsv" in key fish 
+            # If there is no svar_master file put "{MODSDIR}/etc/empty_svar_master.tsv" under both empty_svar and svar_master keys
+            # Always keep all three keys with real file paths 
+                fish: "__UPDATE__"
+                empty_svar: "{MODSDIR}/etc/empty_svar_master.tsv"
```
lkhilton: Same as above - probably a good idea to ensure this key/value is always included.

modules/lymphgen/1.0/config/default.yaml
```diff
@@ -3,18 +3,53 @@ lcr-modules:
...
+            # If there is no fish sv file put "{MODSDIR}/etc/empty_sv_fish.tsv" in key fish 
+            # If there is no svar_master file put "{MODSDIR}/etc/empty_svar_master.tsv" under both empty_svar and svar_master keys
+            # Always keep all three keys with real file paths 
+                fish: "__UPDATE__"
+                empty_svar: "{MODSDIR}/etc/empty_svar_master.tsv" 
+                svar_master: "__UPDATE__" 
             # Path to the directory where LGenIC should be installed
             lgenic_exec: "__UPDATE__"
+            gamblr_branch: "" # Leave blank unless you need to install from a specific branch from the GAMBLR repo (expert users only)
+            gamblr_config_url: "https://raw.githubusercontent.com/morinlab/GAMBLR/master/config.yml"
```
lkhilton: Did any of your testing include installing GAMBLR, or did your tests all use an existing GAMBLR repo?

modules/lymphgen/1.0/config/default.yaml
```diff
@@ -3,18 +3,53 @@ lcr-modules:
...
+            genome:
+                run_paired_tumours: True
+                run_unpaired_tumours_with: "unmatched_normal"
+                run_paired_tumours_as_unpaired: False
+            capture:
+                run_paired_tumours: True
+                run_unpaired_tumours_with: "unmatched_normal"
+                run_paired_tumours_as_unpaired: False
+        
+        unmatched_normal_ids:
```
lkhilton: You can remove the `unmatched_normal_ids` specifications here, that usually goes into a "general.yaml" config and isn't specified per-module.

modules/lymphgen/1.0/config/default.yaml
```diff
@@ -3,18 +3,53 @@ lcr-modules:
...
+            capture--grch37: "__UPDATE__"
+            # Add additional {seq_type}--{genome_build}
 
         options:
-            outprefix: "my_cohort"
             reformat_seg:  #  Default columns for Battenberg output
                 Tumor_Sample_Barcode: "ID"
-                chromosome: "chrom"
+                chromosome: "chr"
```
lkhilton: Is this how the CNV modules encode the header for the seg files? It's `ID chr start end` not `ID chrom start end`?

modules/lymphgen/1.0/config/default.yaml
```diff
@@ -29,11 +64,16 @@ lcr-modules:
                 bcl6truevalues: "__UPDATE__"  # Can use a list
                 bcl6falsevalues: "__UPDATE__"
             lymphgen_input:
-                seq_type: "genome"  # Availible options: exome/genome (they do the same thing) and targeted (for panels)
+                seq_type: "__UPDATE__"  # Availible options: exome/genome (they do the same thing) and targeted (for panels)
                 use_log_ratio: "__UPDATE__"
```
lkhilton: Maybe it makes sense to set the `use_log_ratio` value to `TRUE` as the default, since the CNV modules all output seg files with log ratio values. With this new set up I'm guessing it would be rare for this to be false.

modules/lymphgen/1.0/config/default.yaml
```diff
@@ -29,11 +64,16 @@ lcr-modules:
...
             lymphgen_run:
                 lymphgen_path: "/projects/rmorin/software/lcr-scripts/LymphGen/Run_Lymphgen.R"
```
lkhilton: Could you update this to the new path that the LymphGen module will be tracked at?

modules/lymphgen/1.0/lymphgen.smk
```diff
@@ -28,6 +31,46 @@ CFG = op.setup_module(
     subdirectories = ["inputs", "reformat_seg", "lymphgen_input", "add_svs", "lymphgen_run", "composite_other", "outputs"],
 )
 
+def _find_best_seg(wildcards):
+    this_tumor = op.filter_samples(RUNS, genome_build = wildcards.genome_build, tumour_sample_id = wildcards.tumour_id, normal_sample_id = wildcards.normal_id, pair_status = wildcards.pair_status, seq_type = wildcards.seq_type)
+    return this_tumor.cnv_path
+
+def _find_best_svar_master(wildcards):
+    this_tumor = op.filter_samples(RUNS, genome_build = wildcards.genome_build, tumour_sample_id = wildcards.tumour_id, normal_sample_id = wildcards.normal_id, pair_status = wildcards.pair_status, seq_type = wildcards.seq_type)
+    return this_tumor.svar_master_path
```
lkhilton: How does svar_master_path get added to the RUNS table?

ckrushton:
Will look over and test tomorrow morning PDT

modules/lymphgen/1.0/lymphgen.smk
lkhilton: Maybe this function isn't used anywhere?

modules/lymphgen/1.0/lymphgen.smk
```diff
@@ -409,10 +550,6 @@ rule _lymphgen_output_txt:
 
 # Set the applicable wildcards, based on the provided input files
 # Are we running LymphGen with CNV/SV data?
-if "sample_seg" in CFG["inputs"] and CFG["inputs"]["sample_seg"] != "" and CFG["inputs"]["sample_seg"] != "None":
-    cnvs_wc = ["with_cnvs", "no_cnvs"]
-else:
-    cnvs_wc = ["no_cnvs"]
 
 if "sample_sv_info" in CFG["inputs"] and CFG["inputs"]["sample_sv_info"] != "" and CFG["inputs"]["sample_sv_info"] != "None":
```
lkhilton: Is there more to do here to make sure this works on a per-sample basis? The way you have it set up, the `sample_sv_info` field in the config will never be "None", so this might need more refinement. Ditto for A53, which should maybe be set at the same time as the CNV info is set (e.g. the RUNS_NO_CNV table should only be expanded on `no_cnv` and `no_A53` and RUNS_CNV should be expanded on `["with_cnv", "no_cnv"]` and `["with_A53", "no_A53"]`?

modules/lymphgen/1.0/src/curate_svs_for_lymphgen.R
```diff
@@ -0,0 +1,60 @@
...
+  filter(str_detect(fusion, "BCL6")) %>%
+  filter(str_detect(fusion, paste(config_bcl6, collapse = "|"))) %>%
+  pull(fusion)
+
+real_bcl2 <- svar_master_annotated %>%
+  filter(str_detect(fusion, "BCL2")) %>%
+  filter(str_detect(fusion, paste(config_bcl2, collapse = "|"))) %>%
+  pull(fusion)
+
+if (nrow(one_fish) == 0) {
```
lkhilton: I don't think it's right to output "NEG" values for missing data. Let's discuss how to handle this scenario.

lkhilton (changes requested):
This looks excellent! A few key requests to fix up a couple of logical issues, we can discuss.

ckrushton:
Just as a general comment; does the LymphGen R code work correctly for a single sample? IIRC, the web version of lymphgen had a bug where it was impossible to run a single sample through...

faranehym:
@ckrushton I have run it with just one sample so the LymphGen R code works fine for that - not sure why it'd be erroring out in the web version

modules/lymphgen/1.0/config/default.yaml
```diff
@@ -65,9 +62,10 @@ lcr-modules:
                 bcl6falsevalues: "__UPDATE__"
             lymphgen_input:
                 seq_type: "__UPDATE__"  # Availible options: exome/genome (they do the same thing) and targeted (for panels)
-                use_log_ratio: "__UPDATE__"
+                use_log_ratio: "True"
             lymphgen_run:
-                lymphgen_path: "/projects/rmorin/software/lcr-scripts/LymphGen/Run_Lymphgen.R"
+                lymphgen_path: "/home/fymoayyed/LymphGen/Run_Lymphgen.R"
```
lkhilton: I think it's actually better to make this 
```
lymphgen_path: "__UPDATE__"
```

We have had some external users make use of this module to prepare their files for the LymphGen webtool, so ideally the default is that users have to specify where they've saved the source code if they even have access to it.

modules/lymphgen/1.0/lymphgen.smk
```diff
@@ -31,15 +31,40 @@ CFG = op.setup_module(
     subdirectories = ["inputs", "reformat_seg", "lymphgen_input", "add_svs", "lymphgen_run", "composite_other", "outputs"],
 )
 
+assert ("empty" in list(CFG["inputs"]["sample_seg"].keys())), (
```
lkhilton: Is there a better way to test this that also ensures that the value is unmodified, not just present?

modules/lymphgen/1.0/lymphgen.smk
```diff
@@ -31,15 +31,40 @@ CFG = op.setup_module(
...
     this_tumor = op.filter_samples(RUNS, genome_build = wildcards.genome_build, tumour_sample_id = wildcards.tumour_id, normal_sample_id = wildcards.normal_id, pair_status = wildcards.pair_status, seq_type = wildcards.seq_type)
     return this_tumor.cnv_path
 
-def _find_best_svar_master(wildcards):
+def _find_best_sv(wildcards):
     this_tumor = op.filter_samples(RUNS, genome_build = wildcards.genome_build, tumour_sample_id = wildcards.tumour_id, normal_sample_id = wildcards.normal_id, pair_status = wildcards.pair_status, seq_type = wildcards.seq_type)
-    return this_tumor.svar_master_path
-
-def best_seg(mod_config):
+    print(this_tumor.has_sv)
```
lkhilton: Could you please remove any print statements you used for debugging throughout?

modules/lymphgen/1.0/lymphgen.smk
```diff
@@ -64,11 +89,46 @@ def best_seg(mod_config):
...
+
+        df[index]["sv_path"] = possible_outputs[0]
+
+        mod_config["runs"].at[index, "sv_path"] = df[index]["sv_path"]
+
     all = mod_config["runs"]
+
+    all['has_sv'] = False
+    for i, row in all.iterrows():
+        if (row['tumour_seq_type'] == "capture") & (row['sv_path'] != config["lcr-modules"]["lymphgen"]["inputs"]["sample_sv_info"]["other"]["empty_sv"]):
```
lkhilton: Just confirming the logic here because I'm not sure I understand this: 

- If a sample is capture, it should only go through the `with_sv` pipeline if it has FISH
- If a sample is genome, it should go through `with_sv` whether it has FISH OR svar_master OR manta outputs

Is this how it's working?

lkhilton: Could you add a few comments to keep track of that logic so future users/editors know what's going on and why?

modules/lymphgen/1.0/lymphgen.smk
```diff
@@ -291,38 +351,40 @@ rule _lymphgen_install_GAMBLR:
...
 #print(CFG["inputs"]["sample_sv_info"]["svar_master"])
 
-def get_svar_master(wildcards):
-    possible_output = (expand(config["lcr-modules"]["lymphgen"]["inputs"]["sample_sv_info"]["svar_master"], zip,
-                                        tumour_id=wildcards.tumour_id,
-                                        normal_id=wildcards.normal_id,
-                                        pair_status=wildcards.pair_status,
-                                        genome_build=wildcards.genome_build,
-                                        seq_type=wildcards.seq_type))
+# def get_svar_master(wildcards):
```
lkhilton: Please delete commented out code!

modules/lymphgen/1.0/src/curate_svs_for_lymphgen.R
```diff
@@ -1,32 +1,51 @@
...
+#   3) fish w/o sample_id (like empty fish) + svar master -- ~has_fish & has_sv
+#   4) fish w/o sample_id (like empty fish) + manta -- ~has_fish & has_sv
+#   5) fish w/ sample_id + empty sv -- has_fish & ~has_sv
+
 all_fish <- snakemake@input[["fish"]]
-one_svar_master <- snakemake@input[["svar_master"]]
+sv <- snakemake@input[["sv"]]
+
+if (str_detect(sv, "manta")) { # manta
+  sv <- fread(sv, skip = 130) %>%
```
lkhilton: A more reliable way to do this (in case manta ever produces a different number of header lines) is to use `read_tsv(..., comment = "#", col_names = c("CHROM_A", "START_A", ...))`. You have to specify the col_names because the column header line in the bedpe has a "#". (Unless they've updated the behaviour of read_tsv to accept `##` as a comment character, in which case you would just need to switch fread to read_tsv. Test it out.)

lkhilton (changes requested):
Looking great, thanks for all your hard work on this! Just a few minor edits/clarifications requested.

modules/lymphgen/1.0/config/default.yaml
lkhilton: Just realized I contradicted my earlier comment about this, hah. I think it's better to leave this as `__UPDATE__` still!

modules/lymphgen/1.0/src/curate_svs_for_lymphgen.R
faranehym: I kept it as fread with `skip = "CHROM"` cause I think it's quicker than read_tsv but if read_tsv is still better in this situation then I can change it :)

lkhilton (approved):

lkhilton (approved):

PR #272, opened 2023-07-10
Module/fishhook/1.0

Jacky-Yiu:
# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable

Example:
```
```


modules/fishhook/1.0/config/default.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+        sample_set: ["__UPDATE__"]
+        seq_types: ["genome", "capture"]
+        include_non_coding: True
+        prepare_mafs: "{SCRIPTSDIR}/generate_smg_inputs/1.0/generate_smg_inputs.R"
+        src_dir: "{MODSDIR}/src/"
+
+        options:
+            tiles_size: 1000 #This is the size of tiles you want to split the genome into
+            coveriates: "_UPDATE_" #please supply a coveriate file in bed format, leave it blank if you want to run FishHook without coveriates
```
Kdreval: We should leave this blank if the mode without covariates is preferred (or will be more often used), but thank you for documentation!

modules/fishhook/1.0/config/default.yaml
```diff
@@ -0,0 +1,38 @@
...
+            sample_sets: "__UPDATE__"
+
+        sample_set: ["__UPDATE__"]
+        seq_types: ["genome", "capture"]
+        include_non_coding: True
+        prepare_mafs: "{SCRIPTSDIR}/generate_smg_inputs/1.0/generate_smg_inputs.R"
+        src_dir: "{MODSDIR}/src/"
+
+        options:
+            tiles_size: 1000 #This is the size of tiles you want to split the genome into
```
Kdreval: This is assuming the tiling will also be used - but there is also a way to run it without tiles, right? Just wondering if the documentation here should be something like the one below that says Leave blank ...

modules/fishhook/1.0/scr/R/run_fishhook.R
```diff
@@ -0,0 +1,64 @@
...
+    library(rtracklayer)
+  })
+)
+#load maf and convert to GRange
+maf = dt2gr(fread(snakemake@input[[2]]))
+
+#split maf to numerous tiles
+tiles = gr.tile(seqinfo(maf), snakemake@params[[1]])
+
+events = maf %Q% (Variant_Classification != 'Silent')
```
Kdreval: So the silent mutations are explicitly excluded here? Can we make it optional and have in configurable?

modules/fishhook/1.0/scr/R/run_fishhook.R
```diff
@@ -0,0 +1,64 @@
...
+
+if(file.exists(snakemake@params[[2]])){
+  message("Running FishHook with Coveriate...")
+  fish_tiles = Fish(hypotheses = tiles, 
+                    events = events, 
+                    eligible = eligible, 
+                    idcol = 'Tumor_Sample_Barcode',
+                    use_local_mut_density=TRUE)
+  
+  coveriate_data = gr.sub(import(snakemake@params[[2]]), 'chr', "") ## import from bed then gUtils::gr.sub to strip 'chr' identifier
```
Kdreval: Does this applies to any genome build? Or FishHook does not considers genome build at all? Just wonder if the `chr` should always be stripped or is dependent on the grch37/hg38

modules/fishhook/CHANGELOG.md
```diff
@@ -0,0 +1,16 @@
...
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [1.0] - 2023-06-27
+
+This release was authored by Jacky Yiu.
+
+<!-- TODO: Explain each important module design decision below. -->
+
+- Both grch37-based and grach38-based  maf files are supported, please supply coveriate file of the same type
```
Kdreval: grch38

modules/fishhook/CHANGELOG.md
```diff
@@ -0,0 +1,16 @@
...
+All notable changes to the `fishhook` module will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [1.0] - 2023-06-27
+
+This release was authored by Jacky Yiu.
+
+<!-- TODO: Explain each important module design decision below. -->
```
Kdreval: remove this TODO as it is done

modules/fishhook/1.0/scr/R/run_fishhook.R
Kdreval: I saw later in the CANGELOG that both grch37 and grch38 are supported - is it also important whether the maf `Chromosome` should be without chr prefix?

Jacky-Yiu: FishHook does not consider genome build at all since it only use the maf file. however the covariate file does need to be the same type. But you are right, I will apply gr.sub to both maf and covariate file

modules/fishhook/CHANGELOG.md
Jacky-Yiu: done

modules/fishhook/CHANGELOG.md
Jacky-Yiu: done

modules/fishhook/1.0/scr/R/run_fishhook.R
Jacky-Yiu: Done

modules/fishhook/1.0/config/default.yaml
Jacky-Yiu: Done

modules/fishhook/1.0/config/default.yaml
Jacky-Yiu: Done, using tile will be the default mode, is tile_size is blank, user should include a gene list for gene-centric analysis.

modules/fishhook/1.0/scr/R/run_fishhook.R
Jacky-Yiu: include_silent_mutation is added as an option

Kdreval (approved):

PR #274, opened 2023-07-18
Module/gistic2/1.0

gillissierra:
# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable

Example:
```
```


modules/gistic2/1.0/config/default.yaml
```diff
@@ -0,0 +1,43 @@
+lcr-modules:
+    gistic2:
+        inputs:
+            # The seg files should be "filled", meaning missing segments should be replaced with a dummy segment with log2(CN)=0 and no LOH.
+            # The cnv_master module outputs seg files in this filled format
+            # Available wildcards: {seq_type} {projection} {case_set}
+            seg: "__UPDATE__"
+            all_sample_sets: "__UPDATE__" 
+        case_set: "__UPDATE__"
```
Kdreval: I just want to make sure this accepts multiple strings as case set. I think they need to be passed as a list, like `["__UPDATE__"]`. This will allow to have one or more case sets to run at single snakefile setup. For example, if the module hasn't been run before, you can imagine there are few case sets of interest like `["all", "DLBCL-only", "BL-only", "BL-DLBCL", "FL-only"]` for example. So I think it has to be a list here to be able to run all at a single snakefile launch

modules/gistic2/1.0/config/default.yaml
```diff
@@ -0,0 +1,43 @@
...
+        prepare_seg: "{SCRIPTSDIR}/generate_smr_inputs/1.0/generate_smr_inputs.R"
+        projections: ["grch37", "hg38"]
+
+        scratch_subdirectories: []
+
+        options:
+            conf_level: ["90"] # Confidence level used to calculate the region containing a driver.
+            gistic2_run: "-gcm extreme -maxseg 5000 -brlen 0.5 -js 1 -qvt 0.5 -td 0.05"
```
Kdreval, line 17: Thanks for documentation, this is great!

modules/gistic2/1.0/envs/gistic2-2.0.23.yaml
```diff
@@ -0,0 +1 @@
+/projects/rmorin_scratch/sgillis_temp/lcr-modules/envs/gistic2/gistic2-2.0.23.yaml
```
Kdreval: This looks like a symlink to scratch space - which will break if someone tries to run it outside of the GSC. Instead, can you please point the symlink tot recipe that is now in `envs/gistic2/gistic2-2.0.23.yaml`?

modules/gistic2/1.0/envs/samtools-1.9.yaml
```diff
@@ -0,0 +1 @@
+../../../../envs/samtools/samtools-1.9.yaml
```
Kdreval: This symlink is an example created by cookiecutter, and can be deleted in this module :smile:

modules/gistic2/1.0/gistic2.smk
```diff
@@ -0,0 +1,304 @@
...
+        op.absolute_symlink(input.all_sample_sets, output.all_sample_sets)
+
+# Download refgene reference MAT file
+rule _gistic2_download_ref:
+    output:
+        refgene_mat = CFG["dirs"]["inputs"] + "references/{projection}.refgene.mat"
+    group: 
+        "input_and_format"
+    params:
+        url = "http://bcgsc.ca/downloads/morinlab/gistic2_references/{projection}.refgene.mat",
```
Kdreval, line 86: Thanks for adding these to the downloads page!

modules/gistic2/1.0/gistic2.smk
```diff
@@ -0,0 +1,304 @@
...
+    input:
+        seg = str(rules._gistic2_prepare_seg.output.seg)
+    output:
+        temp_markers = temp(CFG["dirs"]["markers"] + "temp_markers--{case_set}--{projection}.txt"),
+        markers = CFG["dirs"]["markers"] + "markers--{case_set}--{projection}.txt"
+    log:
+        stderr = CFG["logs"]["markers"] + "{case_set}--{projection}--markers.stderr.log"
+    shell: 
+        op.as_one_line("""
+        sed '1d' {input.seg} | cut -f2,3 > {output.temp_markers}
```
Kdreval, line 145: I don't remember if this has to deal with the header of the seg file or assumes there is no header - I trust this works as expected because I know you run this successfully already!

modules/gistic2/1.0/gistic2.smk
```diff
@@ -0,0 +1,304 @@
...
+        2> {log.stderr}
+        """)
+
+# Run gistic2 for a single seq_type (capture, genome) for the confidence thresholds listed in the config
+rule _gistic2_run:
+    input:
+        seg = str(rules._gistic2_prepare_seg.output.seg),
+        refgene_mat = str(rules._gistic2_download_ref.output.refgene_mat),
+        markers = str(rules._gistic2_make_markers.output.markers)
+    output:
```
Kdreval, line 159: I think we can consider simplifying this by creating a `complete` file with touch at the end of the rule - or, alternatively, only listing few meaningful outputs as outputs of this rule. This will also simplify the next rule with symlinking and 99-outputs structure.

Kdreval (commented):
Thank you so much for this important update!!

modules/gistic2/1.0/config/default.yaml
gillissierra: Right! I didn't test this. I will have to make sure that's compatible with the lcr-script

modules/gistic2/1.0/gistic2.smk
gillissierra: Yes, the `'1d'` tells sed to skip the first line

modules/gistic2/1.0/gistic2.smk
gillissierra: I agree. I like the idea of listing only a few outputs and symlinking them. Are there other meaningful outputs besides:

- scores.gistic
- all_data_by_genes.txt
- all_lesions.conf_90.txt

Kdreval: Thanks!!
Yes I think it will be good to also add

- amp_genes.conf_90.txt
- del_genes.conf_90.txt

These are what is used for maftools processing if someone decides to go ahead with that way of downstream analysis. Although there are problems with it, I can imagine there will be users who would prefer to do that and at least it will be convenient to have all in one place.

gillissierra:
Comments above have been incorporated

modules/gistic2/1.0/config/default.yaml
Kdreval: Thanks for the updates!! Can you please also add here a comment with an example of how to specify multiple sample sets?

Kdreval (approved):
Thank you so much for this update! Feel free to merge - or push any changes you may still have. I think it looks great!

PR #278, opened 2023-10-26
Module/gistic2/1.0 UPDATE

gillissierra:
# Pull Request Checklists

Example:
```
```

NOTE: a PR for lcr-scripts/generate_smr_inputs/1.0/generate_smr_inputs.R will also need to be approve and merged for this update to work.


modules/gistic2/1.0/config/default.yaml
```diff
@@ -5,10 +5,12 @@ lcr-modules:
             # The cnv_master module outputs seg files in this filled format
             # Available wildcards: {seq_type} {projection} {case_set}
             seg: "__UPDATE__"
-            all_sample_sets: "__UPDATE__" 
+            subsetting_categories: "__UPDATE__" # e.g. "data/metadata/level3_subsetting_categories.tsv"
         case_set: [__UPDATE__] # E.g. ["DLBCL-only", "BL_DLBCL-BL-like_all"]
-        prepare_seg: "{SCRIPTSDIR}/generate_smr_inputs/1.0/generate_smr_inputs.R"
+        prepare_seg: "/projects/rmorin_scratch/sgillis_temp/lcr-scripts/generate_smr_inputs/1.0/generate_smr_inputs.R"
```
Kdreval: This probably should be converted back to the `{SCRIPTSDIR}` wildcard?

Kdreval (commented):
Thanks! Just a small update to the path specified in config but otherwise looks great!!

modules/gistic2/1.0/config/default.yaml
gillissierra: ahh yes, you're right. I had issues when trying to use `{SCRIPTSDIR}' when my `src/lcr-scripts` pointed to my own lcr-scripts repo. I assum it will work once my symlink is changed back to the official lcr-scripts dir

Kdreval (dismissed):

Kdreval:
Thank you 😊

Kdreval (approved):

PR #280, opened 2023-11-01
Module/specshla/1.0

Jwong684:
# Pull Request Checklists

Created an lcr-module for spechla, which performs hla-typing at a rapid speed for mRNA, WES, and WGS (and long reads).

https://pubmed.ncbi.nlm.nih.gov/37714157/

Does not seem particularly useful for FFPE tissue.

### Required

  ```bash
  ```

### If applicable

Example:
```
```


modules/spechla/1.0/envs/spechla_env.yaml
Kdreval, line 1: Can you please put this recipe into the lc-modules/envs/spechla and then symlink here, so it is set up the same way other recipes are set up too?

modules/spechla/1.0/envs/samtools-1.9.yaml
Kdreval, line 1: If samtools are not used this can be removed

modules/spechla/1.0/config/default.yaml
```diff
@@ -0,0 +1,58 @@
...
+                # -z False: Whether only mask exon region, True or False, default is False.
+                # -p The population of the sample [Asian, Black, Caucasian, Unknown, nonuse] for annotation. Default is Unknown, meaning use mean allele frequency in all populations. nonuse indicates only adopting mapping score and considering zero-frequency alleles.
+            
+            purity_ploidy_file: "" 
+                # path to a text file containing the purity and ploidy information of each sample
+                # file format is: sample_id\tploidy\tpurity
+                # if information is not there, then assumed ploidy = 2, purity = 1
+
+        conda_envs:
+            samtools: "{MODSDIR}/envs/samtools-1.9.yaml"
```
Kdreval, line 31: If samtools are not used this can be removed

modules/spechla/1.0/config/default.yaml
```diff
@@ -0,0 +1,58 @@
...
+        # TODO: Update the list of available wildcards, if applicable
+        inputs:
+            # Available wildcards: {seq_type} {genome_build} {sample_id}
+            sample_bam: "__UPDATE__"
+            sample_bai: "__UPDATE__"
+
+        scratch_subdirectories: []
+
+        options:
+            spechla:
```
Kdreval, line 14: Amazing, thanks for all this documentation!!

modules/spechla/1.0/spechla.smk
```diff
@@ -0,0 +1,310 @@
...
+# Symlinks the input files into the module results directory (under '00-inputs/')
+rule _spechla_input_bam:
+    input:
+        bam = CFG["inputs"]["sample_bam"],
+        bai = CFG["inputs"]["sample_bai"]
+    output:
+        bam = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{sample_id}.bam",
+        bai = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{sample_id}.bai",
+        crai = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{sample_id}.crai"
+    group:
```
Kdreval: Looks like this is only rule in this group, should it be combined with other rules or dropped out of a group?

modules/spechla/1.0/spechla.smk
```diff
@@ -0,0 +1,310 @@
...
+rule _spechla_setup:
+    output:
+        txt = touch(CFG["dirs"]["inputs"] + "SpecHLA/clone.done")
+    log:
+        stdout = CFG["logs"]["inputs"] + "git_clone.stdout.log"
+    params:
+        scriptdir = CFG["dirs"]["inputs"] + "SpecHLA"
+    shell:
+        op.as_one_line("""
+        git clone git@github.com:deepomicslab/SpecHLA.git --depth 1  {params.scriptdir} > {log.stdout} 2>&1 &&
```
Kdreval: Does it assume git is installed and available in the env? Should this be run in conda env?

modules/spechla/1.0/spechla.smk
```diff
@@ -0,0 +1,310 @@
...
+    run:
+        op.absolute_symlink(input.bam, output.bam)
+        op.absolute_symlink(input.bai, output.bai)
+        op.absolute_symlink(input.bai, output.crai)
+
+
+# Example variant calling rule (multi-threaded; must be run on compute server/cluster)
+rule _spechla_setup:
+    output:
+        txt = touch(CFG["dirs"]["inputs"] + "SpecHLA/clone.done")
```
Kdreval: Is this automatically created after the command is run? Don't see this in the shell command so wondering whether `&& touch {output.txt}` is needed

modules/spechla/1.0/spechla.smk
```diff
@@ -0,0 +1,310 @@
...
+        scriptdir = CFG["dirs"]["inputs"] + "SpecHLA"
+    shell:
+        op.as_one_line("""
+        git clone git@github.com:deepomicslab/SpecHLA.git --depth 1  {params.scriptdir} > {log.stdout} 2>&1 && 
+        chmod +x -R {params.scriptdir}/bin/*
+        """)
+
+rule _spechla_index:
+    input:
+        txt = CFG["dirs"]["inputs"] + "SpecHLA/clone.done"
```
Kdreval: Can you please change this to `str(rules._spechla_setup.output.txt)`?

modules/spechla/1.0/spechla.smk
```diff
@@ -0,0 +1,310 @@
...
+# Extract HLA reads from bam and cram
+def _which_genome(wildcards):
+    if "38" in str({wildcards.genome_build}):
+        return "hg38"
+    else:
+        return "hg19"
+
+rule _spechla_extract_reads:
+    input:
+        txt = CFG["dirs"]["inputs"] + "SpecHLA/index.done",
```
Kdreval: Here (and in the rules below)  the inputs also need a switch to the str(rule..output....) so it is consistent with the rest of the modules

modules/spechla/1.0/spechla.smk
```diff
@@ -0,0 +1,310 @@
...
+        outdir = CFG["dirs"]["hla_reads"] + "{seq_type}--{genome_build}/"
+    conda: CFG["conda_envs"]["spechla"]
+    group: "extract_and_run"
+    threads:
+        CFG["threads"]["extract_reads"]
+    resources:
+        **CFG["resources"]["extract_reads"]  
+    shell:
+        op.as_one_line("""
+        bash {params.scriptdir}/script/ExtractHLAread.sh -s {params.sample_id} -b {input.bam} -r {params.genome} -o {params.outdir} > {log.stderr} 2>&1
```
Kdreval: Does not really matter but if you change here the `{params.sample_id}` to `{wildcards.sample_id}` there will be no need to define a parameter for this :smile:

modules/spechla/1.0/spechla.smk
```diff
@@ -0,0 +1,310 @@
...
+def _which_seqtype(wildcards):
+    if any(seq_type in str({wildcards.seq_type}) for seq_type in ['exome', 'capture', 'target', 'mrna']):
+        return 1
+    else:
+        return 0
+
+rule _spechla_hla_typing:
+    input:
+        fq_1 = CFG["dirs"]["hla_reads"] + "{seq_type}--{genome_build}/{sample_id}_extract_1.fq.gz",
+        fq_2 = CFG["dirs"]["hla_reads"] + "{seq_type}--{genome_build}/{sample_id}_extract_2.fq.gz",fastq_unpaired = CFG["dirs"]["hla_reads"] + "{seq_type}--{genome_build}/{sample_id}_extract.unpaired.fq.gz"
```
Kdreval: This line seems to be missing a return (I mean this should be 2 lines instead of 1)

modules/spechla/1.0/spechla.smk
```diff
@@ -0,0 +1,310 @@
...
+        -v {params.long_indels} {params.opts} > {log.stderr} 2>&1
+        """)
+        
+# Remove unwanted outputs
+rule _spechla_cleanup:
+    input:
+        realigned_bam = CFG["dirs"]["spechla"] + "{seq_type}--{genome_build}/{sample_id}/{sample_id}.realign.sort.bam",
+        realigned_bam_depth = CFG["dirs"]["spechla"] + "{seq_type}--{genome_build}/{sample_id}/{sample_id}.realign.sort.bam.depth"
+    output:
+        done = touch(CFG["dirs"]["spechla"] + "{seq_type}--{genome_build}/{sample_id}/.removed.bloat")
```
Kdreval, line 193: Is this rule missing a run/shell directive? Not clear what it does having just input/output

modules/spechla/1.0/config/default.yaml
```diff
@@ -0,0 +1,58 @@
...
+                var_qual: 0.01 #  Minimum variant quality. Default is 0.01. Set it larger in high quality samples.
+                var_depth: 5 # Minimum variant depth
+                min_depth_masking: 5 # The mean depth in a window lower than this value will be masked by N, default is 5. Set 0 to avoid masking.
+                long_indels: False
+                opts: "" # Other options to manually enter:
+                # -r 0.1: The minimum Minor Allele Frequency (MAF), default is 0.05 for full length and 0.1 for exon typing.
+                # -z False: Whether only mask exon region, True or False, default is False.
+                # -p The population of the sample [Asian, Black, Caucasian, Unknown, nonuse] for annotation. Default is Unknown, meaning use mean allele frequency in all populations. nonuse indicates only adopting mapping score and considering zero-frequency alleles.
+            
+            purity_ploidy_file: ""
```
lkhilton, line 25: Maybe a future update to this could include a script to generate this file automagically from the output of other modules like Battenberg and Sequenza?

modules/spechla/1.0/envs/spechla_env.yaml
```diff
@@ -0,0 +1,115 @@
+name: /home/wangshuai/softwares/SpecHLA/spechla_env
```
lkhilton: The proper home for this is in `lcr-modules/envs` with a symlink to this file as for the `samtools-1.9.yaml` file. 

Could you please also regenerate this with the `--no-builds` option? The builds frequently make the environment impossible to reproduce.

Kdreval (commented):
Thanks Jasper! I left some comments!

modules/spechla/1.0/spechla.smk
```diff
@@ -0,0 +1,310 @@
...
+# Example variant calling rule (multi-threaded; must be run on compute server/cluster)
+rule _spechla_setup:
+    output:
+        txt = touch(CFG["dirs"]["inputs"] + "SpecHLA/clone.done")
+    log:
+        stdout = CFG["logs"]["inputs"] + "git_clone.stdout.log"
+    params:
+        scriptdir = CFG["dirs"]["inputs"] + "SpecHLA"
+    shell:
+        op.as_one_line("""
```
lkhilton, line 79: I see that there are multiple releases on the GitHub repo, can you use the config to tie the install to the current version with the option to update it if there are future releases? https://github.com/deepomicslab/SpecHLA/releases

modules/spechla/1.0/spechla.smk
lkhilton: There's a newline missing between fq_2 and fastq_unpaired.

modules/spechla/1.0/spechla.smk
```diff
@@ -0,0 +1,310 @@
...
+        -u {params.which_seqtype} 
+        -j {threads} 
+        -q {params.var_quality} 
+        -s {params.var_depth} 
+        -k {params.min_depth_masking} 
+        -v {params.long_indels} {params.opts} > {log.stderr} 2>&1
+        """)
+        
+# Remove unwanted outputs
+rule _spechla_cleanup:
```
lkhilton, line 188: I'm inferring that this is here to ensure the temp files get deleted - do they not get deleted without this rule? Does including the bam files as inputs to the next rule `_spechla_generage_freqlist` rule not do the same thing?

modules/spechla/1.0/spechla.smk
lkhilton: I think this is in a group by itself - if you group it with the next rule that runs per-sample, it will run more efficiently as part of a larger group of rules.

modules/spechla/1.0/spechla.smk
```diff
@@ -0,0 +1,310 @@
...
+        -purity {params.purity}
+        -ploidy {params.ploidy}
+        -S {params.sample_id} 
+        -F {input.freq_list}
+        -T {input.hla_results}
+        -O {params.outdir} > {log.stderr} 2>&1
+        """)
+
+# Symlinks the final output files into the module results directory (under '99-outputs/')
+rule _spechla_output_txt:
```
lkhilton, line 278: Similar to how you've grouped the input symlinks, you could group the output symlinks with the previous rule.

lkhilton (changes requested):
Nice work, Jasper. Just a few issues that need to be resolved.

modules/spechla/1.0/spechla.smk
Jwong684: Do you have an example of how to configure this? I'm not too sure how to do a version-control for git pulls here

Kdreval: This might be an example
https://github.com/LCR-BCCRC/lcr-modules/blob/3de42f1320ffbe97c2024d93550d2e7433165620/modules/rainstorm/1.0/rainstorm.smk#L57-L69

modules/spechla/1.0/spechla.smk
Jwong684: It just touches an empty file after removing the input (set as temp outputs from a previous rule)

modules/spechla/1.0/spechla.smk
Jwong684: I just included it here as a separate rule so I could generate all the intermediate files if necessary. If I use the input of this in the next rule, it should also do the same thing I believe. I'll just keep it as a separate rule just to keep it clean

modules/spechla/1.0/spechla.smk
Jwong684: Modified to download tar.gz

modules/spechla/1.0/spechla.smk
Jwong684: It is automatically created

modules/spechla/1.0/spechla.smk
Jwong684: Changed all to wildcards.sample_id

modules/spechla/1.0/config/default.yaml
Jwong684: Would be nice to have a cnvmaster version of purity_ploidy

modules/spechla/1.0/envs/spechla_env.yaml
Jwong684: Done

modules/spechla/1.0/spechla.smk
Jwong684: Changed to version-based approach of specHLA set-up

modules/spechla/1.0/spechla.smk
Jwong684: Removed this in the set-up group, but grouped in the loh + 99-symlinks rules

lkhilton (approved):

PR #284, opened 2023-11-17
ichorCNA 1.1 update patch

Jwong684:
Updated ichorcna to allow for more flexibility in reference wigs (though ended up being a moot change for grch37-noalt); updated reference workflow to include hg19-reddy_masked.

Note that chrY info is needed for ichorCNA to complete. In the case of grch37-noalt (with no chrY), added extra lines in the wig to denote 0 coverage for chrY.

# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable

Example:
```
```

### DIFF
--- a/modules/ichorcna/1.1/config/default.yaml
+++ b/modules/ichorcna/1.1/config/default.yaml
@@ -24,19 +24,40 @@ lcr-modules:
                 ichorCNA_normalPanel: 
                     "1000000": "inst/extdata/HD_ULP_PoN_{genome_build}_1Mb_median_normAutosome_median.rds"
                     "500000": "inst/extdata/HD_ULP_PoN_{genome_build}_500kb_median_normAutosome_median.rds"
+                # relative to ichorCNA directory (in 00-inputs/ichorCNA/)
+                ichorCNA_normalPanel_custom:
+                    grch37: ""
+                    hg19: ""
+                    grch38: ""
+                    hg38: ""
+                    hs37d5: ""
                 # must use gc wig file corresponding to same binSize (required)
                 ichorCNA_gcWig:
                     "1000000": "inst/extdata/gc_{genome_build}_1000kb.wig"
                     "500000": "inst/extdata/gc_{genome_build}_500kb.wig"
                     "50000": "inst/extdata/gc_{genome_build}_50kb.wig"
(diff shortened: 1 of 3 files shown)


modules/ichorcna/1.1/config/default.yaml
```diff
@@ -24,19 +24,40 @@ lcr-modules:
...
+                    hg38: ""
+                    hs37d5: ""
                 # must use gc wig file corresponding to same binSize (required)
                 ichorCNA_gcWig:
                     "1000000": "inst/extdata/gc_{genome_build}_1000kb.wig"
                     "500000": "inst/extdata/gc_{genome_build}_500kb.wig"
                     "50000": "inst/extdata/gc_{genome_build}_50kb.wig"
                     "10000": "inst/extdata/gc_{genome_build}_10kb.wig"
+                # relative to ichorCNA directory (in 00-inputs/ichorCNA/)
+                ichorCNA_gcWig_custom: # must create an empty one for each genome_build unless you have a
```
lkhilton: I think part of the comment is cut off here?

modules/ichorcna/1.1/config/default.yaml
```diff
@@ -24,19 +24,40 @@ lcr-modules:
                 ichorCNA_normalPanel: 
                     "1000000": "inst/extdata/HD_ULP_PoN_{genome_build}_1Mb_median_normAutosome_median.rds"
                     "500000": "inst/extdata/HD_ULP_PoN_{genome_build}_500kb_median_normAutosome_median.rds"
+                # relative to ichorCNA directory (in 00-inputs/ichorCNA/)
```
lkhilton, line 27: Can you provide some documentation about when/why someone might want to use custom normalPanel and gcWig files?

lkhilton (changes requested):
Just hoping for a little more documentation for the new custom features - otherwise looks good!

Kdreval:
@Jwong684 just checking if you have any updates for Laura's comments?

Jwong684:
Added requested documentation and a guide to creating a custom panel of normals - should be good now

Kdreval (approved):

Kdreval:
@lkhilton it doesn't let me merge this because of the requested changes - can you please approve if everything is OK so we can get this to master?

lkhilton (approved):

PR #285, opened 2023-11-20
Housekeeping

Kdreval:
This PR:
1. Adds a collection of per-chromosome coverage to qc module (needs samtools 1.10 env where this was first introduced, but must keep the samtools 1.09 version for the flagstat because of issues on some CentOS versions)
2. Fixes bug in R script that aggregate qc metrics (mutate does not accept multiple `.before` or `.after` in one call)
3. Fixes bug in R script that prepares SVs for lymphgen (drop `data.table::fread()` call because it can't properly set column type on empty files and replace that with `readr::read_tsv()`)
4. Drops deprecated and unused modules in demo
5. Fix bug in demo configs where `sample_bai` key was missing for qc module

### DIFF
--- a/modules/qc/1.0/config/default.yaml
+++ b/modules/qc/1.0/config/default.yaml
@@ -11,6 +11,7 @@ lcr-modules:
 
         options:
             samtools_stat: "" # can specify --remove-dups --remove-overlaps
+            samtools_coverage: "" # can specify a particular region of interest
             QualityScoreDistribution: "--VALIDATION_STRINGENCY LENIENT" # can specify --ALIGNED_READS_ONLY true
             CollectWgsMetrics: "--USE_FAST_ALGORITHM true --VALIDATION_STRINGENCY LENIENT"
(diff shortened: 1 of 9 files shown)


modules/qc/1.0/config/default.yaml
```diff
@@ -11,6 +11,7 @@ lcr-modules:
 
         options:
             samtools_stat: "" # can specify --remove-dups --remove-overlaps
+            samtools_coverage: "" # can specify a particular region of interest
```
lkhilton: Can you give an example how you would specify this? I'm guessing it's just the usual samtools `-R chr:start-end`? And if you don't specify a region, does it give coverage of the whole genome?

Kdreval: Thanks! Yes it is chr:start-end and  otherwise calculated for all contigs in bam header. I've added this to config for proper documentation!

lkhilton (approved):

PR #287, opened 2023-11-27
Patch: Changed functions in cnvkit and controlfreec to remove reliance on re…

Jwong684:
Patched controlfreec and cnvkit modules together - removed reliance on reference_files() in get functions for generating mpileups in both. In controlfreec, also updated the get function for getting fasta which follows a similar logic.


Kdreval (approved):

Kdreval:
Thanks!!

PR #291, opened 2023-12-29
bug fix: handle non-standard bed files in reference workflow

Kdreval:
Similar to a previous small update in the qc module, the reference_files workflow also needs a small fix to handle bed files with uneven number of columns


workflows/reference_files/2.4/reference_files.smk
```diff
@@ -482,7 +482,7 @@ rule get_capspace_bed_download:
         capture_bed = "genomes/{genome_build}/capture_space/{capture_space}.bed"
     conda: CONDA_ENVS["coreutils"]
     shell:
-        "ln -srf {input.bed} {output.capture_bed}"
+        "cut -f 1-3 {input.bed} > {output.capture_bed}"
```
Kdreval, line 485: This is the only modification, rest are just whitespace

lkhilton (approved):

PR #292, opened 2024-01-15
Module/dnds/1.1

gillissierra:
# Pull Request Checklists

This PR will include updates to both gistic2-1.1 and dnds, and requires https://github.com/LCR-BCCRC/lcr-scripts/pull/55

For gistic2, it is now agnostic of the columns in the subsetting_categories tsv. The seq_type no longer needs to be specified when running in GAMBL, and instead the subsetting by seq_type is done only in the RScript.

For DNDS, it now tracks sample sets by md5sum and launch date of the snakemake as well as the sample set name. The script to preprocess the maf files has the same additions as for that of gistic2.

Both ran successfully on a sample set of only capture, only genome, and both capture and genome. No changes were made to the "launch_date" parameter, which was tested in the gistic2-1.1 PR so it was not tested here.


Kdreval (approved):
Thanks, great update and I like the new setup!

Kdreval (approved):

PR #300, opened 2024-03-15
Module/oncodrivefml/1.0

mannycruz:
# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable

Example:
```
```


modules/oncodrivefml/1.0/config/default.yaml
```diff
@@ -0,0 +1,70 @@
...
+    
+    oncodrivefml:
+        
+        inputs:
+            # Available wildcards: {seq_type} {genome_build}
+            input_maf: "__UPDATE__"
+            subsetting_categories: "__UPDATE__"
+
+        genome_builds: ["__UPDATE__"]
+        # Available regions options: cds, lincrna, 3utr, 5utr, promoter_splice_sites
```
Kdreval: can you please give a short comment about these options? For example, can you specify multiple of these or only one can be specified per each run? Are they mutually exclusive or complimentary (e.g. if you give it for example 3utr does it mean it takes in cds+3utr or only exclusively 3utr)?

modules/oncodrivefml/1.0/config/default.yaml
```diff
@@ -0,0 +1,70 @@
...
+
+        maf_processing:
+            sample_sets: ["__UPDATE__"]
+            prepare_mafs: "{SCRIPTSDIR}/generate_smg_inputs/1.1/generate_smg_inputs.R"
+            include_non_coding: False
+            blacklists: []
+            blacklist_drop_threshold: 4
+
+        format_fml_input:
+            input_columns: "5,6,11,13,16"
```
Kdreval, line 36: can you please add a comment here on what are these numbers?

modules/oncodrivefml/1.0/config/default.yaml
```diff
@@ -0,0 +1,70 @@
...
+            additional_commands: ""
+
+        reference_files_directory: "ref/lcr-modules-references/downloads/oncodrive/"
```
Kdreval, line 42: I wonder if it confilicts with the paths below? If they are used at the same place does this creates the duplication of the common part `downloads/oncodrive/`?

modules/oncodrivefml/1.0/config/default.yaml
```diff
@@ -0,0 +1,70 @@
...
+        pairing_config:
+            genome:
+                run_paired_tumours: False
+                run_unpaired_tumours_with: "no_normal"
+                run_paired_tumours_as_unpaired: True
+            capture:
+                run_paired_tumours: False
+                run_unpaired_tumours_with: "no_normal"
+                run_paired_tumours_as_unpaired: True
+
```
Kdreval: missing end of file here :smile:

modules/oncodrivefml/1.0/config/default.yaml
```diff
@@ -0,0 +1,70 @@
...
+            blacklists: []
+            blacklist_drop_threshold: 4
+
+        format_fml_input:
+            input_columns: "5,6,11,13,16"
+            additional_commands: ""
+
+        reference_files_directory: "ref/lcr-modules-references/downloads/oncodrive/"
+        regions_files:
+            grch37:
```
Kdreval, line 44: maybe, it will make sense also to add these to the `etc` folder just like for hg38? How big are these files?

modules/oncodrivefml/1.0/etc/oncodrivefml.conf
```diff
@@ -0,0 +1,33 @@
+[genome]
+build = 'hg19'
```
Kdreval, line 3: does this need to be adjusted if running in hg38 mode?

modules/oncodrivefml/1.0/etc/oncodrivefml.conf
```diff
@@ -0,0 +1,33 @@
...
+
+[signature]
+method = 'complement'
+
+[score]
+# score file path must be the directory containing scoe files
+file = 
+format = 'pack'
+chr = 0
+chr_prefix = ""
```
Kdreval, line 16: same comment as above regarding possible need of adjustment for hg38?

modules/oncodrivefml/1.0/etc/oncodrivefml.conf
```diff
@@ -0,0 +1,33 @@
...
+sampling_chunk = 100
+sampling_min_obs = 10
+
+    [[indels]]
+        include = True
+        method = 'max'
+        max_consecutive = 7
+
+[settings]
+# cores = 1
```
Kdreval, line 44: the config gives 4 by default, I think? - but here it says 1, so does the value in config gets used?

modules/oncodrivefml/1.0/oncodrivefml.smk
```diff
@@ -0,0 +1,252 @@
...
+        """)
+
+ONCODRIVE_BUILD_DICT = {
+    'hg38': 'hg38',
+    'grch37': 'hg19'
+}
+
+rule _oncodrivefml_get_hg19_scores:
+    output:
+        scores = CFG["dirs"]["cadd"] + "grch37/genomicscores/caddpack/1.0-20170217/whole_genome_SNVs.fml",
```
Kdreval, line 135: is this hard-coded `grch37`, do we need to wildcard this?

modules/oncodrivefml/1.0/oncodrivefml.smk
```diff
@@ -0,0 +1,252 @@
...
+        md5sum = SUMS
+    )
+
+rule _oncodrivefml_aggregate:
+    input:
+        _get_oncodrivefml_outputs
+    output:
+        aggregate = CFG["dirs"]["outputs"] + "{genome_build}/{sample_set}--{launch_date}/aggregate/{sample_set}--{launch_date}--{region}.done"
+    shell:
+        "touch {output.aggregate}"
```
Kdreval, line 235: Usually there is a symlinking rule at the end, but here the dummy file? What is this rule intended to do here?

Kdreval (commented):
Nice work! :clap: Looking very good! :star:

modules/oncodrivefml/1.0/config/default.yaml
mannycruz: Done! I added a more detailed description. If you specify ["cds","3utr"], it will return 2 sets of output files: one set containing genes in coding sequence regions, and the other set containing genes in 3' UTR region. I added the ability to provide a "custom" region file using the format specified here (https://oncodrivefml.readthedocs.io/en/latest/files.html#regions-file-format) in case the available regions files aren't sufficient

mannycruz: # Available regions options: cds, lincrna, 3utr, 5utr, promoter_splice_sites, custom
        # You can specify more than one region at a time in list format, e.g ["cds","lincrna"]
        # This will result in 2 sets of output files, one set containing genes in coding sequence regions
        # And the other set containing long intergenic non-coding RNA results.
        # cds: Coding sequence regions
        # lincrna: Long intergenic non-coding RNA regions
        # 3utr: 3' UTR regions
        # 5utr: 5' UTR regions
        # promoter_splice_sites: Promoter regions and splice site regions
        # custom: Custom made regions file specified by user. More info for regions file format: 
        # https://oncodrivefml.readthedocs.io/en/latest/files.html#regions-file-format

modules/oncodrivefml/1.0/config/default.yaml
mannycruz: Added more details:
        format_fml_input:
            # String in input_columns is passed to the cut command to subset MAF columns to only include:
            # Chromosome, Start_Position, Reference_Allele, Tumor_Seq_Allele2, Tumor_Sample_Barcode
            # e.g "cut -f 5,6,11,13,16"
            input_columns: "5,6,11,13,16"
            # If additional columns are incorporated, add additional commands here to change column names if necessary
            # e.g additional_commands: "sed 's/Normal_Sample_Barcode/SIGNATURE/'"
            # More info: https://oncodrivefml.readthedocs.io/en/latest/files.html#file-formats
            additional_commands: ""

mannycruz: Added more info:
>
      format_fml_input:
            # String in input_columns is passed to the cut command to subset MAF columns to only include:
            # Chromosome, Start_Position, Reference_Allele, Tumor_Seq_Allele2, Tumor_Sample_Barcode
            # e.g "cut -f 5,6,11,13,16"
            input_columns: "5,6,11,13,16"
            # If additional columns are incorporated, add additional commands here to change column names if necessary
            # e.g additional_commands: "sed 's/Normal_Sample_Barcode/SIGNATURE/'"
            # More info: https://oncodrivefml.readthedocs.io/en/latest/files.html#file-formats
            additional_commands: ""

modules/oncodrivefml/1.0/config/default.yaml
mannycruz: The path in `reference_files_directory:` is only specified to be passed to the global variable `BGDATA_LOCAL`:

```
     params:
        local_path = CFG["reference_files_directory"]
 ...
     shell:
        op.as_one_line("""
        export BGDATA_LOCAL={params.local_path} &&  
```

This is required so that OncodriveFML looks in the "ref/lcr-modules-references/downloads/oncodrive/" directory for the hg19/hg38 genome reference file instead of looking in the directory "~/.bgdata/" (the global variable BGDATA_LOCAL defaults to ~/.bgdata/ if it is not manually set).

modules/oncodrivefml/1.0/config/default.yaml
mannycruz: Fixed

modules/oncodrivefml/1.0/config/default.yaml
mannycruz: The files are not very big at all. The biggest is about 2MB
https://bitbucket.org/bbglab/oncodrivefml/downloads/

I can move them to the etc folder, but they are currently set up to be downloaded through the reference files directory because they are used by both OncodriveCLUSTL and OncodriveFML. I can move the regions files into the `etc/` folder of both modules if that is preferred!

modules/oncodrivefml/1.0/oncodrivefml.smk
mannycruz: It is hard coded because only the hg19 scores are available through the bgdata framework provided by bbglab

modules/oncodrivefml/1.0/oncodrivefml.smk
mannycruz:
```
rule _oncodrivefml_out:
    input:
        tsv = str(rules._oncodrivefml_run.output.tsv),
        png = str(rules._oncodrivefml_run.output.png),
        html = str(rules._oncodrivefml_run.output.html)
    output:
        tsv = CFG["dirs"]["outputs"] + "{genome_build}/{sample_set}--{launch_date}/{md5sum}/{region}/oncodrivefml.tsv.gz",
        png = CFG["dirs"]["outputs"] + "{genome_build}/{sample_set}--{launch_date}/{md5sum}/{region}/oncodrivefml.png",
        html = CFG["dirs"]["outputs"] + "{genome_build}/{sample_set}--{launch_date}/{md5sum}/{region}/oncodrivefml.html"
    run:
        op.relative_symlink(input.tsv, output.tsv, in_module=True)
        op.relative_symlink(input.png, output.png, in_module=True)
        op.relative_symlink(input.html, output.html, in_module=True)

def _get_oncodrivefml_outputs(wildcards):
    CFG = config["lcr-modules"]["oncodrivefml"]
    checkpoint_output = os.path.dirname(str(checkpoints._oncodrivefml_prep_input.get(**wildcards).output[0]))
    SUMS, = glob_wildcards(checkpoint_output+"/{md5sum}.maf.content")

    return expand(
        [
            CFG["dirs"]["outputs"] + "{{genome_build}}/{{sample_set}}--{{launch_date}}/{md5sum}/{{region}}/oncodrivefml.tsv.gz",
            CFG["dirs"]["outputs"] + "{{genome_build}}/{{sample_set}}--{{launch_date}}/{md5sum}/{{region}}/oncodrivefml.png",
            CFG["dirs"]["outputs"] + "{{genome_build}}/{{sample_set}}--{{launch_date}}/{md5sum}/{{region}}/oncodrivefml.html"
        ],
        md5sum = SUMS
    )

```
    
The last rule looks for the final output files symlinked to 99-outputs by expanding the md5sum wildcard. This rule is a dummy file because it requires an input function to expand the md5sum values so it can't be in rule all I believe, because rule all expands based on "genome_build", "sample_set", "launch_date", and "regions" but to expand the md5sum wildcard it requires checking the outputs of the checkpoint rule

modules/oncodrivefml/1.0/etc/oncodrivefml.conf
mannycruz: Yes - I added more detailed information to the default config

modules/oncodrivefml/1.0/etc/oncodrivefml.conf
mannycruz: I added more information to this section as well

modules/oncodrivefml/1.0/etc/oncodrivefml.conf
mannycruz: It will be overwritten by the value in the config, I added more info

Kdreval:
Thanks!!

Kdreval (approved):

PR #301, opened 2024-03-19
Module/hotmaps/1.0

mannycruz:
# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable

Example:
```
```


modules/hotmaps/1.0/config/default.yaml
```diff
@@ -0,0 +1,92 @@
+lcr-modules:
+    
+    hotmaps:
+
+        inputs:
+            # Available wildcards: {seq_type}
+            # No genome_build wildcard because only grch37 is compatible!
```
Kdreval: I know it might be somewhat overwhelming but I think there is something important to address: we need to keep in some way the genome build because absence of the wildcard in this module breaks the structure of the output.
I think there is somewhat consistent and more hacky way: one option is to add the genome build wildcard as in other modules but add the check at the top of the snakefile so if this wildcard is something different then it stops and does not execute (there is an example in hmftools or gridss module). The other option is to just add it to the path of the 99-outputs  and have it hard-coded

modules/hotmaps/1.0/config/default.yaml
```diff
@@ -0,0 +1,92 @@
...
+            # Make sure input MAF paths are for grch37 only
+            input_maf: "__UPDATE__"
+            subsetting_categories: "__UPDATE__"
+
+        maf_processing:
+            sample_sets: ["__UPDATE__"]
+            prepare_mafs: "{SCRIPTSDIR}/generate_smg_inputs/1.0/generate_smg_inputs.R"
```
Kdreval: shouldn't this be 1.1 version?

modules/hotmaps/1.0/config/default.yaml
```diff
@@ -0,0 +1,92 @@
...
+        
+        vcf2maf:
+            # Path pointing to VEP files for running vcf2maf e.g "ref/ensembl_vep_cache"
+            vep_cache: "__UPDATE__"
+
+        # Path to HotMAPS repo containing updated scripts
+        hotmaps_repo: "https://github.com/morinlab/HotMAPS/archive/refs/heads/master.zip"
+
+        pdb_stucture_dirs:
+            # modbase_dir: set to the directory containing the ModBase theoretical protein structures
```
Kdreval, line 24: Can you please here maybe add comments suggesting how to get these modbase and pdb?

modules/hotmaps/1.0/config/default.yaml
```diff
@@ -0,0 +1,92 @@
...
+        hotmaps_repo: "https://github.com/morinlab/HotMAPS/archive/refs/heads/master.zip"
+
+        pdb_stucture_dirs:
+            # modbase_dir: set to the directory containing the ModBase theoretical protein structures 
+            modbase_dir: "__UPDATE__"
+            # pdb_dir: set to the directory containing PDB structures downloaded from wwPDB database
+            pdb_dir: "__UPDATE__"
+            # refseq_homology_dir: specify the path containing the ModBase RefSeq structures. 
+            # %(modbase_dir)s placeholder will be replaced with path set in modbase_dir above
+            refseq_homology_dir: "%(modbase_dir)s/ModBase_H_sapiens_2013_refseq/models/model/"
```
Kdreval, line 30: Curious what are these s after the closing brackets? :smile:

modules/hotmaps/1.0/config/default.yaml
```diff
@@ -0,0 +1,92 @@
...
+            # ensembl_homology_dir: specify path containing ModBase Ensembl structures
+            ensembl_homology_dir: "%(modbase_dir)s/H_sapiens_2013/models/"
+            # biological_assembly_dir: specify path containing wwPDB biological structures
+            biological_assembly_dir: "%(pdb_dir)s/biounit/coordinates/all/"
+            # non_biological_assembly_dir: specify the path containing wwPDB PDB structures
+            non_biological_assembly_dir: "%(pdb_dir)s/structures/all/pdb/"
+
+        options:
+            mysql:
+                mysql_host: "__UPDATE__"
```
Kdreval: can you show an example of the value that should go here?
Maybe even something like "When running on GSC, set this to this value" and provide the path that users should use?

modules/hotmaps/1.0/config/default.yaml
```diff
@@ -0,0 +1,92 @@
...
+                mysql_user: "__UPDATE__"
+                mysql_db: "__UPDATE__"
+                mysql_passwd: "__UPDATE__"
```
Kdreval, line 50: of course in the comment above I meant giving examples without showing password, this is self-explanatory :smile:

modules/hotmaps/1.0/envs/bcftools.yaml
```diff
@@ -0,0 +1,172 @@
+name: vcf2maf
```
Kdreval: can we reuse the existing vcf2maf or the one in that module is missing something?

modules/hotmaps/1.0/etc/detailed_hotspots.py
```diff
@@ -0,0 +1,734 @@
...
+                output_header = ['GENE', 'HOTSPOT_NUM', 'N_SAMPLES', 'ALL_GENES', 'HOTMAPS_RES', 'MUTATED_RES']
+                handle.write('\t'.join(output_header) + "\n")
+        with open(genomic_coordinates_out, 'w') as handle:
+                output_header = ['Hugo_Symbol', 'Chromosome', 'Start_Position', 'Hotspot_ID', 'N_samples', 'Protein_Residue', 'PDB_Structure', 'PDB_Chain', 'PDB_Chain_Position', 'Central_Residue']
+                handle.write('\t'.join(output_header) + "\n")
+    # verbose version
+    if verbose:
+        print("Proceeding with verbose output...")
+
+    # Create a hotspot dict to keep track of origin and metadata of hotspot residues
```
Kdreval: does these commented lines need to be here or can be cleaned up?

modules/hotmaps/1.0/etc/detailed_hotspots.py
```diff
@@ -0,0 +1,734 @@
...
+                            for pcr in possible_chain_residues:
+                                possible_res_p = (pdbid, chain, pcr, float(min_p))
+                                if possible_res_p in long_output_dict.keys():
+                                    chain_residue = pcr
+                                    print(f"Found PDB match for {hotspot_gene}:{residue}:\tpdb: {pdbid} chain: {chain} chain_residue: {chain_residue} p_value: {min_p}")
+                                    res_p = possible_res_p
+                                    chain_winner = chain
+                                    residue_winner = chain_residue
+                                    break
+                        #else:
```
Kdreval: same here

modules/hotmaps/1.0/hotmaps.smk
```diff
@@ -0,0 +1,837 @@
...
+        CFG["conda_envs"]["prepare_mafs"]
+    params:
+        include_non_coding = str(CFG["maf_processing"]["include_non_coding"]).upper(),
+        mode = "HotMAPS",
+        metadata_cols = CFG["samples"],
+        metadata = CFG["samples"].to_numpy(na_value='')
+    script:
+        PREPARE_MAFS
+
+rule _hotmaps_split_dnps:
```
Kdreval, line 131: nice :woman_scientist: :computer:

modules/hotmaps/1.0/hotmaps.smk
```diff
@@ -0,0 +1,837 @@
...
+        completed = CFG["dirs"]["inputs"] + "maf/{sample_set}--{launch_date}/{md5sum}_split_dnps.complete"
+    shell:
+        op.as_one_line("""
+        variant_type_col=$(head -n 1 {input.maf} | sed 's/\\t/\\n/g' | nl | grep "Variant_Type" | cut -f 1) &&
+        protein_position_col=$(head -n 1 {input.maf} | sed 's/\\t/\\n/g' | nl | grep "Protein_position" | cut -f 1) &&
+        cat <( head -n 1 {input.maf} ) <( awk -v var_col="$variant_type_col" -v protein_col="$protein_position_col" ' {{ if ( $var_col=="DNP" && $protein_col ~ /[0-9?]+[-][0-9?]+/) print $0 }} ' {input.maf} ) > {output.dnps} &&
+        touch {output.completed}
+        """)
+
+checkpoint _hotmaps_maf2vcf:
```
Kdreval, line 145: I think if the conda env is used only for this it should work with the yaml from vcf2maf module

modules/hotmaps/1.0/hotmaps.smk
```diff
@@ -0,0 +1,837 @@
...
+        stderr = CFG["logs"]["bcftools"] + "{sample_set}--{launch_date}/{md5sum}/{tumour_id}_vs_{normal_sample_id}/bcftools_norm.stderr.log"
+    shell:
+        op.as_one_line("""
+        bcftools norm --atomize 
+        --output {output.vcf} 
+        --output-type v {input.vcf} 
+        > {log.stdout} 2> {log.stderr}
+        """)
+
+rule _hotmaps_vcf2maf:
```
Kdreval, line 199: so basically you are running maf to vcf and then back to maf just to get rid of DNPs?
The potential issue with this is not sure how the alternative transcripts are handled. In gambl the "standard" vcf2maf uses a set of custom ENST because they are incorrectly annotated by vcf2maf and there are some important genes like MYD88, FOXO1 etc. Is that getting lost here after another pass to vcf2maf?

modules/hotmaps/1.0/hotmaps.smk
```diff
@@ -0,0 +1,837 @@
...
+        --mysql-user {params.mysql_user} 
+        --mysql-passwd {params.mysql_pass} 
+        --tumor-type {params.tumor_type} 
+        --no-stratify {params.hypermut} 
+        -i {params.cov_dir} 
+        --output {output.annotation} 
+        > {log.stdout} 2> {log.stderr}
+        """)
+
+rule _hotmaps_filter_hypermutated:
```
Kdreval, line 519: what's the purpose of this? Can we document it?

modules/hotmaps/CHANGELOG.md
```diff
@@ -0,0 +1,34 @@
...
+`rsync.rcsb.org::ftp_data/biounit/coordinates/all/`<br />
+`rsync.rcsb.org::ftp_data/structures/all/pdb/`<br />
+More information here:<br />
+https://www.wwpdb.org/ftp/pdb-ftp-sites <br />
+Take a look at the directory structure of the wwPDB database here:<br />
+https://files.wwpdb.org/pub/pdb/data/ <br />
+Finally, the example config provided in the HotMAPS repo can help if you are confused about which PDB structures are required and how to specify their paths in the config file: https://github.com/KarchinLab/HotMAPS_2016/blob/master/config.txt. The HotMAPS scripts have been updated such that PDB structures from either `ftp_data/structures/all/pdb/` and `ftp_data/structures/divided/pdb/` are accepted.
+
+- HotMAPS depends on a MySQL database created by the Karchin Lab that maps genomic coordinates to PDB residues. This database must be downloaded before running HotMAPS and the database name, host, user, and password information must be specified in the config. More information here: https://github.com/KarchinLab/HotMAPS_2016/wiki/MySQL-database
```
Kdreval, line 22: so this is user-specific?

Kdreval (commented):
Thanks!!

modules/hotmaps/1.0/config/default.yaml
mannycruz: The 's' means that the modbase_dir variable will be a string (for python string interpolation)

modules/hotmaps/1.0/config/default.yaml
mannycruz: Yes good catch -- fixed

modules/hotmaps/1.0/config/default.yaml
mannycruz: Added more info!

modules/hotmaps/1.0/config/default.yaml
mannycruz: Done!

modules/hotmaps/1.0/envs/bcftools.yaml
mannycruz: The vcf2maf in this env is essentially the same, the only difference is that it contains an updated bcftools version for using "bcftools norm" command. I will try to create a new environment with only the updated bcftools version and use the existing vcf2maf environment for the vcf2maf steps and run tests to make sure it works

modules/hotmaps/1.0/hotmaps.smk
mannycruz: I think you are right -- I will change the vcf2maf conda environment to use the existing one and make a separate one for bcftools and test the module to make sure it works

modules/hotmaps/CHANGELOG.md
mannycruz: This is user specific but there is a MySQL db on the GSC that contains the mupid_modbase already at 10.9.208.67

Kdreval: Thanks! Maybe those comments in the config will be helpful then with what to put in there

modules/hotmaps/1.0/hotmaps.smk
Kdreval: thanks!!

modules/hotmaps/1.0/etc/detailed_hotspots.py
mannycruz: Removed ^-^

modules/hotmaps/1.0/etc/detailed_hotspots.py
mannycruz: Removed ^-^

modules/hotmaps/1.0/hotmaps.smk
mannycruz: Hmm I didn't realize -- I'll look into this

mannycruz: I looked into the previous results and found that the output protein positions were being changed in DNP variants compared to their previous MAFs! I added the ability to provide a "custom enst" file in the options for the vcf2maf rule, supplied the one available in gambl at `etc/custom_enst.grch37.txt` and this fixed the discordant protein positions

modules/hotmaps/1.0/hotmaps.smk
mannycruz: I added a description above the rule

modules/hotmaps/CHANGELOG.md
mannycruz: I added more information to the default.yaml config for gsc users

modules/hotmaps/1.0/hotmaps.smk
mannycruz: I tested the module with the existing vcf2maf conda environment from the version 1.3 lcr-module, Initially the rule _hotmaps_vcf2maf would fail using the older vcf2maf version so I had to add an option in the default.yaml config to provide the src dir of a custom vcf2maf script
```
options:
            vcf2maf:
                src_dir: "{REPODIR}/modules/vcf2maf/1.3/src"
```
And add some more code in the rule _hotmaps_vcf2maf to point to the new script. Plus I added some error handling steps because when I ran it initially, some MAFs would return blank and had to be run through vcf2maf again! I also created a different conda environment that just contains the updated bcftools version, and I removed the old vcf2maf environment that I had made from the conda_environments directory in scratch. I tested running the module with all the changes and it works!

modules/hotmaps/1.0/config/default.yaml
mannycruz: I added the genome build wildcard and hardcoded it to grch37!

Kdreval (approved):

Kdreval:
Thanks for checking this and solving the issue with vcf2maf, and addressing other comments too!

PR #303, opened 2024-03-27
Module/igv/1.0

mannycruz:
# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable

Example:
```
```


modules/igv/1.0/config/default.yaml
```diff
@@ -0,0 +1,118 @@
+lcr-modules:
+    
+    igv:
+
+        inputs:
+            # Available wildcards: {unix_group} {seq_type} {tumour_sample_id} {normal_sample_id} {pair_status} {genome_build}
```
mannycruz: Is it okay to have `{unix_group}` as a wildcard?

Kdreval: It is not a standard column in the lcr-modules schema and is something specific to gambl. There are ways to deal with this - can show some examples and we can talk about this more before lab meeting today 😄

modules/igv/1.0/config/default.yaml
```diff
@@ -0,0 +1,118 @@
...
+            maf: "__UPDATE__"
+
+        regions: 
+            # Provide regions files as lists in their respective genome builds so that liftover of coordinates occurs properly
```
Kdreval, line 15: What happens if nothing is specified here? Can we add the ` "__UPDATE__"` to anything that has to be filled in?

modules/igv/1.0/config/default.yaml
```diff
@@ -0,0 +1,118 @@
...
+            hotmaps:
+                grch37: []
+                hg38: []
+            bed:
+                grch37: []
+                hg38: []
+            maf:
+                grch37: []
+                hg38: []
+            mutation_id:
```
Kdreval, line 29: Can we add here the example formatting of what is expected in that file? Does it have to have a header and expects certain column names?

modules/igv/1.0/config/default.yaml
```diff
@@ -0,0 +1,118 @@
...
+                grch37: []
+                hg38: []
+
+        # Stop snakefile after MAF filtering step to estimate total number of snapshots that will be taken without running IGV
+        estimate_only: False 
+
+        options:
+            genome_map: 
+                # Map metadata builds to grch37 and hg38 so that MAF file locations are determined correctly
+                grch37: ["__UPDATE__"] # e.g ["grch37","hg19","hs37d5"]
```
Kdreval: you sort of know this upfront, so maybe it is better to fill in these keys and then the comment will be that you can add other genome builds to this list?

modules/igv/1.0/config/default.yaml
```diff
@@ -0,0 +1,118 @@
...
+                hg38: ["__UPDATE__"] # e.g ["hg38","grch38"]
+
+            liftover_regions:
+                reference_chain_file:
+                    grch37: "genomes/grch37/chains/grch37/hg19ToHg38.over.chain"
```
Kdreval: I think this is generated with the reference files so you can directly request the reference files output instead of having this listed here. This can be an example:
https://github.com/LCR-BCCRC/lcr-modules/blob/7b495a3e92c3813cc58ff6569c92476cbc6e53c0/modules/liftover/2.0/liftover.smk#L108-L130

modules/igv/1.0/config/default.yaml
```diff
@@ -0,0 +1,118 @@
...
+            genome:
+                run_paired_tumours: True
+                run_unpaired_tumours_with: "unmatched_normal"
+                run_paired_tumours_as_unpaired: False
+            capture:
+                run_paired_tumours: True
+                run_unpaired_tumours_with: "unmatched_normal"
+                run_paired_tumours_as_unpaired: False
+
+    slms_3:
```
Kdreval: can you instead just include config from slms-3?
I think this will conflict if you have snakefile that will attempt to run both slms-3 and this module because snakemake won't like duplicated config values - or will just use whatever is imported last so may have unexpected consequence in that scenario

modules/igv/1.0/envs/crossmap.yaml
```diff
@@ -0,0 +1,56 @@
+name: crossmap
```
Kdreval: Can you please check if this env can be re-used instead?
https://github.com/LCR-BCCRC/lcr-scripts/blob/master/envs/crossmap/convert_maf_coords.yaml

modules/igv/1.0/envs/format_regions.yaml
```diff
@@ -0,0 +1,38 @@
+name: test-generate_igv_batch
```
Kdreval: What is the "acting part" here? Is it just to have python env? We can then see if there are other envs in lcr-modules that we can re-use here instead of adding new one 😄

modules/igv/1.0/etc/liftover_regions.sh
```diff
@@ -0,0 +1,73 @@
+#!/bin/bash
```
Kdreval: This did not work? https://github.com/LCR-BCCRC/lcr-scripts/blob/master/crossmap/1.1/convert_maf_coords.sh
If this script has something specific the existing lcr-script doesn't support already, should we instead add that as 1.2 version instead of creating new one? 😄

modules/igv/1.0/igv.smk
```diff
@@ -0,0 +1,824 @@
...
+# Setup module and store module-specific configuration in `CFG`
+# `CFG` is a shortcut to `config["lcr-modules"]["igv"]`
+CFG = op.setup_module(
+    name = "igv",
+    version = "1.0",
+    subdirectories = ["inputs", "batch_scripts", "igv", "snapshots", "outputs"],
+)
+
+# Rename genome_build values in sample metadata to correlate with MAF values
+CFG["runs"]["tumour_genome_build"].mask(CFG["runs"]["tumour_genome_build"].isin(CFG["options"]["genome_map"]["grch37"]), "grch37", inplace=True)
```
Kdreval, line 47: I don't get what's happening here, sorry! Can you show example of what this does?

modules/igv/1.0/igv.smk
```diff
@@ -0,0 +1,824 @@
...
+# Symlinks the input files into the module results directory (under '00-inputs/')
+rule _igv_symlink_bams:
+    input:
+        bam = get_bam,
+        bai = get_bai
+    output:
+        bam = CFG["dirs"]["inputs"] + "bams/{seq_type}/{sample_id}.bam",
+        bai = CFG["dirs"]["inputs"] + "bams/{seq_type}/{sample_id}.bam.bai"
+    run:
+        op.absolute_symlink(input.bam, output.bam)
```
Kdreval, line 124: Curious if this needs crai index whe the bam is a cram 😃

modules/igv/1.0/igv.smk
```diff
@@ -0,0 +1,824 @@
...
+        op.absolute_symlink(input.bai, output.bai)
+
+rule _igv_symlink_maf:
+    input:
+        maf = get_maf
+    output:
+        maf = CFG["dirs"]["inputs"] + "maf/{seq_type}--{genome_build}/{tumour_id}--{normal_sample_id}--{pair_status}.maf"
```
Kdreval, line 131: I think this might be missing the `sample` in the tumour_id wildcard?
Here is the part from config:
```
# Available wildcards: {unix_group} {seq_type} {tumour_sample_id} {normal_sample_id} {pair_status} {genome_build}
```

modules/igv/1.0/igv.smk
```diff
@@ -0,0 +1,824 @@
...
+    params:
+        regions_format = lambda w: w.tool_type,
+        regions_build = lambda w: w.tool_build
+    log:
+        stdout = CFG["logs"]["inputs"] + "regions/format_regions_{tool_type}.{tool_build}.stdout.log",
+        stderr = CFG["logs"]["inputs"] + "regions/format_regions_{tool_type}.{tool_build}.stderr.log"
+    script:
+        config["lcr-modules"]["igv"]["scripts"]["format_regions"]
+
+REGIONS_FORMAT = {
```
Kdreval: Why do we need a dictionary if everything is mapped to the same key `bed`?  Maybe we can just refer to bed regardless of what is the region format?

modules/igv/1.0/igv.smk
```diff
@@ -0,0 +1,824 @@
...
+        metadata = CFG["runs"]
+    log:
+        stdout = CFG["logs"]["inputs"] + "filter_maf/{seq_type}--{genome_build}/{tumour_id}--{normal_sample_id}--{pair_status}.stdout.log",
+        stderr = CFG["logs"]["inputs"] + "filter_maf/{seq_type}--{genome_build}/{tumour_id}--{normal_sample_id}--{pair_status}.stderr.log"
+    wildcard_constraints:
+        seq_type = "[a-zA-Z]+"
+    script:
+        config["lcr-modules"]["igv"]["scripts"]["filter_script"]
+
+def _get_maf(wildcards):
```
Kdreval, line 266: Curious why this does not conflict with the function defined above?
```
def get_maf(wildcards):
    unix_group = config["unix_group"]
    return expand(config["lcr-modules"]["igv"]["inputs"]["maf"], allow_missing=True, unix_group=unix_group)
```
Just because it is defined later the further call will use the "latest" definition?

modules/igv/1.0/igv.smk
```diff
@@ -0,0 +1,824 @@
...
+                gene = maf_table["Hugo_Symbol"],
+                allow_missing = True
+            ),
+            preset = wildcards.preset,
+            allow_missing = True
+        )
+
+rule _igv_download_igv:
+    output:
+        igv_zip = CFG["dirs"]["igv"] + "IGV_2.7.2.zip",
```
Kdreval: What if someone doesn't run this on linux or wants to have a different version? Maybe this part should be configurable or better even use the conda instead of downloading the source file?
https://anaconda.org/bioconda/igv

modules/igv/1.0/config/default.yaml
mannycruz: I added an `"__UPDATE__"` string and specified that at least one regions file must be provided

modules/igv/1.0/config/default.yaml
mannycruz: Done ! I added what column and format is required for mutation_id file format

modules/igv/1.0/config/default.yaml
mannycruz: Done ✅

modules/igv/1.0/config/default.yaml
mannycruz: Done ✅

modules/igv/1.0/config/default.yaml
mannycruz: Yes good idea, I removed this part from the config

modules/igv/1.0/envs/crossmap.yaml
mannycruz: I switched to using liftover so this env file is not needed anymore, I removed it from the module

modules/igv/1.0/envs/format_regions.yaml
mannycruz: This is a leftover artifact from when I tried to use VCF files to specify regions to snapshot, I removed it since it's not necessary anymore

modules/igv/1.0/etc/liftover_regions.sh
mannycruz: I switched to using liftover and moved the main functions of the liftover_regions.sh script to the actual rule itself , so I can remove this shell script from the module

modules/igv/1.0/igv.smk
mannycruz: if the runs table has a genome build specified as "hg38-reddy", this will change the value to "hg38" just to make sure that when looking for sample MAF files, it looks for the path "{tumour_id}--{normal_sample_id}--{pair_status}.slms-3.final.hg38.maf" instead of "{tumour_id}--{normal_sample_id}--{pair_status}.slms-3.final.hg38-reddy.maf"

modules/igv/1.0/igv.smk
mannycruz: I removed the earlier function because I do not need it to fill in unix_group value (this can be done from the snakefile that is used to launch the run), so not an issue anymore

modules/igv/1.0/igv.smk
mannycruz: You're right- I removed it

mannycruz:
I have addressed most of the comments above, but I need to further investigate regarding:

- if crams can be used for snapshots https://github.com/LCR-BCCRC/lcr-modules/pull/303#discussion_r1543446315
- downloading IGV vs using conda environment https://github.com/LCR-BCCRC/lcr-modules/pull/303#discussion_r1543483073

plus i have made added some more changes since, so I will ask for an updated review once I'm finished

Issue #306, opened 2024-04-08
fishHook - add covariates option


gillissierra:
The module is currently written to only take in ONE covariate bed file. Users have been using more than one, for example
- GC content, provided by fishHook developers http://mskilab.com/fishHook/hg19/nucleotide.context.rds
- heterochromatin regions
- replication timing
- expression based on a DESeq2 vst expression matrix

The option to use these can be added to the R script, with a parameter in the yaml to specify which ones to use. These files will be static parts of the lcr-module. If future covariates are used, they can be added later.

Issue #305, opened 2024-04-08
fishHook - add gene list


gillissierra:
- Add the GRCh37 gene list that is provided by the developers. http://mskilab.com/fishHook/hg19/gencode.v19.genes.gtf'))
- Change the yaml to have a TRUE/FALSE for whether to use the gene list in the model. Note in the yaml that it is mutually exclusive with finding significant regions using tiles, which will remain the default.

Related is the lcr-scripts [Issue 60](https://github.com/LCR-BCCRC/lcr-scripts/issues/60), which will ensure the input to the fishHook script is for GRCh37

Kdreval:
This can be closed now, right?

gillissierra:
Not yet, I haven't implemented this

gillissierra:
To keep this flexible with the use of either tiles or gene lists, the module can stay as is. Users following the fishHook developer's tutorial will be able to easily find the gene list .gtf they used.

PR #307, opened 2024-04-08
Module/fishhook/1.1

gillissierra:
# Pull Request Checklists

- Added md5sum and launch data tracking for sample sets
-Ran 1.0 module and ran after thisincorporating this update, both with defaults (tiling, not gene list). Outputs are the same.
- [Issue 305](https://github.com/LCR-BCCRC/lcr-modules/issues/305) will address gene list input and [Issue 306](https://github.com/LCR-BCCRC/lcr-modules/issues/306) will address covariate input

Example:
```
```


modules/fishhook/1.0/fishhook.smk
```diff
@@ -116,23 +115,27 @@ rule _install_fishhook:
         input = CFG["logs"]["inputs"] + "install_fishhook.log"
     shell:
         """
-        R -q -e 'devtools::install_github(c("jokergoo/ComplexHeatmap","mskilab/gTrack", "mskilab/fishHook"))' >> {log.input} &&
-        touch {output.complete}"""
-
-
-# Example variant calling rule (multi-threaded; must be run on compute server/cluster)
-rule _run_fishhook:
+        R -q -e 'devtools::install_github("mskilab/gUtils")' >> {log.input} &&
```
Kdreval: so the Complexheatmap is not required?

Kdreval (approved):
Thanks for update!
Just one small comment, otherwise good to go :smile: 
Not merging since this will be bumped to the new version

modules/fishhook/1.0/fishhook.smk
gillissierra: I don't believe so. It may have been that the developers removed it's dependency. I followed their installation instructions [here](https://github.com/mskilab-org/fishHook?tab=readme-ov-file#installation) which don't mention it. 

Just to double check, I found and activated the fishhook conda env that snakemake created when running this through GAMBL.
- `conda activate <env named by md5sum>`
- `which R` shows the one in this env
- R
    - `find.package("ComplexHeatmap")` gives an error 
    ```
    Error in find.package("ComplexHeatmap") : 
        there is no package called ‘ComplexHeatmap’
    ```
    - `library(ComplexHeatmap)` gives an error
    ```
    Error in library(ComplexHeatmap) : 
        there is no package called ‘ComplexHeatmap’
    ```

**Conclusion**: Package most likely not needed

Kdreval (approved):

PR #312, opened 2024-05-10
Introducing: mm10 compatibility!

lkhilton:
These updates will require version bumps to the reference files workflow because of significant changes to how the gencode references are handled for Salmon. A near-future update will include changes to vcf2maf to use the mouse VEP!


Kdreval (approved):
Important update!

lkhilton:
@Kdreval I've finished the version bump - could you please take a look and re-approve?

Kdreval (approved):
Thanks, I think looks great!

PR #313, opened 2024-05-14
Small patch for better cram compatibility

lkhilton:
The Sage module was written to assume that the input alignment had a ".bam" and ".bam.bai" extension - which obviously doesn't hold if you have cram files as input. Rather than updating the input to include another `CFG["inputs"]["bai"]`, which would require updates to existing run scripts that use  Sage, I included a conditional "check if the .bai exists."


Kdreval (approved):
Thanks!!

PR #316, opened 2024-06-26
Module/fishhook/1.1 - can now use covariates supplied in yaml

gillissierra:
# Pull Request Checklists

- [x]  tested tiles model with and without using covariates in yaml

- [x] testing gene model with and without covariates

- [x] tested on all seq_type combinations

- [x] updated changelog

Example:
```
```


kmcoyle (dismissed):

modules/fishhook/1.1/config/default.yaml
```diff
@@ -15,10 +15,13 @@ lcr-modules:
...
             include_silent_mutation: False # Include silent mutations in maf file
-            tiles_size: 1000 # This is the size of tiles you want to split the genome into, leave blank if you want to use a gene list as hypothese
-            target_gene_list: "" # If you do not provide a tile size, please provide a gene list in gtf format (e.g GENCODE v19 genes for GRCH37: gencode.v19.genes.gtf) for fishHook to do a gene-centric analysis
+            tiles_size: 1000 # This is the size of tiles you want to split the genome into, if target_gene_list is set to False
+            target_gene_list: False # Set to True if you want to use a gene list instead of tiling
             target_gene_list_only_protein_coding: True # Only include protein coding genes in gene list
-            covariates: "" # Please supply a covariate file in bed format, leave it blank if you want to run fishHook without covariates
+            covariates:
+            # comment out lines below if you do not want to use covariates
+                chrom_hmm: "/projects/dscott_prj/CCSRI_1500/exomes/ref/resources/E032_primaryBcells_fromPB_chromHMM_15state_segments.bed"
```
Kdreval: Can we set these to `__UPDATE__` in the module and leave it to the user to specify path - and for the gambl analysis these file paths can be provided in the gambl config.
This will make it possible to use outside of GSC, otherwise the user will have error about file missing and they will (probably) have no idea where this is coming from :smile:

Kdreval (commented):
only one small comment, thanks!!

gillissierra (commented):
> Can we set these to `__UPDATE__` in the module and leave it to the user to specify path - and for the gambl analysis these file paths can be provided in the gambl config.
This will make it possible to use outside of GSC, otherwise the user will have error about file missing and they will (probably) have no idea where this is coming from :smile:  

ahh yes of course!

Kdreval (approved):
Thanks!

PR #317, opened 2024-07-15
Ega download module

Kdreval:
# Pull Request Checklists

### Required

  ```bash
  ```


modules/ega_download/1.0/config/default.yaml
```diff
@@ -0,0 +1,52 @@
...
+              # in the output of pyega3 -cf </Path/To/CREDENTIALS_FILE> files EGAD<NUM>
+            # file_format: format of the files to be downloaded. For example: bam, cram, fq.gz, fastq.gz
+            # seq_type: seq type of the sample
+            # genome_build: genome build to which the samples are already aligned or will be aligned
+            # EGAF: file ID from EGA
+            # EGAN: sample ID from EGA
+            # EGAD: dataset ID from EGA
+            # EGAS: study ID from EGA
+            sample_csv: "__UPDATE__" # csv sample table containing columns described above.
+        study_id: "__UPDATE__" # String with cohort name that helps you easily identify the downloaded samples.
```
gillissierra: Could this have an example value? It's not clear to me if it should be the EGAS or what would later go into the gambl metadata under "cohort"

modules/ega_download/1.0/ega_download.smk
```diff
@@ -0,0 +1,176 @@
...
+            [
+                str(rules._ega_output_files.output.ega_file)
+            ],
+            zip,  # Run expand() with zip(), not product()
+            sample_id=CFG["samples"]["sample_id"],
+            file_format=CFG["samples"]["file_format"],
+            seq_type=CFG["samples"]["seq_type"],
+            study_id=[CFG["study_id"]]*len(CFG["samples"]["file_name"]),
+            read = [""]
+        ) if out_file_type in ["cram", "bam"] else expand(
```
gillissierra, line 155: I've never seen an ifelse formatted like this :o  Does it not work if the `if` is stated at the top, instead of after the first `expand()` as it is here? 
Nothing wrong with this, I'm just curious about it

modules/ega_download/CHANGELOG.md
```diff
@@ -0,0 +1,25 @@
...
+All notable changes to the `ega_download` module will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [1.0] - 2021-10-06
+
+This release was authored by Kostiantyn Dreval.
+
+- The input is a sample table of the following example:
```
gillissierra, line 12: I love this example table here. Is it worth putting a note in the default.yaml that this example is here in the changelog? I don't think users would think to look here for an example. The description in the default.yaml seems okay as it is though, if you don't want to mention this example table.

modules/ega_download/1.0/ega_download.smk
Kdreval: haha, I copied this from one of the other modules some time ago and follow it since then - I think this is a more "pythonic" way of a one-liner. The standard way should work too, I think :smile:

gillissierra (commented):
Was this tested on an EGA dataset that is currently in GAMBL to see if it's outputs match what is already existing? PBL_UKM (EGAS00001004659) was straight forward when I was scraping the EGA IDs for the dataframe I'm making

Kdreval:
Yep, this was tested and all the samples in the recent gambl dataset are downloaded with this module. They live here: `/projects/rmorin_scratch/ega_download/` and this is the gambl PR that added that specific cohort: https://github.com/morinlab/gambl/pull/322

modules/ega_download/1.0/ega_download.smk
gillissierra: great to know!

gillissierra (approved):

Kdreval:
Thanks  for the quick turnaround @gillissierra !

PR #318, opened 2024-07-15
Updates to bwa and utils modules

Kdreval:
The alignment to bam is not used as much anymore and the cram-compression is more often required. This PR adds config to the bwa_mem module which will dictate whether the output should be cram-compressed (default) or not (reverts to legacy behaviour). No version update is needed as the update not significant enough to be not compatible with previous outputs.
In addition, the default resources specified in the config are pretty generous and in turn some even exomes may take veeeeryyyy loooooong time to process. Easy to address in the project config but also made default values more sensible here.
This setup was tested on the gambl data and used in the download of samples from this PR: https://github.com/morinlab/gambl/pull/322

This PR also contains fix for handling of empty files in the vcf2maf reannotation reported in the issue #321

### DIFF
--- a/modules/bam2fastq/1.2/config/default.yaml
+++ b/modules/bam2fastq/1.2/config/default.yaml
@@ -1,5 +1,5 @@
 lcr-modules:
-    
+
     bam2fastq:
 
         inputs:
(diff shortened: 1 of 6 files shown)


Kdreval:
Will migrate the cram compression rule to the utils module so it can be re-used by STAR module :star2:

gillissierra (approved):

Kdreval:
Thanks @gillissierra !!

PR #322, opened 2024-07-30
Module/rainstorm/1.1 md5sum tracking update

gillissierra:
- Compared outputs of 1.0 module and this update for hg38 inputs on a DLBCL subset, and saw no differences. 

- Ensured this module works when `options: genome_build:` is a list like `["hg38", "grch37"]`

# Pull Request Checklists

Example:
```
```


modules/rainstorm/1.0/config/default.yaml
```diff
@@ -1,30 +1,37 @@
...
 
+        include_non_coding: True
+        prepare_mafs: "{SCRIPTSDIR}/generate_smg_inputs/1.1/generate_smg_inputs.R"
+
         options:
-            cohort_name: "__UPDATE__" # this will be used as prefix to the output files
-            rainstorm_version: 0.3 # here the version tag for the rainstorm version can be specified. The default version used in the module is 0.3
-            genome_build: "__UPDATE__" # specify genome build of variants in maf file
+            rainstorm_version: 0.4 # here the version tag for the rainstorm version can be specified.
+            genome_build: "__UPDATE__" # specify genome build of variants in maf file, hg38 or grch37
```
Kdreval: Maybe the example from the PR can also be added here as comment?
`["hg38", "grch37"]`
so that the user knows it can be  a list

Kdreval (dismissed):
Thanks!

Kdreval (approved):

PR #325, opened 2024-08-21
Level3 modules new directory for prepare step

gillissierra:
The prepared mafs from the generate_smg_inputs.R lcr-script are now in their own directory. Input file symlinks are named using wildcards to better track which files were used as inputs. Level3 modules that were tested with this new setup and the prepare step using threads, as in [this PR](https://github.com/LCR-BCCRC/lcr-scripts/pull/67):

- gistic2
- dnds
- mutsig
- rainstorm
- fishhook
- hotmaps
- oncodrivefml
- oncodriveclustl

# Pull Request Checklists

### If applicable

Example:
```
```

```
CFG = op.setup_module(
    name = "strelka",
    version = "1.2",
)
```


lkhilton (commented):
This is a very reasonable change. Out of curiosity (before I approve without a version bump) does it trigger existing pipelines to re-run since the names of all the numbered directories will be different after this PR?

gillissierra:
version bumps complete

lkhilton (approved):

PR #326, opened 2024-08-23
LymphGen Updates

lkhilton:
This PR includes significant updates to both LymphGen 1.0 and 2.0 modules that will require a version bump. 

Related changes have been made to [LGenIC](https://github.com/LCR-BCCRC/LGenIC/pull/1) and [LymphGen](https://github.com/LCR-BCCRC/LymphGen/pull/2) to accommodate these changes. LGenIC has now been release tagged and the correct release is linked here. It will maintain backwards compatibility with the old LymphGen modules. 

There are also some modifications to Battenberg modules using an improved method for installing non-Conda R packages. This will force these conda envs to be rebuilt, but should not trigger any re-runs of any existing outputs, therefore I don't think they require a version bump.


lkhilton:
> I don't know if this was already reviewed by someone else, maybe? I looked over anyways and the only question I have is whether the update to vcf2maf will trigger rerun of stuff in gambl? Or we may need a version bump?

Thanks Kostia. Kurt is reviewing the LymphGen changes. The changes to vcf2maf will not trigger a re-run because the GAMBLR `config.yml` is specified as an input with the `ancient()` flag, and the input is the same as it was previously. The new GAMBLR conda environment has now been built in our shared conda environment directory, so that will not be rebuilt by anyone (nor would building a new conda environment trigger a re-run).

Kdreval:
Ok, sounds good! Looks good to me - can approve any time!

Kdreval:
sorry just double-checked - can you please add `in_module = True` to the final symlinking rule?
Thanks!

lkhilton:
Will do!

kyaki:
Setup seems a lot easier than before, and more flexible to add to an existing pipeline.  Tested it on the montreal retrospective and prospective ctDNA cohort which includes only SNV data, and many samples that have empty .mafs. Setup was relatively painless, and it worked!

One note on setup:
-In the config I had to comment out "svar_master" and "manta", or else they get added to the caller list, so that messes up the code that looks for the caller to just be "empty_sv", minor, but maybe a note in the config to do that would be nice.

lkhilton:
The version bump is ready for approval!

kyaki (approved):

PR #328, opened 2024-09-10
Level 3 modules: correct version in op.setup_module

gillissierra:
Fix ensures outputs are written to the correct directories


Kdreval (approved):
Thanks @gillissierra !

PR #332, opened 2024-09-19
demo: new structure

Kdreval:
This PR takes advantage of the new minimal env recipe for lcr modules and updates the structure of the demo along with cleaning up the data directory and sample table. This has all been tested in dry run and the next step will be to fix all unreproducible conda envs while making the real run with this new setup.
The genome build for capture samples is changed to show example how the samples can be realigned to new genome build and the realigned samples to be used in variant calling.


lkhilton (approved):
I thought I had reviewed this back in September - it looks good!

Kdreval:
thanks!

PR #340, opened 2024-12-09
Fishhook fixes and level3 modules metadata dimensions are passed to the prepare script

gillissierra:
See lcr-scripts PR https://github.com/LCR-BCCRC/lcr-scripts/pull/70

This was tested with fishhook, hence the branch name, although all level3 modules have been updated. They should not reuire a version bump since outputs should remain the same.

This PR also includes adding a post-deploy script to install the fishhook R package in the conda env. Other minor issues found and corrected were

 - the "mode" parameter for the prepare step was mispelled
 - the fishhook RScript used numbers to get the inputs/outputs/parameters from the snakemake list object in R, which changed when no longer needing the install rule. So the script now uses the names in the list object

### DIFF
--- a/envs/fishhook/fishhook.post-deploy.sh
+++ b/envs/fishhook/fishhook.post-deploy.sh
@@ -2,6 +2,6 @@
 set -o pipefail
 set -e # Exit on error
 
-R --vanilla -q -e 'devtools::install_github("mskilab/gUtils", dependencies = FALSE, upgrade = "never")' > /home/sgillis/fishhook_deploy.txt
-R --vanilla -q -e 'devtools::install_github("mskilab/gTrack", dependencies = FALSE, upgrade = "never")' >> /home/sgillis/fishhook_deploy.txt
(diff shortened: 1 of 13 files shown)


envs/fishhook/fishhook.post-deploy.sh
```diff
@@ -0,0 +1,7 @@
+#!env bash
+set -o pipefail
+set -e # Exit on error
+
+R --vanilla -q -e 'devtools::install_github("mskilab/gUtils", dependencies = FALSE, upgrade = "never")' > /home/sgillis/fishhook_deploy.txt
```
Kdreval: Does this path to home dirs needs to be updated?

Kdreval (commented):
There seem to be few paths in the deploy script that point to home directory - do they need to be updated?

envs/fishhook/fishhook.post-deploy.sh
gillissierra: It needs to be removed entirely actually. Thanks for catching that!

Kdreval (approved):
Thanks!

PR #341, opened 2024-12-13
dnds support custom reference for hg38

Kdreval:
This will add support for hg38 in dnds cv


modules/dnds/1.2/src/R/run_dnds.R
```diff
@@ -27,6 +27,18 @@ maf = read.table(
 
 maf = as.data.frame(maf)
 
+if (grepl("38", maf$NCBI_Build[1])) {
+    cat("Will run using the hg38 reference.\n")
+    load(snakemake@params[[3]]) # This will load the RefCDS from the path at config
```
rdmorin: Isn't there a way to use named parameters here instead of referring to them positionally?

rdmorin (commented):
This looks fine. I noticed while testing this that when a gene list is supplied, the covariates are ignored. This isn't really clearly documented but perhaps we should change the code to not supply the covariates when a gene list is supplied and put some comments in there explaining why so it's more clear?

Kdreval:
> This looks fine. I noticed while testing this that when a gene list is supplied, the covariates are ignored. This isn't really clearly documented but perhaps we should change the code to not supply the covariates when a gene list is supplied and put some comments in there explaining why so it's more clear?

Will do!

modules/dnds/1.2/src/R/run_dnds.R
Kdreval: yep, there is - was just going with a consistent style to match other spots in this script where we refer to the snakemake objects but will swicth!

rdmorin (approved):

PR #342, opened 2025-01-18
First draft of fusioncatcher version 1.0

rdmorin:
# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable

Example:
```
```

```
CFG = op.setup_module(
    name = "strelka",
    version = "1.2",
)
```


rdmorin:
Some reflections after running this on some more samples in GAMBL:
* Temporary space used by the pipeline can exceed 150G for a single sample (not counting the input fastq that are created by bam2fastq). This might be exacerbated by low-quality data as the unmapped reads from bowtie are written to an uncompressed file. I haven't figured out if there's a way to reliably predict this footprint. Presumably users will need to set a resource limit using either fastq or scratch_gb
* If running on a large data set, grouping of jobs is necessary to ensure the fastq are cleared as soon as possible (i.e. when fusioncatcher completes)

modules/fusioncatcher/1.0/fusioncatcher.smk
```diff
@@ -0,0 +1,150 @@
...
+
+
+rule _fusioncatcher_run:
+    input:
+        fastq_1 = str(rules._fusioncatcher_input_fastq.output.fastq_1),
+        fastq_2 = str(rules._fusioncatcher_input_fastq.output.fastq_2),
+        fastq_1_real = CFG["inputs"]["sample_fastq_1"], # Prevent premature deletion of fastqs marked as temp
+        fastq_2_real = CFG["inputs"]["sample_fastq_2"],
+        ref_path = "/projects/rmorin/projects/gambl-repos/gambl-rmorin/fusioncatcher/fusioncatcher/data/human_v102/"
+    output:
```
lkhilton, line 74: Does the tool clean up its own temp files, or do you need to include some outputs marked as `temp()` to clear them?

modules/fusioncatcher/1.0/fusioncatcher.smk
```diff
@@ -0,0 +1,150 @@
...
+
+
+
+rule _fusioncatcher_run:
+    input:
+        fastq_1 = str(rules._fusioncatcher_input_fastq.output.fastq_1),
+        fastq_2 = str(rules._fusioncatcher_input_fastq.output.fastq_2),
+        fastq_1_real = CFG["inputs"]["sample_fastq_1"], # Prevent premature deletion of fastqs marked as temp
+        fastq_2_real = CFG["inputs"]["sample_fastq_2"],
+        ref_path = "/projects/rmorin/projects/gambl-repos/gambl-rmorin/fusioncatcher/fusioncatcher/data/human_v102/"
```
lkhilton, line 73: Do you want to come up with a way to get the ref files automagically and put that in a rule?

lkhilton: It doesn't *have* to be via the reference files workflow, since these are specific to this module. See the GRIDSS/hmftools modules as examples, if you want. It can be as simple as putting this file into `/gsc/www/bcgsc.ca/downloads/lcr-modules` and then referencing that URL in a rule.

modules/fusioncatcher/CHANGELOG.md
```diff
@@ -0,0 +1,14 @@
...
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [1.0] - 2025-01-17
+
+This release was authored by Ryan Morin.
+
+* The [fusioncatcher](https://github.com/ndaniel/fusioncatcher/tree/master) pipeline conveniently has a conda recipe available, which allowed almost all the dependencies to be handled automatically. 
+* I haven't incorporated the installation of the reference files into this initial version of the module. Instead, the module expects the user to do that themselves. Because the process is straightforward, well documented and doesn't require any dependencies, this seemed reasonable to me.
```
lkhilton, line 13: See my comment above - I think you should make a rule for downloading the references but just make it a really simple one. This would also insulate against those files disappearing from or getting corrupted where they are currently stored.

modules/fusioncatcher/1.0/envs/fusioncatcher_nobuilds.yaml
```diff
@@ -0,0 +1,128 @@
+name: fusioncatcher
```
lkhilton, line 1: Could you regenerate this using the `--from-history` flag? You can specify the versions of the specific packages you requested, but otherwise the dependencies can be resolved during environment construction. We've had lots of trouble with our oldest conda environment yaml files being non-reproducible after a few years.

modules/fusioncatcher/1.0/envs/fusioncatcher_nobuilds.post-deploy.sh
```diff
@@ -0,0 +1,13 @@
...
+set -e # Exit on error
+
+#get path to broken file and overwrite it with one we know works
+
+broken_bin=`which faToTwoBit`
+echo $broken_bin
+
+# Warning to would-be testers: 
+# don't uncomment/test this until the environment is not in use because it will theoretically clobber a file that may be in use!
+# curl https://hgdownload.soe.ucsc.edu/admin/exe/linux.x86_64/faToTwoBit > $broken_bin
```
lkhilton: What is this doing if the only command is commented? It should only ever run in an environment that is under construction, so it should never clobber an in-use file.

modules/fusioncatcher/1.0/envs/fusioncatcher_nobuilds.yaml
lkhilton: Also this env should be in `lcr-modules/envs/fusioncatcher/1.0` and symlinked to the module directory.

lkhilton (changes requested):

modules/fusioncatcher/1.0/fusioncatcher.smk
```diff
@@ -0,0 +1,148 @@
...
+    resources:
+        **CFG["resources"]["fusioncatcher_run"]
+    params:
+        reference = CFG["options"]["reference_path"],
+        out_dir = CFG["dirs"]["fusioncatcher"] + "{seq_type}--{genome_build}/{sample_id}/"
+    shell:
+        op.as_one_line("""
+        fusioncatcher --no-update-check -d {params.reference} -i {input.fastq_1},{input.fastq_2}
+        -o {params.out_dir} > {log.stdout} 2> {log.stderr} &&
+        touch {output.complete}
```
kyaki, line 98: I dont think it would be incompatible with oncopipe stuff, but snakemake has a touch() function that can create/touch sentinel files. Just wrap the output path with touch() in the output directives. Helps be more explicit in the rule setup 

eg 
output:
           complete = touch(CFG["dirs"]["fusioncatcher"] + "{seq_type}--{genome_build}/{sample_id}/fusioncatcher.complete"),

PR #344, opened 2025-02-05
Update CNV callers

Kdreval:
The version of fill_segments was hard-coded across the board and was not flexible. This update allows user to specify which version of that script they want to use through config, and this bug fixes all of battenberg 1.1, 1.2, sequenza, controlfreec, and cnvkit 🙌


gillissierra (approved):

Kdreval:
Thanks @gillissierra!! 🚀

PR #349, opened 2025-04-15
New DLBclass module

lkhilton:
# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable

Example:
```
```

```
CFG = op.setup_module(
    name = "strelka",
    version = "1.2",
)
```

### DIFF
--- a/envs/dlbclass/dlbclass.yaml
+++ b/envs/dlbclass/dlbclass.yaml
@@ -1,0 +1,16 @@
+name: dlbclass-test
+channels:
+  - conda-forge
+  - bioconda
+  - nodefaults
+dependencies:
+  - pandas==1.3.5
+  - matplotlib==3.8.4
+  - numpy==1.21.5
+  - pip==24.3.1
+  - python=3.9.21
+  - pytorch==1.10.2
+  - notebook==7.3.2
+  - r-base=4.4.2
+  - pip:
+      - jupyter==1.1.1
--- a/modules/dlbclass/1.0/dlbclass.smk
+++ b/modules/dlbclass/1.0/dlbclass.smk
@@ -46,7 +46,9 @@ CFG = op.setup_module(
 
 # All rules should be run locally
 localrules:
-    _dlbclass_input_variants,
+    _dlbclass_input_maf,
+    _dlbclass_input_seg,
+    _dlbclass_input_sv,
     _dlbclass_input_subsetting_categories,
     _dlbclass_download_dlbclass,
     _dlbclass_download_refs,
@@ -73,18 +75,28 @@ PREPARE_SVS =  os.path.abspath(config["lcr-modules"]["dlbclass"]["scripts"]["pre
 TODAY = datetime.now().strftime("%d%b%Y")
 
 # Symlinks the input files into the module results directory (under '00-inputs/')
-rule _dlbclass_input_variants:
+rule _dlbclass_input_maf:
     input:
-        maf = CFG["inputs"]["master_maf"], 
-        seg = CFG["inputs"]["master_seg"], 
-        sv = CFG["inputs"]["master_sv"]
+        maf = CFG["inputs"]["master_maf"]
     output:
-        maf = CFG["dirs"]["inputs"] + "variants/{sample_set}--{launch_date}/{seq_type}.input.maf", 
-        seg = CFG["dirs"]["inputs"] + "variants/{sample_set}--{launch_date}/{seq_type}.input.seg", 
-        sv = CFG["dirs"]["inputs"] + "variants/{sample_set}--{launch_date}/{seq_type}.input.sv"
+        maf = CFG["dirs"]["inputs"] + "variants/{sample_set}--{launch_date}/{seq_type}.input.maf"
     run:
         op.absolute_symlink(input.maf, output.maf)
+        
+rule _dlbclass_input_seg:
+    input:
+        seg = CFG["inputs"]["master_seg"] if CFG["inputs"]["master_seg"] != "" else CFG["inputs"]["empty_seg"]
+    output:
+        seg = CFG["dirs"]["inputs"] + "variants/{sample_set}--{launch_date}/{seq_type}.input.seg", 
+    run:
         op.absolute_symlink(input.seg, output.seg)
+        
+rule _dlbclass_input_sv:
+    input:
+        sv = CFG["inputs"]["master_sv"] if CFG["inputs"]["master_sv"] != "" else CFG["inputs"]["empty_sv"]
+    output:
+        sv = CFG["dirs"]["inputs"] + "variants/{sample_set}--{launch_date}/{seq_type}.input.sv"
+    run:
         op.absolute_symlink(input.sv, output.sv)
 
 # Symlinks the subsetting categories input file into the module results directory (under '00-inputs/')
@@ -149,7 +161,7 @@ rule _dlbclass_download_refs:
 checkpoint _dlbclass_prepare_maf:
     input:
         maf = expand(
-                    str(rules._dlbclass_input_variants.output.maf),
+                    str(rules._dlbclass_input_maf.output.maf),
                     allow_missing=True,
                     seq_type=CFG["samples"]["seq_type"].unique()
                     ),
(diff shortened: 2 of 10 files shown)


modules/dlbclass/1.0/dlbclass.smk
gillissierra, line 1: Do you want to have `projection` as a wild card as well? 

e.g. in gistic2 module

```
checkpoint _gistic2_prepare_seg:
    input:
        seg = expand(str(rules._gistic2_input_seg.output.seg),
                    seq_type=CFG["samples"]["seq_type"].unique(),
                    allow_missing=True
                    ),
        subsetting_categories = str(rules._gistic2_input_subsetting_categories.output.subsetting_categories)
    output:
        CFG["dirs"]["prepare_seg"] + "{sample_set}--{projection}--{launch_date}/done"
    log:
        log = CFG["logs"]["prepare_seg"] + "{sample_set}--{projection}--{launch_date}.log"
```

The config `input` also specifies that wildcard

```
inputs:
            # The seg files should be "filled", meaning missing segments should be replaced with a dummy segment with log2(CN)=0 and no LOH.
            # The cnv_master module outputs seg files in this filled format
            # Available wildcards: {seq_type} {projection}
            seg: "__UPDATE__"
```

modules/dlbclass/1.0/dlbclass.smk
```diff
@@ -0,0 +1,418 @@
...
+    launch_date = CFG['launch_date']
+else:
+    launch_date = datetime.today().strftime('%Y-%m')
+
+# Interpret the absolute path to this script so it doesn't get interpreted relative to the module snakefile later.
+PREPARE_MAFS =  os.path.abspath(config["lcr-modules"]["dlbclass"]["scripts"]["prepare_mafs"])
+PREPARE_SVS =  os.path.abspath(config["lcr-modules"]["dlbclass"]["scripts"]["prepare_svs"])
+
+# Get date as used in DLBCLass output files
+TODAY = datetime.now().strftime("%d%b%Y")
```
gillissierra, line 76: Is this baked into the DLBCLass code and not changeable? Will it cause issues if users run this module on different days but the same inputs?

modules/dlbclass/1.0/dlbclass.smk
```diff
@@ -0,0 +1,418 @@
...
+        include_non_coding = str(CFG["include_non_coding"]).upper(),
+        mode = "dlbclass",
+        metadata_cols = CFG["samples"],
+        metadata_dim = CFG["samples"].shape,
+        metadata = CFG["samples"].to_numpy(na_value='')
+    script:
+        PREPARE_MAFS
+
+# Prepare MAF GSM file
+def _get_input_maf(wildcards):
```
gillissierra, line 209: Not sure if this case will arise, but there is the possibility of the same sample_set and launch_date having different md5sums. Will this still work in that case?

modules/dlbclass/1.0/dlbclass.smk
```diff
@@ -0,0 +1,418 @@
...
+
+# Setup module and store module-specific configuration in `CFG`
+# `CFG` is a shortcut to `config["lcr-modules"]["dlbclass"]`
+CFG = op.setup_module(
+    name = "dlbclass",
+    version = "1.0",
+    subdirectories = ["inputs", "prepare_inputs", "dlbclass", "outputs"],
+)
+
+# All rules should be run locally
```
gillissierra, line 47: `_dlbclass_prepare_maf` isn't in the list of localrules

gillissierra: if the concern is that it is set to use threads for faster reading of the maf files, it is hard coded to 4, so it shouldn't use up too many resources when run locally.

envs/dlbclass/dlbclass.yaml
```diff
@@ -0,0 +1,16 @@
+name: dlbclass-test
```
Kdreval, line 1: just wondering if this is exported using --from-history flag? Maybe we should consider switch to that one so the envs are easier to rebuild over time 😅

envs/dlbclass/dlbclass.yaml
```diff
@@ -0,0 +1,16 @@
...
+channels:
+  - conda-forge
+  - bioconda
+  - nodefaults
```
Kdreval, line 5: really like the nodefaults here!!

modules/dlbclass/1.0/dlbclass.smk
```diff
@@ -0,0 +1,418 @@
...
+        
+    if(wildcards.cnv_wc == "with_cnv"):
+        cnv_param = "-c " + str(expand(rules._dlbclass_seg_to_gsm.output.cnv_gsm, **wildcards, allow_missing = True)[0])
+    else:
+        cnv_param = ""
+        
+    return cnv_param + " " + sv_param
+
+
+# def get_combine_input(wildcards):
```
Kdreval: This probably can be dropped?

Kdreval (approved):
Thanks!

gillissierra (changes requested):

modules/dlbclass/1.0/dlbclass.smk
lkhilton: Yes and yes. It's annoying, but it's hardcoded into the DLBclass scripts to write the outputs like that.

envs/dlbclass/dlbclass.yaml
lkhilton: Yes this is the `--from-history` version - very minimal and much less than what you'd see if I'd exported everything!

Kdreval: Thanks for confirming! This is great!

modules/dlbclass/1.0/dlbclass.smk
lkhilton: Good catch - I just missed that rule from the list. This isn't a large multiple-samples-at-once module so it's not practical to run it on the cluster. No reason to use it at all, IMO.

modules/dlbclass/1.0/dlbclass.smk
lkhilton: It's hardcoded to run on only hg19/grch37, so we don't need to allow for different projections to be run.

modules/dlbclass/1.0/dlbclass.smk
lkhilton: I ran into this while testing - e.g. if you change any of the criteria under a given sample_set name, you'll get mafs with a different md5sum. It's definitely not optimal but I couldn't think of a workaround. Don't other level 3 modules work similarly? How have you handled this before?

modules/dlbclass/1.0/dlbclass.smk
lkhilton: Yes - will delete.

modules/dlbclass/1.0/dlbclass.smk
gillissierra: The other modules carry the md5sums all the way through the rules including symlinking outputs, then there is an `aggregate` input function that removes the md5sum wildcard before rule all. 

https://github.com/LCR-BCCRC/lcr-modules/blob/a6b99042f7d92e16490134c647c6ac2c8c8a47ea/modules/gistic2/1.2/gistic2.smk#L192C1-L228C10

lkhilton: Yes, now I remember. I think I had to opt not to do that because this integrates both CNV and MAF data with different MD5 sums. I'll add a check for if there are multiple md5sums returned by the checkpoint function, and maybe actually an assertion statement so that it doesn't continue if >1 is found. That would mean the user would have to delete any existing outputs before proceeding to make sure the module is using the right one. Does that make sense?

gillissierra: > both CNV and MAF data with different MD5 sums

The md5sums would only be different between the CNV and MAF data preprocessing if the samples only occur in one or the other. If the samples occur in both the input maf and input seg (or are absent from both), I believe the md5sums will be the same for the preprocessed mafs and seg files for that sample_set. The md5sums are calculated by the sample_ids, not the MAF/CNV info.

gillissierra: Since I think this module is the only level3 that uses both MAF and CNV data, I am also fine with leaving it this way if the md5sums will be different between the preprocesses maf and seg files. Similar future level3 modules can follow this one as an  example for how to set it up when both are needed

gillissierra (approved):

PR #351, opened 2025-04-22
modules/1.0,2.0/tempest

kyaki:
# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable

Example:
```
```

```
CFG = op.setup_module(
    name = "strelka",
    version = "1.2",
)
```


modules/lymphgen/2.1/lymphgen.smk
```diff
@@ -809,4 +809,4 @@ rule _lymphgen_all:
 
 # Perform some clean-up tasks, including storing the module-specific
 # configuration on disk and deleting the `CFG` variable
-op.cleanup_module(CFG)
```
kyaki, line 812: I had done some temporary mods to make lymphgen run easier on this branch, I rolled them back, but for some reason git is tracking this as a change, but you can see its just the same line being deleted then added back. Odd, but actually makes no changes to the lymphgen module.

rdmorin (commented):
This is a large and sophisticated pipeline — thank you for contributing it. However, it shares several structural issues with your other open PR (#358/artifact_alert), and has a few additional ones specific to this PR. The cookiecutter note from that review applies here too, since many of these items would have been scaffolded automatically.

## Blocking

**Neither 1.0 nor 2.0 uses `op.setup_module` / `op.cleanup_module`** — same issue as #358. All five smk files across both versions manage their own output directories manually. This bypasses the standard module infrastructure and prevents the modules from co-existing cleanly with others in a shared workflow.

**`crossmap.yaml` has a hardcoded `prefix:` from a different user's machine** (`/home/dreval/miniconda3/envs/crossmap`). This appears to be a full environment export pasted from another user's system. Remove the `prefix:` line and trim the pinned dependency list to the essential packages only.

**Env files are all inline real files, not symlinks** — all of `modules/tempest/1.0/envs/` and `modules/tempest/2.0/envs/` contain real YAML files. Per repo convention these should live in `envs/<tool>/` at the repo root, with relative symlinks in the module `envs/` directories.

**`bcftools copy.yaml` — filename contains a space**, which breaks on many systems and shell scripts. Rename it (or remove it if it's a duplicate of `bcftools.yaml`).

**No CHANGELOG.md** — required for all new modules.

**Config placeholders should use `__UPDATE__`** — config files use `"..."` as stand-in paths (e.g., `".../GRCh38_no_alt.fa"`). The convention is `__UPDATE__` so that oncopipe can detect unconfigured values at startup.

**File created at parse time in `lymphgen_outputs`** — the `with open(empty_file, "w") as f:` block in `patient_reports.smk` writes a file every time Snakemake parses the workflow, including dry-runs. Any side effects should happen inside rules, not in Python functions called during DAG construction.

**Rule names lack a `_tempest_` prefix** — `trim_umi`, `bwa_align_unsorted`, `generate_pileup`, etc. will collide with same-named rules in other modules when both are included in a shared workflow. All rules should follow the `_<module>_<action>` naming convention.

## Non-blocking (please address)

**Two versions in one PR** — submitting 1.0 and 2.0 simultaneously makes it difficult to review what changed between them and keeps 1.0 from being independently merged and used. Please submit 1.0 first; once it's approved and merged, open a separate PR for 2.0 showing the diff between versions. Alternatively, if 2.0 supplants 1.0 then just open a PR with version 2.0.

**Multiple smk files per version** — having five separate smk files (`align_and_deduplicateTEMPEST.smk`, `variantTEMPEST.smk`, `unpaired_variantTEMPEST.smk`, `make_somaticPON.smk`, `patient_reports.smk`) is non-standard. The convention is one smk per module version. If the workflow is too large for one file, sub-workflows or separate modules (each with their own version directory) are the preferred approach.

**Mostly unpinned env dependencies** — `bwa`, `samtools`, `gatk4`, `fgbio>2`, `fastp`, etc. are all unpinned or loosely pinned. Pin to exact versions and name the env files accordingly (e.g., `bwa_picard_fgbio-3.1.1.yaml`) for reproducibility.

**Script paths built at module load time** — several rules build paths to utility scripts with `os.path.join(UTILSDIR, "script.py")` and embed them directly in f-string shells. Use `workflow.basedir` or pass the path as a `params:` entry so Snakemake can track it correctly.

rdmorin (changes requested):

PR #358, opened 2026-02-02
1.0/artifact_alert -- module to calculate background mutation rates

kyaki:
Here is the code for artifact alert! A snakemake workflow for calculating background mutation rates across a bed file of targets. Usage is fairly simple, its all in the documentation.


rdmorin (commented):
Thanks for the detailed README — the incremental aggregation design is well thought out. However, this module departs from lcr-modules conventions in several ways that need to be addressed before merging.

As a general note: many of the structural issues below (missing `op.setup_module`, rule naming, CHANGELOG, env symlinks) are inserted automatically by the cookiecutter template. If you didn't use it when starting this module, it's worth running it for future modules — it scaffolds all of these components so you don't have to remember them manually.

## Blocking

**Missing `op.setup_module` / `op.cleanup_module`** — every lcr-modules module must set up with `op.setup_module(name, version, subdirectories=...)` and call `op.cleanup_module(CFG)` at the end. The current code manages output directories manually via `os.path.join(config[...], ...)`. Beyond the style issue, skipping this means the module won't register properly when included alongside other modules, and all the shared infrastructure (run tracking, scratch subdirectories, etc.) is bypassed.

**File deletion at parse time** — the `RESET_INDEX` block (lines 53–68) calls `os.remove()` at module load time. Snakemake re-parses the workflow on every invocation, including dry-runs (`--dry-run`). This means the index and tracker files are deleted unconditionally whenever anyone inspects the workflow, before any rules run. Move this logic into a dedicated rule (e.g., `_artifact_alert_reset_index`) that the user targets explicitly.

**`exit(0)` called in the workflow body** — when no new samples are found (line 91), `exit(0)` terminates the Python process immediately, preventing Snakemake from cleaning up. It also returns exit code 0, so calling scripts will not detect that nothing happened. The correct approach is to let the target list be empty (which Snakemake handles gracefully with "Nothing to be done") rather than forcing an exit.

**Orphaned script** — `scripts/make_notspot_manifest.py` references `snakemake.input.aggregated` and `snakemake.params` but there is no corresponding rule in the smk file. Either add the rule or remove the script.

**No CHANGELOG.md** — required by the checklist for all new modules.

**Hardcoded user-specific path in config** — `artifact_alert_config.yaml` contains `/projects/rmorin/reference/igenomes/Homo_sapiens/GSC/GRCh38/...`. Replace with `__UPDATE__`.

**Env file is not a symlink** — `modules/artifact_alert/1.0/envs/samtools.yaml` is a real file. Per repo convention the file should live at `envs/samtools/samtools-<version>.yaml` and the module `envs/` directory should contain a relative symlink pointing to it.

## Non-blocking (please address)

**Smk filename should match module name** — the convention is `<module_name>.smk` (i.e., `artifact_alert.smk`), not `FetchMutationRate.smk`.

**Rule names must be prefixed with `_artifact_alert_`** — e.g., `_artifact_alert_generate_pileup`, `_artifact_alert_calculate_mutation_rates`, etc. Without this prefix, rule names can collide with other modules when multiple modules are included in the same workflow.

**Script path** — `SCRIPTS_DIR` is resolved via `config["lcr-modules"]["_shared"]["lcr-modules"]`, which requires users to add an extra key to their config. Use Snakemake's built-in `workflow.basedir` instead (it always points to the directory containing the smk file).

**Unpinned env dependencies** — `samtools.yaml` uses `samtools>=1.15` rather than an exact version. Pin to a specific version (e.g., `samtools=1.21`) so the environment is reproducible, and name the file accordingly (`samtools-1.21.yaml`).

**Missing trailing newlines** in `envs/samtools.yaml` and `config/artifact_alert_config.yaml`.

rdmorin (changes requested):

PR #375, opened 2026-05-07
Container support

rdmorin:
This attempts to provide Apptainer (Singularity) support for most of the existing modules while attempting to keep existing code backwards-compatible to the extent possible. The main change is to introduce container_envs to default.yaml with verified BioContainers URIs where available and URIs for our own containers deployed through the LCR-scripts repo otherwise. Every rule with a conda environment will now have a container directive that points to the corresponding value in the config. 
This PR also includes an update to oncopipe setup_module() to initialise container_envs and resolve local
.sif paths while leaving URIs and null values unchanged. 

This overhaul required a flexible approach to launch snakefiles that allows users to decide whether to use conda or apptainer at runtime. This is handled by specifying a simple config as a fourth argument to the launcher shell scripts.  These scripts contain detailed usage/examples.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>


rdmorin:
Testing status:

* Reference files workflow runs and sets up the references needed for the capture demo
* capture demo has been run up to (and including) Battenberg

modules/battenberg/1.2/config/default.yaml
```diff
@@ -16,6 +16,13 @@ lcr-modules:
             bedtools: "{MODSDIR}/envs/bedtools-2.29.2.yaml"
             liftover: "{SCRIPTSDIR}/liftover/1.0/liftover.yaml"
 
+        container_envs:
+            battenberg: "docker://ghcr.io/lcr-bccrc/lcr-scripts/battenberg:1.0"
+            wget: null
```
lkhilton, line 26: This is weird. I know it seems silly to have an env just for wget, but we do it because of version incompatibility.

modules/battenberg/1.2/src/calc_sex_status.sh
```diff
@@ -15,12 +15,12 @@ DEBUG="${6:-MISSING}"
 if [[ $X_CHROM == "MISSING" ]]
 then
     X_CHROM=$(samtools view -H ${BAM} |\
-     sed -r 's/\S+:\S+/\n&/g' | perl -ne 's/\s+//g;print "$_\n"' | awk 'BEGIN{FS=":"} $1=="SN" && $2 ~ /X$/ {print $2}')
+     sed -r 's/\S+:\S+/\n&/g' | tr -d ' \t' | awk 'BEGIN{FS=":"} $1=="SN" && $2 ~ /X$/ {print $2}')
```
lkhilton, line 18: Wow, you let Claude remove your perl!?

modules/manta/CHANGELOG.md
```diff
@@ -5,6 +5,16 @@ All notable changes to the `manta` module will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.4] - 2026-05-06
+
+This release was authored by Ryan Morin.
+
+- Added `container_envs` to `config/default.yaml` with verified BioContainers URIs for all four environments (`tabix`, `manta`, `augment_manta_vcf`, `svtools`). The `augment_manta_vcf` container is hosted at `ghcr.io/lcr-bccrc/lcr-scripts/augment_manta_vcf:1.1` and built from the lcr-scripts repository. A `null` value for any entry causes that rule to fall back to its conda environment when running with `--use-apptainer`.
```
lkhilton, line 12: Seems like this is the only module where the changelog has been edited - also does this require a version bump?

CONTAINER_STATUS.md
```diff
@@ -0,0 +1,132 @@
...
+
+---
+
+## Future Work: Pre-flight container check
+
+**Idea:** Add an `onstart:` block to the demo (and eventually production) Snakefiles that aborts early when `--use-singularity` is active but one or more rules in the workflow have `container: None`.
+
+Currently, rules with `container: None` silently fall back to the host environment under `--use-singularity`, so failures only surface when the tool is missing — not at startup.
+
+```python
```
lkhilton, line 20: So essentially every rule in a Snakefile has to have a container to be able to use singularity? How do we resolve the partial modules?

lkhilton (commented):
Shaping up nicely. Seems to have implemented very consistent changes throughout. Wondering out loud if this is also an opportunity to upgrade to a current version of Snakemake?

Also how will we confirm that the containerized software is equivalent to the condo version?

modules/battenberg/1.2/src/calc_sex_status.sh
rdmorin: It was definitely a tough call to make but the issue is that many of the basic Linux containers don't have Perl installed by default.

modules/manta/CHANGELOG.md
rdmorin: I was hoping to avoid a new version for any tool as long as the changes only impact how the tool runs in apptainer mode

CONTAINER_STATUS.md
rdmorin: My goal is to not have any partial modules left by the time this PR is complete. If there's any tool that can't be converted we'll just need a way to have those exit in a way that alerts the user to conda being the only option currently available.

modules/manta/CHANGELOG.md
lkhilton: Sorry the missing context is that the changelog says this is now version 2.4, which doesn't exist unless there's a version bump to Manta, so maybe this changelog update is unnecessary?

PR #377, opened 2026-06-04
Update Exome CNV callers and add panel_of_normals module

gillissierra:
Large update includes:

- added `panel_of_normals` module that makes files for `cnvkit` and `purecn` modules
- updated `cnvkit` and `purecn`  to use the new `panel_of_normals` files
- updated `cnvkit` and `purecn`  to use updated lcr-scripts
- updated purecn to try different segmentation methods in the `run` step

For more details see the `CHANGELOG.md` files

`purecn` and `cnvkit` will get version bumps

Example:
```
```

```
CFG = op.setup_module(
    name = "strelka",
    version = "1.2",
)
```


modules/cnvkit/1.0/cnvkit.smk
```diff
@@ -20,577 +20,532 @@ import oncopipe as op
...
+        cnvkit.py metrics {input.cnr} -s {input.cns} > {output.metrics} 2> {log.log}
+        """)
 
-rule _cnvkit_to_seg:
+# Create seg formatted for cnv_master and other lcr-mdoules
+rule _cnvkit_cnv2igv:
     input:
-        cns = str(rules._run_cnvkit_call_vcf.output.cns),
+        cns = str(rules._cnvkit_call.output.cns),
+        cnv2igv = ancient(CFG["inputs"]["cnv2igv"])
```
lkhilton: Should this be a param instead of an input, to be more consistent with how the other scripts are referenced?

modules/cnvkit/1.0/cnvkit.smk
lkhilton, line 1: You should add your name as a contributor to all of these 🙂

modules/panel_of_normals/1.0/schemas/base-1.0.yaml
lkhilton, line 1: Highly recommend adding a custom schema for this module that mandates the capture_space wildcard.

lkhilton (commented):
SO happy to see this in a PR! I don't see any major issues. Would be interested to have @Jwong684 take a look.

modules/cnvkit/1.0/cnvkit.smk
gillissierra: It's like this to be consistent with the other modules. I would like to take this opportunity to change it for all cnv modules.

modules/purecn/1.0/purecn.smk
```diff
@@ -0,0 +1,1356 @@
...
+    tbl = CFG["runs"]
+    this_build = tbl[(tbl.tumour_sample_id == wildcards.tumour_id) & (tbl.tumour_seq_type == wildcards.seq_type)]["tumour_genome_build"].tolist()[0]
+    this_space = tbl[(tbl.tumour_sample_id == wildcards.tumour_id) & (tbl.tumour_seq_type == wildcards.seq_type)]["tumour_capture_space"].tolist()[0]
+    
+    txt = CFG["dirs"]["outputs"] + "{mode}/{seq_type}--{genome_build}/{tumour_id}/{tumour_id}_segmentation_fn.txt".replace("{capture_space}", this_space[0]).replace("{genome_build}", this_build[0])
+    return txt
+
+rule _purecn_best_seg_info:
+    input:
+        # txt = _get_segmentation_fn_txt,
```
Jwong684: commented out

modules/panel_of_normals/1.0/config/default.yaml
```diff
@@ -0,0 +1,105 @@
+lcr-modules:
+
+    panel_of_normals:
```
Jwong684, line 3: Is there a way to dictate it so that custom panels-of-normals can be used/generated? I.e. PoNs that use tumour samples as pseudo-normals?

modules/panel_of_normals/1.0/panel_of_normals.smk
```diff
@@ -0,0 +1,1392 @@
...
+# Collects all normals per genome_build--capture_space combo
+def _get_normal_bams_per_combo(wildcards):
+    tbl = META
+    samples = tbl[(tbl.genome_build == wildcards.genome_build) & (tbl.capture_space == wildcards.capture_space)]["sample_id"].tolist()
+    normals = expand(
+        str(rules._panel_of_normals_input_bam.output.bam),
+        sample_id = samples,
+        allow_missing = True)
+    return normals
+
```
Jwong684, line 175: What would happen if there are new samples that get added over time that match the same capture_space and genome_build? Is there a way to factor in "cohorts" or exclude "cohorts" that shouldn't be included in the PoN?

modules/purecn/1.1/purecn.smk
```diff
@@ -0,0 +1,1356 @@
...
+                else
+                    touch {output.done}
+                fi
+            else
+                touch {output.done}
+            fi
+        else
+            touch {output.done}
+        fi
+        """
```
Jwong684, line 772: Nice implementation. Does this rule indicate which method it ultimately uses? I guess it's based on whichever method comes last in the log file. Seems like you get a "done" file for all three methods as input initially and then a final "done" file.

Jwong684: Does this pipeline work if you just want pureCN to be done (i.e. no cnvkit?)

Jwong684 (commented):
Just some questions/comments. I like the separation of PoN generation in its own module. 
1) The only thing that may come up is if new samples get added in the future that matches the capture_kit + genome_build (which happens a lot for agilent_v6, grch37). Would the PoN be regenerated every time this happens, and consequently, all the tumours have to be reran? What about for cohorts that have poor quality in general (ex. Reddy) which would drag down/add noise to certain PoNs.

2) What about custom PoNs derived from tumours? 

3) And also FFPE vs FF data? Maybe a custom list of normals can be useful.

4) Do both CNVkit and PureCN PoN have to be generated at the same time? Specifically, I noticed that CNVkit uses the refFlat panel of normal if there are no normals (nice add), but pureCN does not have any equivalent for this.

modules/panel_of_normals/1.0/config/default.yaml
gillissierra: This was pretty extensively discussed in lab meeting, slack, etc. based on the unix group - genome_build - capture_space combos that we have currently. The dlbcl_schmitdz cohort actually uses tumours as pseudo-normals already, chosen based on array data. So yes it's possible to do so in the smk that runs the panel_of_normals (i.e. for us, the one in gambl), and then specify those files in CNVkit and PureCN

modules/panel_of_normals/1.0/panel_of_normals.smk
gillissierra: We discussed that this module is run one time and only get's ran when an entirely new unix_group - capture_space - genome_build combo is added to the metadata. It will not get reran when there are new samples added, so that the tumours will not need to be reran. 
For excluding specific cohorts, this can be done in the smk that runs the module (ie. for us, it would be in the gambl smk). Those cohorts / samples would be excluded from the metadata table going through the module. It wouldn't make sense to implement "cohort" in the lcr-module itself since that column isn't required to be in the metadata.

modules/purecn/1.1/purecn.smk
gillissierra: > I guess it's based on whichever method comes last in the log file

I don't think that's true. There isn't a done file for all three methods, there is only one done file for the checkpoint. Then the input function for the next rule (the symlink rule) checks for existance of the output files from this run rule. They would only exist if the command completed, and even if they existed for multiple segmentation methods, it chooses the files to symlink based on the hierarchy that was used in the current implementation of the module.

> Does this pipeline work if you just want pureCN to be done (i.e. no cnvkit?)

Yes, the config sets the `mode`. So if you only wanted PureCN:
```
modes:  ["purecn_denovo"]
````
Then run with subsetting the metadata to the samples of interest

File: modules/panel_of_normals/1.0/panel_of_normals.smk (hunk 4 above)
Comment (gillissierra) [reply modules/panel_of_normals/1.0/panel_of_normals.smk:175]: If new samples are added and the files are regenerated, then it should also get a version bump so that those files do not over-write current ones

**Comment by gillissierra** [comment]
> Do both CNVkit and PureCN PoN have to be generated at the same time?

No, but if PureCN is to be ran with mode "purecn_cnvkit" then CNVkit would have to be ran first for those samples.

PureCN does not have an equivalent to CNVkit's refFlat file unfortunately, meaning combos with no panel of normals cannot be ran with "purecn_denovo" but could be ran with "purecn_cnvkit" mode

**Comment by gillissierra** [comment]
> FFPE vs FF data

This wouldbe nice to have, but is out of the scope of this update, since ffpe_vs_frozen is not a required column in the metadata, and making it one through the schema here could make the modules less flexible for external users. There is also some evidence that using FFPE normals for both FFPE and frozen gives cleaner results, so I don't want to implement splitting them apart if it wouldn't actually make the results better

**Comment by gillissierra** [comment]
> Do both CNVkit and PureCN PoN have to be generated at the same time?

Based on how I set it up, it won't be an issue to run both at the same time. The refFlat is based on the config, not the metadata. I think the combos that get the refFlat files wouldn't be in the metadata so they won't be sent through the PureCN rules in the panel_of_normals module

**Comment by gillissierra** [comment]
Brett showed me a few example runs where using the unpadded bed file for the capture space gave better estimates of purity. I had thought that I set up the pipelines to use the unpadded bed files from the reference workflows, but I will double check

**Comment by rdmorin** [comment]
Hi Sierra, here is a file-by-file breakdown of the conflicts and how to resolve them.

## Background

While your branch was open, master received a large batch of infrastructure changes that touched several of the same files you edited: container directives were added to every module, `pkg_resources` was replaced with `importlib.metadata`, `ancient()` was added to sequenza inputs to fix persistent reruns, and `op.relative_symlink` was changed to `op.absolute_symlink` for input-file symlinks. Your changes and those need to be combined.

## Step 1 — merge master into your branch

```bash
git fetch origin
git merge origin/master
# Git will drop you into conflict-resolution mode for the files below
```

---

## File-by-file resolution guide

### `modules/battenberg/1.1/battenberg.smk` and `config/default.yaml`

**What master added:** `importlib.metadata` import, container directives on multiple rules.  
**What you added:** your name in `# Contributors:`, moved `cnv2igv` from `input:` to `params.cnv2igv_script`.

**Resolution:** accept **master's version** of the file (take the `=======` / `>>>>>>>` half of each conflict), then re-apply your two small changes on top:
- Add `, Sierra Gillis` to the `# Contributors:` line.
- In `_battenberg_convert_coordinates` (or whichever rule runs cnv2igv): remove `cnv2igv` from `input:`, add `cnv2igv_script = CFG["options"]["cnv2igv_script_path"]` to `params:`, and update the shell command to use `{params.cnv2igv_script}`.
- Add `cnv2igv_script_path` to the relevant block in `config/default.yaml`.

---

### `modules/controlfreec/1.3/controlfreec.smk` and `config/default.yaml`

Same pattern as battenberg. Master added `importlib.metadata` and many `container:` directives.

**Resolution:** accept **master's version**, then re-apply:
- Your name in `# Contributors:`.
- Move `cnv2igv` from `input:` to `params.cnv2igv_script` in the relevant rule and update the shell command.
- Add `cnv2igv_script_path` to `config/default.yaml`.

---

### `modules/sequenza/1.4/sequenza.smk` and `config/default.yaml`

Master made significant sequenza-specific fixes here (added `ancient()` to several script inputs to prevent persistent reruns, added container directives, fixed the filter_seqz thread argument, removed `temp()` from fill_regions outputs). These are important bug fixes you need to keep.

**Resolution:** accept **master's version**, then re-apply:
- `Bruno Grande, Sierra Gillis` in `# Contributors:`.
- Move `cnv2igv` from `input:` to `params.cnv2igv_script` and update the shell command.
- Add `cnv2igv_script_path` to `config/default.yaml`.

---

### `modules/cnvkit/1.0/cnvkit.smk` and `config/default.yaml`

You rewrote this module completely. Master only made small changes to the old version (swapped `op.relative_symlink` → `op.absolute_symlink` for input symlinks, added container directives).

**Resolution:** accept **your rewritten version** for these files. Then check:
- Input-file symlink rules (anything that symlinks data *into* `00-inputs/`) should use `op.absolute_symlink`, not `op.relative_symlink`. Output symlink rules (anything that creates the `99-outputs/` symlinks) should use `op.relative_symlink(…, in_module=True)`.

---

### `workflows/reference_files/2.5/reference_files.smk` and `reference_files_header.smk`

You added new rules (`get_baitspace_bed_download`, `check_baitspace_contigs`, `download_baitspace_bed`, `add_remove_chr_prefix_baits_bed`) and a `capture_space` wildcard constraint. Master fixed a dead repeatmasker URL and the gnomad VCF processing pipeline.

**Resolution:** these changes are in different parts of the file, so conflicts are likely small. Keep **both sets of changes** — accept the non-overlapping hunks from each side. Where there is a true line-level conflict (e.g. a wildcard constraint block), merge the values manually so both sets of constraints are present.

---

## After resolving all files

```bash
git add modules/battenberg/1.1/battenberg.smk modules/battenberg/1.1/config/default.yaml \
        modules/controlfreec/1.3/controlfreec.smk modules/controlfreec/1.3/config/default.yaml \
        modules/sequenza/1.4/sequenza.smk modules/sequenza/1.4/config/default.yaml \
        modules/cnvkit/1.0/cnvkit.smk modules/cnvkit/1.0/config/default.yaml \
        workflows/reference_files/2.5/reference_files.smk \
        workflows/reference_files/2.5/reference_files_header.smk
git commit
```

If you want me to take a stab at resolving these conflicts let me know!

gillissierra:
Conflicts outlined in the above comment have been resolved

rdmorin (dismissed):

gillissierra:
Note: PureCN was verison bumped to 1.1, but there is no 1.0 to restore. The reasoning for doing this version bump is that there are existing results in GAMBL under version 1.0, despite there not being a version 1.0 of the module in master

schemas/base/base-1.0.yaml
lkhilton, line 1: For custom schemas, you have to create a new yaml file to add new required values. If you add it to `base-1.0.yaml`, it will make that value required for all modules that use the `base-1.0.yaml` schema (i.e. every module we have). So you can make a new schema called `capture_space-1.0.yaml` in `schemas/capture_space` and include it in the modules that need to require `capture_space`.

modules/cnvkit/CHANGELOG.md
```diff
@@ -5,6 +5,20 @@ All notable changes to the `cnvkit` module will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+# [2.0] 2026-06-04
```
lkhilton, line 8: The changelog references V2.0 but all the module files are still on 1.1 - do you plan to update them all to 2.0?

lkhilton: Ditto for the purecn changelog.

lkhilton (changes requested):
Just need to make a new schema and revert the changes to base-1.0.yaml, and then resolve the version discrepancy. Looking good!

modules/cnvkit/CHANGELOG.md
gillissierra: I think it should have been 2.0 to reflect the large, but it makes more sense to just edit the changelogs now because outputs exist under 1.1 for CNVkit since I've already ran it.

gillissierra: Changed to 2.0 to reflect the large change

lkhilton (approved):

PR #378, opened 2026-07-03
Module/drivemutr/1.0

HoumanLM:
# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable

Example:
```
```

```
CFG = op.setup_module(
    name = "strelka",
    version = "1.2",
)
```

### DIFF
--- a/envs/drivemutr/drivemutr-1.0.post-deploy.sh
+++ b/envs/drivemutr/drivemutr-1.0.post-deploy.sh
@@ -1,0 +1,16 @@
+#!/usr/bin/env bash
+
+set -o pipefail
+set -e  # Exit immediately if a command fails
+
+R --vanilla -q -e 'devtools::install_github(
+    "Simon-Coetzee/motifBreakR",
+    dependencies = FALSE,
+    upgrade = "never"
+)'
+
+R --vanilla -q -e 'devtools::install_github(
+    "morinlab/GAMBLR.data",
+    dependencies = FALSE,
+    upgrade = "never"
+)'
--- a/envs/drivemutr/drivemutr-1.0.yaml
+++ b/envs/drivemutr/drivemutr-1.0.yaml
@@ -1,0 +1,26 @@
+name: drivemutr_r44
+channels:
+  - conda-forge
+  - bioconda
+  - defaults
+dependencies:
+  - r-base=4.4
+  - snakemake
+  - r-dplyr
+  - r-purrr
+  - r-tidyr
+  - r-readr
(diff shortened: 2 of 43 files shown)


rdmorin:
Thank you for this detailed and well-structured module! A few issues to address before merge:

**Blocking**

1. **`prefix:` in conda YAML** — `envs/drivemutr/drivemutr-1.0.yaml` contains `prefix: /home/hmirhosseini/miniconda3/envs/drivemutr_r44`. This path is user-specific and will cause conda to write into your home directory on every other system. Please remove the `prefix:` line entirely.

2. **Global `wildcard_constraints:`** — The `wildcard_constraints:` block at module scope (for `gene`, `track`, `lam`) must be moved inside each rule that uses those wildcards. Module-level constraints leak into the global namespace when modules are co-included, which can silently constrain identically-named wildcards in unrelated modules. See other modules (e.g. battenberg) for the per-rule pattern.

**Should fix**

3. **`snakemake` as a conda dependency** — `snakemake` is already provided by the execution environment; listing it as a dependency in the conda YAML can cause version conflicts. Please remove it.

4. **`op.relative_symlink` missing `in_module=True`** — `_drivemutr_output_tracks` calls `op.relative_symlink(input.tsv, output.tsv)` without `in_module=True`. Without that flag the relative target path is computed from the snakemake launch directory rather than the module output directory, producing a broken symlink. Please add `in_module=True` (see `_cnaqc_output` in the cnaqc module for the pattern).

**Potential bug**

5. **`BSgenome.Hsapiens.UCSC.hg19` hardcoded in `src/R/get_tf_results.R`** — The module config accepts both `GRCh37` and `GRCh38` builds, but `get_nucleotides_prop` unconditionally loads the hg19 BSgenome. A GRCh38 run would either error or return nucleotide compositions from the wrong genome. Please make the genome object conditional on the build passed through the Snakemake config/params.

**Minor**

6. **PR checklist** — All checklist boxes are currently unchecked. Please tick off the items that apply before requesting a final review.

7. **Contributors line** — Please add your name to the `# Contributors:` line at the top of `drivemutr.smk`.

rdmorin (changes requested):
Please see the detailed review comment above. Requesting changes on the two blocking issues (prefix: in conda YAML, global wildcard_constraints:) and the op.relative_symlink in_module=True fix before this can be merged.

HoumanLM:
Thanks for reviewing this module and for the feedback. I’ve addressed the blocking issues you mentioned.

PR #380, opened 2026-07-06
Module/wgcna/1.0

HoumanLM:
# Pull Request Checklists

### Required

  ```bash
  ```

### If applicable

Example:
```
```

```
CFG = op.setup_module(
    name = "strelka",
    version = "1.2",
)
```


rdmorin (commented):
Thanks for this — a substantial new module. Several items to address before merging:

## Blocking

**`envs/wgcna/wgcna-1.73.yaml` has a hardcoded `prefix:` field** (`/home/hmirhosseini/miniconda3/envs/wgcna_r44`). Please remove this line.

**`saveTOMs = TRUE` with `saveTOMFileBase = "pwrTOM"` in `get_co_expressed_genes_modules.R`** — `blockwiseModules` will write large TOM files (`pwrTOM-block.1.RData` etc.) to whatever directory Snakemake is launched from, not to the module output directory. Either set `saveTOMs = FALSE` or point `saveTOMFileBase` at a path inside the module output directory.

**`_wgcna_filter_variance_genes` is missing a `threads:` directive** — the checklist requires that every non-`localrule` has both `threads` and `resources`. This rule has `resources` but no `threads`. Add `threads: 1` (or add it to the `threads:` section of `default.yaml` with a meaningful value).

**Global wildcard constraint** — `wgcna.smk` sets `wildcard_constraints: pathology = "[^/]+"` at module scope. The checklist requires all wildcard constraints to be per-rule. Move this into each rule that uses the `{pathology}` wildcard.

## Non-blocking (please address while the file is open)

**Possible bug in soft-threshold selection** — `get_co_expressed_genes_modules.R` selects the optimal soft power with:
```r
opt_power = sft_filtered_median_mad$fitIndices %>%
    filter(SFT.R.sq >= 0.8) %>%
    slice_min(SFT.R.sq) %>%
    pull(Power)
```
`slice_min(SFT.R.sq)` picks the row with the *lowest R²* that still passes, which can return a suboptimal higher power when R² is non-monotonic. The standard WGCNA approach is the *lowest power* achieving R² ≥ 0.8: `slice_min(Power)`. Also worth adding a check for the case where no power achieves the threshold (empty result), so the error is informative rather than cryptic.

**`op.relative_symlink` used for an input symlink** — `_wgcna_input_matrix` calls `op.relative_symlink`. Input symlinks should use `op.absolute_symlink` (relative symlinks to external files break if the module directory is moved).

**Dead code in `normalize_gene_expression.R`** — `convert_ids = "TRUE"` and `GRCh <- "38"` (lines ~551–552) are assigned but never referenced. Remove them.

**`save.image(...)` in `normalize_gene_expression.R`** — this dumps the entire R session to `DESeq2.RData` in the parent of the plots directory on every run. Looks like a debug artifact; please remove or gate it behind an option.

**`snakemake` listed as a conda dependency in `wgcna-1.73.yaml`** — Snakemake manages the environments from outside; listing it as a dependency inside an env is unusual and could cause version conflicts.

**Missing trailing newlines** — `filter_high_variance_genes.R`, `get_co_expressed_genes_modules.R`, `run_filter_high_variance_genes.R`, `run_get_co_expressed_genes_modules.R`, and `run_normalize_gene_expression.R` are all missing a trailing newline.

rdmorin (changes requested):

HoumanLM:
Thanks for the detailed review. I’ve addressed the requested changes and pushed the updates. 
For the `Possible bug in soft-threshold selection` comment, my approach already considers the lowest power where R² is greater than or equal to 0.8. I have also added a message for cases where no values meet the threshold.

PR #381, opened 2026-07-07
Add CNAqc module v1.0

rdmorin:
There's a lot on this branch. The primary need is the updated Sage with patched code to allow completion on bams that were silently failing prematurely


demo/capture_Snakefile.smk
```diff
@@ -69,8 +69,8 @@ rule all:
     input:
         rules._bam2fastq_all.input,
         rules._bwa_mem_all.input,
-        rules._qc_all.input,
+#        rules._qc_all.input,
```
lkhilton: Do you mean to leave these lines commented out in this PR?

demo/run.sh
```diff
@@ -1,18 +1,42 @@
...
-
-
-# Default to all targets
 snakefile=$1
 TARGETS=${2:-all}
 snakemake_flags=$3
+runtime_config=$4
+
+# Parse runtime config (emits empty strings when no config is provided)
+eval "$(python3 "$SCRIPT_DIR/../utils/parse_runtime_config.py" "${runtime_config:-}")"
```
lkhilton, line 24: This caused some problems in GAMBL when people didn't have access to this script. We patched it like this: 
```suggestion
# Parse runtime config — drives conda vs container mode
if [[ -e "$LCR_MODULES_DIR/demo/parse_runtime_config.py" ]]; then
    eval "$(python3 "$LCR_MODULES_DIR/demo/parse_runtime_config.py" "${runtime_config:-}")"
else
    echo "Warning: $LCR_MODULES_DIR/demo/parse_runtime_config.py does not exist. Defaulting to conda mode. "
fi

# Conda flags (suppressed when running containers — the two modes are mutually exclusive)
conda_flags=()
if [ -z "$SNAKEMAKE_CONTAINER_FLAG" ]; then
    conda_flags+=(--use-conda --conda-frontend conda --conda-prefix /projects/rmorin_scratch/conda_environments)
fi
```

modules/cnaqc/1.0/envs/cnaqc-1.0.yaml
```diff
@@ -0,0 +1,42 @@
+channels:
```
lkhilton: The files in here should all be symlinks to the main envs directory.

modules/cnaqc/1.0/schemas/base-1.0.yaml
```diff
@@ -0,0 +1,30 @@
+$schema: "http://json-schema.org/draft-06/schema#"
```
lkhilton, line 1: Should also be a symlink

modules/cnaqc/1.0/cnaqc.smk
```diff
@@ -0,0 +1,260 @@
...
+        op.absolute_symlink(input.maf, output.maf)
+
+
+# Run CNAqc analysis
+rule _cnaqc_run:
+    input:
+        sub = str(rules._cnaqc_input_subclones.output.sub),
+        cp  = str(rules._cnaqc_input_cellularity_ploidy.output.cp),
+        maf = str(rules._cnaqc_input_maf.output.maf),
+        script = CFG["inputs"]["cnaqc_script"]
```
lkhilton: The scripts should be params instead of inputs. Keeping them as inputs means that if they get updated (e.g. by a git pull) this rule will want to rerun.

modules/cnaqc/README.md
```diff
@@ -0,0 +1,45 @@
+# CNAqc Module
```
lkhilton, line 1: This is a nice addition.

modules/sage/1.2/sage.smk
```diff
@@ -0,0 +1,262 @@
...
+    input:
+        tumour_bam = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{tumour_id}.bam",
+        normal_bam = CFG["dirs"]["inputs"] + "bam/{seq_type}--{genome_build}/{normal_id}.bam",
+        fasta = str(rules._input_references.output.genome_fa),
+        hotspots = rules._download_sage_references.output.hotspots,
+        high_conf_bed = str(rules._download_sage_references.output.high_conf_bed),
+        cache = ancient(str(rules._download_sage_references.output.cache)),
+        panel_bed = _sage_get_capspace,
+        main_chromosomes = reference_files("genomes/{genome_build}/genome_fasta/main_chromosomes.txt"),
+        sage_jar = CFG["options"]["sage_jar"]
```
lkhilton: Again I think this is preferable as a param instead of an input

modules/sage/1.2/envs/bcftools-1.10.2.yaml
```diff
@@ -0,0 +1,36 @@
+name: test-bcftools
```
lkhilton: Make sure these are all symlinks in this module, too.

utils/filter_samples_dryrun.py
```diff
@@ -0,0 +1,302 @@
...
+        })
+
+    return jobs
+
+
+# ---------------------------------------------------------------------------
+# Aggregation
+# ---------------------------------------------------------------------------
+
+def build_pair_table(
```
lkhilton, line 171: Would this work for e.g. RNAseq modules that only have a sample_id wildcard and don't run in paired mode?

lkhilton (changes requested):

demo/capture_Snakefile.smk
rdmorin: Nope. I'll revert those changes

demo/run.sh
rdmorin: I'll add this check. BTW, the script has moved to utils since it's not really part of the demo.

utils/filter_samples_dryrun.py
rdmorin: I'm not sure. This script needs documentation and more testing. I can remove from this PR if you're worried about it

modules/sage/1.2/envs/bcftools-1.10.2.yaml
rdmorin: done

modules/cnaqc/1.0/cnaqc.smk
rdmorin: fixed in both cases

rdmorin:
I responded to all Laura's comments. Let me know if anything else needs to be changed.

lkhilton (dismissed):

utils/filter_samples_dryrun.py
lkhilton: I think it can stay - a problem for another day.

lkhilton:
Kostia had a comment about fixing the demo symlink structure on Slack, did you implement that?

PR #383, opened 2026-07-09
Add maf_diff/1.0 and igv_reports/1.0 modules

rdmorin:
## Summary

- Adds `maf_diff/1.0`: compares MAF files from two variant callers per tumour/normal pair, producing caller-unique MAFs and a long-format summary TSV with variant counts and driver gene statistics (using the bundled LLMPP curated tier-1 gene list)
- Adds `igv_reports/1.0`: generates self-contained per-sample HTML variant review reports using igvteam/igv-reports for coding driver variants unique to each caller; supports multiple MAF inputs via a `{tool}` wildcard so a single module include handles any number of callers
- Both modules include READMEs and CHANGELOGs

### Required

- [x] I tested the module on real data for all supported `seq_type` values. **Note: tested on `genome` (sage vs slms-3). Capture testing is pending availability of sage capture outputs.**
- [x] I recursively searched for all comments containing `TODO` to ensure none were left.

### If applicable

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rdmorin:
Superseded by #384 — same changes but scoped to maf_diff, igv_reports, and vcf2maf only (battenberg changes excluded).

PR #384, opened 2026-07-09
Add maf_diff/1.0 and igv_reports/1.0 modules; fix vcf2maf/1.3

rdmorin:
## Summary

- Adds `maf_diff/1.0`: compares MAF files from two variant callers per tumour/normal pair, producing caller-unique MAFs and a long-format summary TSV with variant counts and driver gene statistics (using the bundled LLMPP curated tier-1 gene list)
- Adds `igv_reports/1.0`: generates self-contained per-sample HTML variant review reports using igvteam/igv-reports for coding driver variants unique to each caller; supports multiple MAF inputs via a `{tool}` wildcard so a single module include handles any number of callers
- Both modules include READMEs and CHANGELOGs
- Fixes `vcf2maf/1.3`: auto-reheader VCF when sample columns aren't named TUMOR/NORMAL; fix deblacklist_ssm.R writing literal NA strings to MAF

### Required

- [x] I tested the module on real data for all supported `seq_type` values. **Note: tested on `genome` (sage vs slms-3). Capture testing is pending availability of sage capture outputs.**
- [x] I recursively searched for all comments containing `TODO` to ensure none were left.

### If applicable

🤖 Generated with [Claude Code](https://claude.com/claude-code)


modules/vcf2maf/1.3/src/deblacklist_ssm.R
```diff
@@ -40,10 +40,10 @@ if (nrow(this_maf)>0){
         verbose = TRUE,
         drop_threshold = drop_threshold
     ) %>%
```
rdmorin, line 42: These changes will get rid of the literal NA values we keep seeing in our MAFs when they should just be empty instead

gbanco4 (changes requested):
These new modules will be two great additions. Changes look great overall. Here are some comments before merging

  What's solid

  - Both new modules' .smk files otherwise follow the module-authoring guidelines well: str()-wrapped rules. references,
  proper localrules grouping (aside from the one gap above), grouped input-symlinking, complete _all target rules, and
  __UPDATE__ placeholders in default.yaml.
  - The deblacklist_ssm.R/maf_diff.py/filter_maf_for_igv.py na=""/na_rep="" fix is applied consistently everywhere a MAF is
  written, which is good — no half-applied fix.
  - The "no driver variants → skip create_report, touch empty output" branch in _igv_reports_run avoids invoking
  igv-reports on an empty MAF and is clearly logged.
  - Good, actionable READMEs and CHANGELOGs for both new modules.
  
Things I'd fix before merge

  1. _igv_reports_index_gtf is missing threads/resources and isn't a localrule (modules/igv_reports/1.0/igv_reports.smk)
  This rule sorts, bgzips, and tabix-indexes a genome-wide GTF — real, non-trivial work — but has neither a
  threads:/resources: directive nor a localrules: entry, and default.yaml's threads/resources blocks have no index_gtf key
  at all. Every comparable indexing rule elsewhere in the repo (e.g. sage/1.2's _run_sage/_sage_filter_vcf, which also do
  bgzip/tabix) declares both. On a cluster-executor profile this will silently fall back to unconfigured defaults. This
  directly contradicts the PR's own checklist claim ("Every rule ... is either listed under localrules or has the threads
  and resources directives" — checked ✅).

  2. Reheadered VCF temp file leaks on vcf2maf.pl failure (modules/vcf2maf/1.3/vcf2maf.smk)
  The new cleanup line is chained with && after the if/else block that runs vcf2maf.pl:
  ... fi && if [[ "$VCF_INPUT" != "{input.vcf}" ]]; then rm -f "$VCF_INPUT"; fi && touch {output.vep}
  If vcf2maf.pl itself fails, the first if/fi returns non-zero, && short-circuits, and the rm -f "$VCF_INPUT" cleanup is
  skipped — every failed run on a non-TUMOR/NORMAL-headered VCF leaks a mktemp'd VCF. The job correctly fails (so this
  isn't silently wrong), but it's a real resource leak on the error path worth a trap or unconditional cleanup.

  3. ? → _loss substitution in filter_maf_for_igv.py is broader than its justification
  (modules/igv_reports/1.0/src/filter_maf_for_igv.py:1557-1563)
  The comment explains this exists to dodge a Tabulator.js crash on ? in HGVS notation like p.M1? (start-loss). But the
  code does a blanket str.replace("?", "_loss", regex=False) across every configured info_columns display column — any
  other legitimate use of ? in HGVS (e.g. a generic "consequence unknown" annotation unrelated to a start codon) would be
  silently relabeled _loss, which asserts something the data doesn't say. Worth either scoping the replacement to the
  actual p.M1? pattern or renaming to something consequence-neutral (e.g. _unk).

  Minor / worth a look

  - Duplicated 1009-line driver gene TSV: modules/maf_diff/1.0/etc/any_tier1_BL_FL_DLBCL_MCL_MZL_PMBL.tsv and
  modules/igv_reports/1.0/etc/any_tier1_BL_FL_DLBCL_MCL_MZL_PMBL.tsv are byte-identical, committed as two separate real
  files. Everything else shared between the modules (envs, schemas) is symlinked — this file isn't, so the two copies can
  silently drift on a future update to one but not the other.
  - Unused pandas-1.3.0 env: envs/pandas/pandas-1.3.0.yaml (plus its symlinks into both new modules) is committed but never
  referenced by either module's conda_envs in default.yaml — both only wire up pandas-1.4.2.yaml. Looks like dead
  leftover.
  - Reheader check only verifies TUMOR, not NORMAL: vcf2maf.smk's new logic triggers reheadering if TUMOR is absent from
  the #CHROM line, then renames both tumour_id→TUMOR and normal_id→NORMAL via awk. If the normal sample's column name
  doesn't happen to match normal_id exactly, that rename silently no-ops while the log claims "TUMOR/NORMAL... reheadered,"
  which will obscure the real cause of a later vcf2maf.pl failure.
  - patch_igv_reports_html.py is a brittle exact-string patch against igv-reports' generated HTML template (pinned to
  v1.16.0 in the env). If igv-reports is ever bumped, a template change will make the patch silently no-op (it only prints
  a warning to a log file, doesn't fail the job) and the browser-init bug it's working around could quietly regress. Worth
  a comment noting this is tied to the pinned version, or a follow-up to file the fix upstream.

Kdreval (approved):

gbanco4 (approved):

PR #387, opened 2026-07-17
use default sage panel bed

Kdreval:
This update addresses the "too many open files" error
Confirmed this works on the hg19-clc-other failing exomes


rdmorin (approved):
Nice that this simplifies the process and causes the same filtering paradigm to be applied to both seq_type

